PR #3, opened 2015-06-30
Refactored entire Reflector to be more flexible

### LINKED ISSUE #4: Replace getClassesFromFile with a better API for getAllClasses
From discussion on https://github.com/asgrim/better-reflection/pull/3/files#r33560669

> `getClassesFromFile` without parameter? Seems a bit weird. The API is probably also redundant, now that you have the locator:
> 
> ```php
$allInString = (new Reflector(new FilenameSourceLocator('path/to/file.php'))->getAllClasses();
$allInFile = (new Reflector(new StringSourceLocator($code))->getAllClasses();
$allAutoloadable = (new Reflector(new ComposerSourceLocator($classLoader))->getAllClasses();
```

### PR DESCRIPTION
**Pull request by asgrim**
# Initial Changes

* Refactored classes into directories now
* `new Reflector` now requires something implementing `SourceLocator` interface
* `SourceLocator` interface has a single method that should return a value object `LocatedSource`
* The `LocatedSource` VO is used to load the class requested internally by `Reflector`
* The public API from `Reflector` is now much simpler, and only has:
  * `$reflector->reflect($className)`
  * `$reflector->getClassesFromFile()` (only works when `FilenameSourceLocator`)
* Creates three `SourceLocator`-implementing classes:
  * `ComposerSourceLocator` - has a constructor dependency on `Composer\Autoload\ClassLoader`
  * `FilenameSourceLocator` - constructor dep is the filename as a string
  * `StringSourceLocator` - constructor dep is valid PHP code in a string
* `README.md` is updated to show example usage of creating the `ReflectionClass` instances
  * [Review the updated README](https://github.com/asgrim/better-reflection/blob/refactor-loading/README.md)
* `Reflector::compileNodeExpression` was moved to own class `NodeCompiler::compile` for easier refactoring later
* Renamed `MethodsTest` fixture as PHPUnit was loading the file

# Refactored changes

* Created a new `Symbol` VO that identifies a symbol with a name and type (checks types are valid)
* Refactored `Reflector` to use the `Symbol` to generically identify "symbols" (instead of just classes)
* Replaced `Reflector::getClassesFromFile` with `Reflector::getAllSymbols` to be more generic and also fixes #4 
* Created a new `BetterReflection\Reflection\Reflection` interface that identifies symbols that can be reflected (currently only `ReflectionClass`)
* Created a `ClassReflector` which simplfies the API to calling `$reflector->reflect(..)` so externally, users don't have to worry about creating `Symbol`s etc.. In future, we can create `FunctionReflector`, `TraitReflector` and so on.

# Todo

* [x] Refactor `SourceLocator` somehow to accomodate for other types of symbols:
  * classes
  * functions
  * traits
  * interfaces
  * constants

### CODE REVIEW

File: README.md [hunk 1]
```diff
@@ -4,18 +4,52 @@ Better Reflection
...
 echo $reflectionClass->getName(); // Foo\Bar\MyClass
 echo $reflectionClass->getNamespaceName(); // Foo\Bar
 ```
+
+Example usage for loading a class from a specific file:
+
+```php
+<?php
+
+$reflector = new Reflector(new FilenameSourceLocator('path/to/MyApp/MyClass.php'));
```
Ocramius: As per discussion, `FilenameSourceLocator` should probably be something like `InSingleFileSourceLocator`

README.md
```diff
@@ -4,18 +4,52 @@ Better Reflection
...
+echo $reflectionClass->getShortName(); // Foo
+```
+
+Example usage to fetch a list of classes from a file
+
+```php
+<?php
+
+$reflector = new Reflector(new FilenameSourceLocator('path/to/file.php'));
+$classes = $reflector->getClassesFromFile();
```
Ocramius: `getClassesFromFile` without parameter? Seems a bit weird. The API is probably also redundant, now that you have the locator:

```php
$allInString = (new Reflector(new FilenameSourceLocator('path/to/file.php'))->getAllClasses();
$allInFile = (new Reflector(new StringSourceLocator($code))->getAllClasses();
$allAutoloadable = (new Reflector(new ComposerSourceLocator($classLoader))->getAllClasses();
```

src/NodeCompiler/NodeCompiler.php
```diff
@@ -0,0 +1,43 @@
...
+     * Compile an expression from a node into a value
+     *
+     * @param Node $node
+     * @return mixed
+     */
+    public static function compile(Node $node)
+    {
+        $type = get_class($node);
+
+        switch ($type) {
```
Ocramius: `switch (get_class($node)) {` works too

README.md
asgrim: Created issue #4 to implement this later

src/SourceLocator/SourceLocator.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+namespace Asgrim\SourceLocator;
+
+interface SourceLocator
```
Ocramius: Should be renamed to `ClassLocator`

src/SourceLocator/SourceLocator.php
```diff
@@ -0,0 +1,16 @@
...
+{
+    /**
+     * Locate some source code
+     *
+     * This method should return a LocatedSource value object
+     *
+     * @param string $className
+     * @return LocatedSource
+     */
+    public function locate($className);
```
Ocramius: Can be `__invoke`, since it's a single method.

src/SourceLocator/SourceLocator.php
asgrim: These things are not actually locating a class as such, it is actually locating source code (the `locate` method should return a `LocatedSource` VO which contains just source code + filename... it is the `Reflector` which is loading that source code and finding out if the class actually exists within that source code... so IMO the name `SourceLocator` is actually more accurate (unless you can suggest something _more_ appropriate)

Ocramius: @asgrim still states `$className` there.

What if I want to locate from a function name?

asgrim: Actually very good point. You're locating the source code that "should" contain the class. Similarly, a new interface should be created for functions. So propose:

```php
<?php
interface ClassSourceLocator {
  public function __invoke($className);
}
interface FunctionSourceLocator { // for future
  public function __invoke($functionName);
}

class ClassReflector { // was Reflector
  public function __construct(ClassSourceLocator $sourceLocator);
  public function reflect($className);
}
class FunctionReflector { // for future
  public function __construct(FunctionSourceLocator $sourceLocator);
  public function reflect($functionName);
}
```

Happy to discuss other options, just an idea :)

Ocramius: @asgrim other symbols that can be defined are:
 - constants
 - traits
 - interfaces

Not sure if a single locator has to deal with all of this or not

asgrim: Hmm, I think a further refactor should be done to account for this :D

README.md
```diff
@@ -4,18 +4,54 @@ Better Reflection
...
 echo $reflectionClass->getNamespaceName(); // Foo\Bar
 ```
+
+Example usage for loading a class from a specific file:
+
+```php
+<?php
+
+$reflector = new ClassReflector(new SingleFileSourceLocator('path/to/MyApp/MyClass.php'));
+$reflectionClass = $reflector->reflect('MyApp\MyClass');
```
Ocramius: Can haz `::class`?

src/NodeCompiler/NodeCompiler.php
```diff
@@ -0,0 +1,45 @@
+<?php
+
+namespace BetterReflection\NodeCompiler;
+
+use PhpParser\Node;
+
+class NodeCompiler
```
Ocramius: Name is too generic.

src/NodeCompiler/NodeCompiler.php
```diff
@@ -0,0 +1,45 @@
...
+{
+    /**
+     * Compile an expression from a node into a value
+     *
+     * @param Node $node
+     * @return mixed
+     */
+    public static function compile(Node $node)
```
Ocramius: Make it an instance and eventually use `__invoke`? :)

src/Reflection/Reflection.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+namespace BetterReflection\Reflection;
+
+interface Reflection
```
Ocramius: Missing docblocks (overall)

Ocramius: If this is for internal use, use `@internal`

src/Reflection/ReflectionClass.php
```diff
@@ -88,8 +95,14 @@ public static function createFromNode(ClassNode $node, NamespaceNode $namespace
         return $class;
     }
 
+    public function getReflectionType()
+    {
+        return Symbol::SYMBOL_CLASS;
```
Ocramius: Use `__CLASS__` or equivalent

src/Reflection/ReflectionClass.php
```diff
@@ -127,13 +142,15 @@ public function getNamespaceName()
...
-     * Decide if this class is part of a namespace. Returns false if global namespace;
+     * Decide if this class is part of a namespace. Returns false if the class
+     * is in the global namespace or does not have a specified namespace
      *
      * @return bool
      */
     public function inNamespace()
     {
-        return !(is_null($this->declaringNamespace)) && !is_null($this->declaringNamespace->name);
+        return !(is_null($this->declaringNamespace))
```
Ocramius: Consider dropping `is_null()` overall ;-)

src/Reflection/ReflectionFunctionAbstract.php
```diff
@@ -97,13 +101,17 @@ public function getNumberOfParameters()
      */
     public function getNumberOfRequiredParameters()
     {
-        return count(array_filter($this->parameters, function (ReflectionParameter $p) {
-            return !$p->isOptional();
-        }));
+        return count(array_filter(
```
Ocramius: Is this API in line with the other optional parameters API you wrote before?
As far as I understand, in your case, given following method:

```php
public function foo($bar, $baz = null, $tab, $taz = null) {}
```
the number of optional parameters is 1

src/Reflection/ReflectionParameter.php
```diff
@@ -251,9 +261,14 @@ public function getDefaultValueAsString()
             case 'object':
             case 'resource':
             case 'unknown type':
-                throw new \RuntimeException(
-                    'Default value as an instance of an ' . $type . ' does not make any sense'
-                );
+            default:
+                $typeExceptionMessage = 'Default value as an instance of an %s'
```
Ocramius: Smash it in the exception?

src/Reflection/ReflectionParameter.php
```diff
@@ -251,9 +261,14 @@ public function getDefaultValueAsString()
...
+                    . ' does not make any sense';
+
+                throw new \RuntimeException(sprintf(
```
Ocramius: Use your own exception namespace

src/Reflection/Symbol.php
```diff
@@ -0,0 +1,83 @@
+<?php
+
+namespace BetterReflection\Reflection;
+
+use PhpParser\Node;
+
+class Symbol
+{
+    const SYMBOL_CLASS = 'class';
```
Ocramius: As suggested above, reference the actual classes here

src/Reflection/Symbol.php
```diff
@@ -0,0 +1,83 @@
...
+
+    /**
+     * @var string
+     */
+    private $type;
+
+    public function __construct($name, $type = self::SYMBOL_CLASS)
+    {
+        if ('\\' == $name[0]) {
+            $name = substr($name, 1);
```
Ocramius: `ltrim($name, '\\')`

src/Reflection/Symbol.php
```diff
@@ -0,0 +1,83 @@
...
+        }
+        $this->name = (string)$name;
```
Ocramius: Consider validating `$name` against the regex `'/([a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]+)(\\\\[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]+)*/'`, as that's what a valid class name can refer to.

Only one segment for functions/methods.

See https://github.com/Ocramius/ProxyManager/blob/7badedb5f8afaaccd496a096daadb19d7bf2b374/tests/ProxyManagerTest/Inflector/ClassNameInflectorTest.php#L127

src/Reflection/Symbol.php
```diff
@@ -0,0 +1,83 @@
...
+
+        if (!in_array($type, $this->validSymbols)) {
```
Ocramius: missing third parameter `true`

src/Reflection/Symbol.php
```diff
@@ -0,0 +1,83 @@
...
+     * @return bool
+     */
+    public function isLoaded()
+    {
+        return false;
+    }
+
+    public function isMatchingReflector(Reflection $reflector)
+    {
+        return $this->getType() == $reflector->getReflectionType();
```
Ocramius: No need for getter usage in here (you are in private scope)

Ocramius: Use a strict comparison

src/Reflector/Reflector.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+namespace BetterReflection\Reflector;
+
+interface Reflector
+{
+    public function reflect($symbolName);
```
Ocramius: Shouldn't this be `Symbol $symbolName`?

src/Reflector/Generic.php
```diff
@@ -0,0 +1,196 @@
...
+    /**
+     * Uses the SourceLocator given in the constructor to locate the $symbolName
+     * specified and returns the \Reflector
+     *
+     * @param Symbol $symbol
+     * @return Reflection
+     */
+    public function reflect($symbol)
+    {
+        if (!($symbol instanceof Symbol)) {
```
Ocramius: You are not checking for the symbol type. Did you consider splitting `ClassSymbol` and other symbols by type, rather than property?

Ocramius: In general, I'd be able to pass in any `Symbol` instance here: confusing

src/Reflector/ClassReflector.php
```diff
@@ -0,0 +1,38 @@
...
+     */
+    private $reflector;
+
+    public function __construct(SourceLocator $sourceLocator)
+    {
+        $this->reflector = new GenericReflector($sourceLocator);
+    }
+
+    /**
+     * @param $className
```
Ocramius: string?

src/Reflector/ClassReflector.php
```diff
@@ -0,0 +1,38 @@
...
+     * @return \BetterReflection\Reflection\ReflectionClass
+     */
+    public function reflect($className)
+    {
+        $symbol = new Symbol($className, Symbol::SYMBOL_CLASS);
```
Ocramius: As suggested above, what is being attempted here is typing via internal state - not really suggested IMO

Ocramius: Can inline with the next call, no need for a variable

src/Reflector/Generic.php
```diff
@@ -0,0 +1,196 @@
...
+        if ($symbol->isLoaded()) {
+            throw new \LogicException(sprintf(
+                '%s "%s" is already loaded',
+                $symbol->getDisplayType(),
+                $symbol->getName()
+            ));
+        }
+
+        $locatedSource = $this->sourceLocator->__invoke($symbol);
+        $reflection = $this->reflectFromLocatedSource($symbol, $locatedSource);
```
Ocramius: Inline with previous line

src/Reflector/Generic.php
```diff
@@ -0,0 +1,196 @@
...
+
+        return $reflection;
```
Ocramius: Return directly from the call

src/Reflector/Generic.php
```diff
@@ -0,0 +1,196 @@
...
+     * Given an array of Reflections, try to find the symbol
+     *
+     * @param Reflection[] $reflections
+     * @param Symbol $symbol
+     * @return Reflection
+     */
+    private function findInArray($reflections, Symbol $symbol)
+    {
+        foreach ($reflections as $reflection) {
+            if ($reflection->getName() == $symbol->getName()) {
```
Ocramius: Strict comparisons

src/Reflector/Generic.php
```diff
@@ -0,0 +1,196 @@
...
+     * @param Node[] $ast
+     * @param string|null $filename
+     * @param Symbol $symbol
+     * @return Reflection[]
+     */
+    private function reflectFromTree(array $ast, $filename, Symbol $symbol)
+    {
+        $reflections = [];
+        foreach ($ast as $node) {
+            switch (get_class($node)) {
```
Ocramius: This is a dangerous way of switching between node types, especially considering that they are not `final`.

Ocramius: Consider `if` + `instanceof` instead

src/Reflector/Generic.php
```diff
@@ -0,0 +1,196 @@
...
+     * Get an array of reflections found in a LocatedSource
+     *
+     * @param LocatedSource $locatedSource
+     * @param Symbol $symbol
+     * @return Reflection[]
+     */
+    private function getReflections(LocatedSource $locatedSource, Symbol $symbol)
+    {
+        $parser = new Parser(new Lexer);
+        $ast = $parser->parse($locatedSource->getSource());
```
Ocramius: Inline the call?

src/Reflector/Generic.php
```diff
@@ -0,0 +1,196 @@
...
+
+    /**
+     * Get all symbols of a matching symbol type from a file
+     *
+     * @param string $symbolType
+     * @return Reflection[]
+     */
+    public function getAllSymbols($symbolType)
+    {
+        $symbol = new Symbol('*', $symbolType);
```
Ocramius: `*` should be a class constant, not a magic value

src/Reflector/Reflector.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+namespace BetterReflection\Reflector;
+
+interface Reflector
```
Ocramius: Docblocks, overall.

src/SourceLocator/ComposerSourceLocator.php
```diff
@@ -0,0 +1,50 @@
+<?php
+
+namespace BetterReflection\SourceLocator;
+
+use Composer\Autoload\ClassLoader;
```
Ocramius: If you import this class, you are also to require `composer/composer` in the project, no?

Ocramius: Maybe it's just an extreme case tho, as it assumes the composer autoloader (fair assumption)

src/SourceLocator/LocatedSource.php
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * Value object containing source code that has been located
+ */
+class LocatedSource
+{
+    /**
+     * @var string
+     */
+    private $source;
```
Ocramius: What's the difference between `$source` and `$fileName`? Naming is unclear

Ocramius: Is it `resource` or `identifier`?

src/SourceLocator/LocatedSource.php
```diff
@@ -0,0 +1,47 @@
...
+            throw new \InvalidArgumentException(
+                'Filename must be a string or null'
+            );
+        }
+
+        $this->source = $source;
+        $this->filename = $filename;
+    }
+
+    public function getSource()
```
Ocramius: Docblocks

src/Reflection/ReflectionFunctionAbstract.php
asgrim: I had an idea to use this `getNumberOfRequiredParameters` to determine `isOptional`, until I discovered that that would cause cyclic dependency... which was annoying :(

Can always create an issue to resolve this later, but it works

src/Reflector/Reflector.php
asgrim: No, a reflector takes a string as the only parameter, the "Generic" reflector is not a reflector, but takes a `Symbol` as it's parameter

src/Reflector/Generic.php
asgrim: Actually, this was a hack because I was trying to make `Generic` implement `Reflector` - but it's not any more, so I'll hint it properly

src/Reflector/Generic.php
asgrim: it actually doesn't matter what it is - or at least in the case of SingleFileSourceLocator it didn't - I need to check this

Ocramius: So it is an over-abstraction?

src/SourceLocator/ComposerSourceLocator.php
asgrim: `suggests` perhaps?

src/SourceLocator/LocatedSource.php
asgrim: `source` is source code itself, `filename` is a potential filename (may be null actually, so will update the doc)

src/Reflector/Generic.php
```diff
@@ -0,0 +1,190 @@
...
+
+    /**
+     * Get all identifiers of a matching identifier type from a file
+     *
+     * @param IdentifierType $identifierType
+     * @return Reflection[]
+     */
+    public function getAllByIdentifierType(IdentifierType $identifierType)
+    {
+        $identifier = new Identifier('*', $identifierType);
```
asgrim: @Ocramius I've created issue #21 to deal with this janky identifier stuff in `getAllByIdentifierType`

src/Reflection/ReflectionFunctionAbstract.php
Ocramius: Created #22 to track this

PR #6, opened 2015-06-30
Move to use BetterReflection namespace

asgrim:


asgrim:
I don't know what's up with Travis, but tests do pass :P

Issue #7, opened 2015-06-30
Need to become "mostly" compatible with PHP's Reflection API


asgrim:
Compatibility is being tracked in [compatibility.md](https://github.com/Roave/BetterReflection/blob/master/docs/compatibility.md). We need to aim for all features it is *possible* to implement. Unfortunately we cannot implement "all" the functionality, for example:

* Extension-related things `ReflectionExtension` - I don't think we have a way to access the information required by this, nor to tie it all up with our userland reflection (I may be wrong?) - see #15 
* Methods `ReflectionMethod->invoke` as we are not actually loading the classes, we cannot really invoke methods (yes, it's _technically_ possible, but not logical) - see #14 
* State modifications like `ReflectionProperty->setAccessible` and `ReflectionProperty->setValue` these require instances of classes, but we never instantiate classes - see #14 
* Instance-requiring `ReflectionObject` simply because it requires an instance, we cannot implement it, because that would require the class to be loaded - see #16 
* `ReflectionFunctionAbstract->isClosure` can't be implemented currently - see #37 
* Internal reflection is not currently possible - see #38 

## Compatibility summary
* [x] ReflectionClass - PR #59
* [x] ReflectionFunctionAbstract - PR #40
* [x] ReflectionMethod - PR #58
* [x] ReflectionParameter - PR #56
* [x] ReflectionFunction - PR #40
* [x] ReflectionProperty - PR #9

PR #9, opened 2015-07-01
Compatibility for ReflectionParameter

asgrim:
As part of #7, this PR brings full compatibility with `ReflectionParameter` (where possible) 

* [x] isArray
* [x] isCallable
* [x] getPosition
* [x] isDefaultValueAvailable
* [x] canBePassedByValue
* [x] getClass
* [x] getDefaultValueConstantName
* [x] isDefaultValueConstant
* [x] isPassedByReference
* [x] isVariadic

### DIFF
--- a/docs/compatibility.md
+++ b/docs/compatibility.md
@@ -109,13 +109,13 @@
 | getDeclaringClass | :heavy_check_mark: Yes |
 | getDeclaringFunction | :heavy_check_mark: Yes |
 | getDefaultValue | :heavy_check_mark: Yes (*some assumptions are made) |
-| getDefaultValueConstantName | todo |
+| getDefaultValueConstantName | :heavy_check_mark: Yes |
 | getName | :heavy_check_mark: Yes |
 | getPosition | :heavy_check_mark: Yes |
 | isArray | :heavy_check_mark: Yes |
 | isCallable | :heavy_check_mark: Yes |
 | isDefaultValueAvailable | :heavy_check_mark: Yes |
-| isDefaultValueConstant | todo |
+| isDefaultValueConstant | :heavy_check_mark: Yes |
 | isOptional | :heavy_check_mark: Yes |
 | isPassedByReference | :heavy_check_mark: Yes |
 | isVariadic | :heavy_check_mark: Yes |
--- a/docs/compatibility.md
+++ b/docs/compatibility.md
@@ -111,7 +111,7 @@
 | getDefaultValue | :heavy_check_mark: Yes (*some assumptions are made) |
 | getDefaultValueConstantName | todo |
 | getName | :heavy_check_mark: Yes |
-| getPosition | todo |
+| getPosition | :heavy_check_mark: Yes |
 | isArray | todo |
 | isCallable | todo |
 | isDefaultValueAvailable | todo |
--- a/docs/compatibility.md
+++ b/docs/compatibility.md
@@ -114,11 +114,11 @@
 | getPosition | :heavy_check_mark: Yes |
 | isArray | :heavy_check_mark: Yes |
 | isCallable | :heavy_check_mark: Yes |
-| isDefaultValueAvailable | todo |
+| isDefaultValueAvailable | :heavy_check_mark: Yes |
 | isDefaultValueConstant | todo |
 | isOptional | :heavy_check_mark: Yes |
-| isPassedByReference | todo |
-| isVariadic | todo |
+| isPassedByReference | :heavy_check_mark: Yes |
+| isVariadic | :heavy_check_mark: Yes |
 
 ## ReflectionFunction
 
--- a/docs/compatibility.md
+++ b/docs/compatibility.md
@@ -104,8 +104,8 @@
 | Method | Supported |
 |--------|-----------|
 | allowsNull | :heavy_check_mark: Yes |
-| canBePassedByValue | todo |
-| getClass | todo |
+| canBePassedByValue | :heavy_check_mark: Yes |
+| getClass | :x: No - could currently cause recursion |
 | getDeclaringClass | :heavy_check_mark: Yes |
 | getDeclaringFunction | :heavy_check_mark: Yes |
 | getDefaultValue | :heavy_check_mark: Yes (*some assumptions are made) |
--- a/docs/compatibility.md
+++ b/docs/compatibility.md
@@ -112,8 +112,8 @@
 | getDefaultValueConstantName | todo |
 | getName | :heavy_check_mark: Yes |
 | getPosition | :heavy_check_mark: Yes |
-| isArray | todo |
-| isCallable | todo |
+| isArray | :heavy_check_mark: Yes |
+| isCallable | :heavy_check_mark: Yes |
 | isDefaultValueAvailable | todo |
 | isDefaultValueConstant | todo |
 | isOptional | :heavy_check_mark: Yes |
--- a/src/ReflectionFunctionAbstract.php
+++ b/src/ReflectionFunctionAbstract.php
@@ -48,6 +48,22 @@ abstract class ReflectionFunctionAbstract
             $this->docBlock = $comment->getReformattedText();
         }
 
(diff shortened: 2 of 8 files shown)


src/ReflectionParameter.php
```diff
@@ -213,4 +220,44 @@ public function getTypes()
...
+     */
+    public function getPosition()
+    {
+        return $this->parameterIndex;
+    }
+
+    /**
+     * Get the type hint declared for the parameter
+     *
+     * @return Type
```
Ocramius: If there is a composite type, do you return `Type[]` or `CompositeType`?

Ocramius: As per discussion, this returns the type hint defined in the language

docs/compatibility.md
```diff
@@ -104,21 +104,21 @@
 | Method | Supported |
 |--------|-----------|
 | allowsNull | :heavy_check_mark: Yes |
-| canBePassedByValue | todo |
-| getClass | todo |
+| canBePassedByValue | :heavy_check_mark: Yes |
+| getClass | :x: No - could currently cause recursion |
```
Ocramius: What's the problem with recursion here?

src/ReflectionFunctionAbstract.php
```diff
@@ -48,6 +48,22 @@ protected function populateFunctionAbstract(MethodNode $node, $filename)
             $this->docBlock = $comment->getReformattedText();
         }
 
+        // We must determine if params are optional or not ahead of time, but
+        // we must do it in reverse...
+        $overallOptionalFlag = true;
+        for ($i = count($node->params) - 1; $i >= 0; $i--) {
```
Ocramius: The `-1` at the end is confusing

src/ReflectionFunctionAbstract.php
```diff
@@ -48,6 +48,22 @@ protected function populateFunctionAbstract(MethodNode $node, $filename)
...
+            $hasDefault = !is_null($node->params[$i]->default);
```
Ocramius: use `null !==` rather than `is_null`

docs/compatibility.md
asgrim: We'd have to load a `ReflectionClass` instance of the class that is hinted in the language type hint, which could be the same class. Currently, because we are not autoloading classes, and the PR #3 is still WIP, we have no way of really loading a `ReflectionClass` for this, unless we used builtin `ReflectionClass`, but that kinda defeats the point. It's a mess that I need to have a think on...

src/ReflectionFunctionAbstract.php
```diff
@@ -48,6 +48,22 @@ protected function populateFunctionAbstract(MethodNode $node, $filename)
...
+
+            // When we find the first parameter that does not have a default,
+            // flip the flag as all params for this are no longer optional
+            // EVEN if they have a default value
```
Ocramius: This is counter-intuitive. Is the internal reflection also behaving like this?

src/ReflectionParameter.php
```diff
@@ -24,6 +29,11 @@ class ReflectionParameter implements \Reflector
     private $isOptional;
 
     /**
+     * @var bool
+     */
+    private $hasDefaultValue;
```
Ocramius: Set a default here rather than in the constructor?

src/ReflectionParameter.php
```diff
@@ -44,12 +54,35 @@ class ReflectionParameter implements \Reflector
     private $types;
 
     /**
+     * @var Type
+     */
+    private $typeHint;
```
Ocramius: Missing default value?

src/ReflectionParameter.php
```diff
@@ -44,12 +54,35 @@ class ReflectionParameter implements \Reflector
...
+
+    /**
      * @var int
      */
     private $parameterIndex;
 
+    /**
+     * @var bool
+     */
+    private $isDefaultValueConstant;
```
Ocramius: Set a default here rather than in the constructor?

src/ReflectionParameter.php
```diff
@@ -44,12 +54,35 @@ class ReflectionParameter implements \Reflector
...
+
+    /**
+     * @var string
+     */
+    private $defaultValueConstantName;
```
Ocramius: Set a default here rather than in the constructor?

src/ReflectionParameter.php
```diff
@@ -44,12 +54,35 @@ class ReflectionParameter implements \Reflector
...
+
+    /**
+     * @var int
+     */
+    private $defaultValueConstantType;
```
Ocramius: Set a default here rather than in the constructor?

src/ReflectionParameter.php
```diff
@@ -84,13 +119,28 @@ public static function createFromNode(ParamNode $node, ReflectionFunctionAbstrac
...
-        $param->parameterIndex = $parameterIndex;
+        $param->isOptional = (bool)$node->isOptional;
+        $param->hasDefaultValue = !is_null($node->default);
+        $param->isVariadic = (bool)$node->variadic;
+        $param->isByReference = (bool)$node->byRef;
+        $param->parameterIndex = (int)$parameterIndex;
+        $param->typeHint = TypesFinder::findTypeForAstType($node->type);
 
-        if ($param->isOptional) {
+        if ($param->hasDefaultValue) {
```
Ocramius: Split into a separate private method?

src/ReflectionParameter.php
asgrim: Don't need a default value, we can only instantiate `ReflectionParameter::createFromNode` which always sets `$typeHint` value

src/ReflectionParameter.php
asgrim: Don't need a default value, we can only instantiate `ReflectionParameter::createFromNode` which always sets `$hasDefaultValue` value

src/ReflectionFunctionAbstract.php
asgrim: Yes it's stupid, but it accounts for situations like this:

```php
function myFunc($foo = 'foo', $bar)
```

In this example, `isOptional()` on the `$foo` should return `false`, but `isDefaultValueAvailable()` should return `true`. AST doesn't give us information about whether params are "optional" or not, so we need to start with the last parameter, and work backwards, determining if each is optional.

asgrim: Actually, I've thought of a better way

docs/compatibility.md
Ocramius: @asgrim so... registry of parsed classes (so far)?

Ocramius:
```php
private function doLoadClass($class, & $currentlyLoaded = [])
{
    if (isset($currentlyLoaded[$class])) {
        return $currentlyLoaded[$class];
    }

    // ... possible recursion here.
}
```

src/ReflectionParameter.php
Ocramius: Then the docblock should state `bool|null`

src/ReflectionParameter.php
Ocramius: Same as above

src/ReflectionParameter.php
asgrim: Why? You can't instantiate the class without setting it to a `bool`, so it will never be anything but a `bool`...?

src/ReflectionParameter.php
asgrim: This should *actually* be `Type|null` :)

docs/compatibility.md
asgrim: I've created new issue #13 to investigate this - for now I'm happy to say it's currently unsupported and implement later.

src/ReflectionParameter.php
Ocramius: Oh, missed the static constructor.

Agreed :-)

docs/compatibility.md
Ocramius: :+1:

Issue #24, opened 2015-07-02
Improve the unit test suite further


asgrim:
* Improve what tests are running
* Improve coverage by writing more tests
* Use `@covers` annotations to limit cover leakage

# Acceptable coverage in:
* [x] Identifier (see #32) - 94%
* [x] NodeCompiler (see #32) - 82%
* [x] Reflection (see #65 and 59)
* [x] Reflector (see #65)
* [x] SourceLocator (see #32) - 97%
* [x] TypesFinder (see #28) - 100%

asgrim:
As of today: 
![coverage](https://cloud.githubusercontent.com/assets/496145/8499571/48c21c9e-218c-11e5-945a-5e4da60a5fcb.png)

asgrim:
![co](https://cloud.githubusercontent.com/assets/496145/8696473/7ce5c35c-2ae4-11e5-93e4-8889e38f909a.png)

and now

Ocramius:
@asgrim is the node compiler incorrectly covered?

asgrim:
@Ocramius no, it is the `@todo` and issue #19 (see PR #65)

PR #23, opened 2015-07-02
Added some more documentation

asgrim:

PR #25, opened 2015-07-03
Reorganised tests and added @covers annotations

asgrim:
~~Needs #23 merging first, then I need to rebase this.~~

Anyway, I added `@covers` everywhere, which now gives a realistic estimate of code coverage... around 40% line coverage.

This is the first step to achieving #24, and if merged will form the basis of writing the rest of the tests and getting coverage up a bit.


test/Reflection/ReflectionClassTest.php
```diff
@@ -18,6 +18,12 @@ private function getComposerLocator()
         return new ComposerSourceLocator($loader);
     }
 
+    /**
+     * @covers \BetterReflection\Reflection\ReflectionClass::inNamespace()
```
Ocramius: I suggest avoiding this sort of strict coverage and defining coverage only at class level: simpler/easier

asgrim: Why? Surely doing it this way means you specify the exact method so you
don't even have in-class coverage leaking. Each test is therefore
restricted to only covering the method we intended to test.

Ocramius: @asgrim if you add private methods, you start lagging back. The SUT is the class per-se in this sort of test, so I wouldn't go this far, mainly to avoid having to constantly update annotations as internal refactoring occurs

PR #32, opened 2015-07-03
Improve coverage #24

asgrim:
Link to issue #24

### DIFF
--- a/src/Identifier/IdentifierType.php
+++ b/src/Identifier/IdentifierType.php
@@ -53,6 +53,8 @@ class IdentifierType
             return $reflector instanceof ReflectionClass;
         }
 
+        // @todo add more type checks
+
         return false;
     }
 }
--- a/src/Identifier/IdentifierType.php
+++ b/src/Identifier/IdentifierType.php
@@ -42,14 +42,6 @@ class IdentifierType
     }
 
     /**
-     * @return string
-     */
-    public function getDisplayName()
-    {
-        return basename(str_replace('\\', '/', $this->name));
-    }
-
-    /**
      * Check to see if a reflector is of a valid type specified by this identifier
      *
      * @param Reflection $reflector
--- a/src/Identifier/IdentifierType.php
+++ b/src/Identifier/IdentifierType.php
@@ -46,11 +46,21 @@ class IdentifierType
      */
     public function getDisplayName()
     {
-        return ucfirst(basename($this->name));
+        return basename(str_replace('\\', '/', $this->name));
     }
 
+    /**
+     * Check to see if a reflector is of a valid type specified by this identifier
+     *
+     * @param Reflection $reflector
+     * @return bool
+     */
     public function isMatchingReflector(Reflection $reflector)
     {
-        return $this->name === get_class($reflector);
+        if ($this->name == self::IDENTIFIER_CLASS) {
+            return $reflector instanceof ReflectionClass;
+        }
+
+        return false;
     }
 }
--- a/test/unit/Identifier/IdentifierTest.php
+++ b/test/unit/Identifier/IdentifierTest.php
@@ -1,0 +1,24 @@
+<?php
+
+namespace BetterReflectionTest\Identifier;
+
+use BetterReflection\Identifier\Identifier;
+use BetterReflection\Identifier\IdentifierType;
+
+/**
+ * @covers \BetterReflection\Identifier\Identifier
+ */
+class IdentifierTest extends \PHPUnit_Framework_TestCase
+{
+    public function testValue()
+    {
+        $beforeName = '\Some\Thing\Here';
+        $afterName = 'Some\Thing\Here';
+
+        $identifierType = new IdentifierType(IdentifierType::IDENTIFIER_CLASS);
+        $identifier = new Identifier($beforeName, $identifierType);
+
+        $this->assertSame($afterName, $identifier->getName());
+        $this->assertSame($identifierType, $identifier->getType());
+    }
+}
--- a/test/unit/Identifier/IdentifierTest.php
+++ b/test/unit/Identifier/IdentifierTest.php
@@ -10,15 +10,20 @@ use BetterReflection\Identifier\IdentifierType;
  */
 class IdentifierTest extends \PHPUnit_Framework_TestCase
 {
-    public function testValue()
+    public function testGetName()
     {
         $beforeName = '\Some\Thing\Here';
         $afterName = 'Some\Thing\Here';
 
-        $identifierType = new IdentifierType(IdentifierType::IDENTIFIER_CLASS);
-        $identifier = new Identifier($beforeName, $identifierType);
-
+        $identifier = new Identifier($beforeName, new IdentifierType(IdentifierType::IDENTIFIER_CLASS));
         $this->assertSame($afterName, $identifier->getName());
+    }
+
+    public function testGetType()
+    {
+        $identifierType = new IdentifierType(IdentifierType::IDENTIFIER_CLASS);
+
+        $identifier = new Identifier('Foo', $identifierType);
         $this->assertSame($identifierType, $identifier->getType());
     }
 }
--- a/test/unit/Identifier/IdentifierTypeTest.php
+++ b/test/unit/Identifier/IdentifierTypeTest.php
@@ -17,20 +17,18 @@ class IdentifierTypeTest extends \PHPUnit_Framework_TestCase
     public function possibleIdentifierTypesProvider()
     {
         return [
-            [IdentifierType::IDENTIFIER_CLASS, 'ReflectionClass'],
+            [IdentifierType::IDENTIFIER_CLASS],
         ];
     }
 
     /**
      * @param string $full
-     * @param string $displayName
      * @dataProvider possibleIdentifierTypesProvider
      */
-    public function testPossibleIdentifierTypes($full, $displayName)
+    public function testPossibleIdentifierTypes($full)
     {
         $type = new IdentifierType($full);
         $this->assertSame($full, $type->getName());
-        $this->assertSame($displayName, $type->getDisplayName());
     }
 
     public function testThrowsAnExceptionWhenInvalidTypeGiven()
--- a/test/unit/Identifier/IdentifierTypeTest.php
+++ b/test/unit/Identifier/IdentifierTypeTest.php
@@ -11,6 +11,9 @@ use InvalidArgumentException;
  */
 class IdentifierTypeTest extends \PHPUnit_Framework_TestCase
 {
+    /**
+     * @return string[][]
+     */
     public function possibleIdentifierTypesProvider()
     {
         return [
--- a/test/unit/Identifier/IdentifierTypeTest.php
+++ b/test/unit/Identifier/IdentifierTypeTest.php
@@ -1,0 +1,70 @@
+<?php
+
+namespace BetterReflectionTest\Identifier;
+
+use BetterReflection\Identifier\IdentifierType;
+use BetterReflection\Reflection\ReflectionClass;
+use InvalidArgumentException;
+
+/**
+ * @covers \BetterReflection\Identifier\IdentifierType
+ */
+class IdentifierTypeTest extends \PHPUnit_Framework_TestCase
+{
+    public function possibleIdentifierTypesProvider()
+    {
+        return [
+            [IdentifierType::IDENTIFIER_CLASS, 'ReflectionClass'],
+        ];
+    }
+
+    /**
+     * @param string $full
+     * @param string $displayName
+     * @dataProvider possibleIdentifierTypesProvider
+     */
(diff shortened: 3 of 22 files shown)


asgrim:
Note to @Ocramius: coverage is not complete with this PR, it will just help if we get progress so far merged in. We'll leave #24 open to continue to track coverage improvements

src/Identifier/IdentifierType.php
```diff
@@ -46,11 +46,21 @@ public function getName()
      */
     public function getDisplayName()
     {
-        return ucfirst(basename($this->name));
+        return basename(str_replace('\\', '/', $this->name));
```
Ocramius: What is the reason for this change?

asgrim: because it didn't work before, writing the test discovered it did nothing ;)

src/Identifier/IdentifierType.php
```diff
@@ -46,11 +46,21 @@ public function getName()
...
+    /**
+     * Check to see if a reflector is of a valid type specified by this identifier
+     *
+     * @param Reflection $reflector
+     * @return bool
+     */
     public function isMatchingReflector(Reflection $reflector)
     {
-        return $this->name === get_class($reflector);
+        if ($this->name == self::IDENTIFIER_CLASS) {
```
Ocramius: Confusing logic. What about other types of reflection?

src/Identifier/IdentifierType.php
```diff
@@ -46,11 +46,21 @@ public function getName()
...
+            return $reflector instanceof ReflectionClass;
+        }
+
+        return false;
```
Ocramius: This code block can be collapsed into a single boolean expression

src/Identifier/IdentifierType.php
asgrim: They will be added as and when they are supported (for example, in #40)

src/Identifier/IdentifierType.php
Ocramius: So what is being done here? Is it just a user-friendly identifier?

src/Identifier/IdentifierType.php
asgrim: It will be added to in #40 so no need for now

test/unit/Identifier/IdentifierTest.php
```diff
@@ -0,0 +1,24 @@
...
+
+use BetterReflection\Identifier\Identifier;
+use BetterReflection\Identifier\IdentifierType;
+
+/**
+ * @covers \BetterReflection\Identifier\Identifier
+ */
+class IdentifierTest extends \PHPUnit_Framework_TestCase
+{
+    public function testValue()
```
Ocramius: Consider splitting into separate methods for the separate public APIs being used

test/unit/Identifier/IdentifierTypeTest.php
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * @covers \BetterReflection\Identifier\IdentifierType
+ */
+class IdentifierTypeTest extends \PHPUnit_Framework_TestCase
+{
+    public function possibleIdentifierTypesProvider()
+    {
+        return [
+            [IdentifierType::IDENTIFIER_CLASS, 'ReflectionClass'],
```
Ocramius: `::class` please :)

src/Identifier/IdentifierType.php
asgrim: yes, exactly, it turns `\BetterReflection\Reflection\ReflectionClass` into just `ReflectionClass` for example

test/unit/Identifier/IdentifierTypeTest.php
```diff
@@ -0,0 +1,70 @@
...
+use BetterReflection\Identifier\IdentifierType;
+use BetterReflection\Reflection\ReflectionClass;
+use InvalidArgumentException;
+
+/**
+ * @covers \BetterReflection\Identifier\IdentifierType
+ */
+class IdentifierTypeTest extends \PHPUnit_Framework_TestCase
+{
+    public function possibleIdentifierTypesProvider()
```
Ocramius: Docblock

test/unit/NodeCompiler/CompileNodeToValueTest.php
```diff
@@ -0,0 +1,48 @@
...
+
+use BetterReflection\NodeCompiler\CompileNodeToValue;
+use LogicException;
+use PhpParser\Lexer;
+use PhpParser\Node\Expr\Yield_;
+use PhpParser\Parser;
+
+class CompileNodeToValueTest extends \PHPUnit_Framework_TestCase
+{
+    public function nodeProvider()
```
Ocramius: Docblock

test/unit/Identifier/IdentifierTypeTest.php
asgrim: this is not the same as `\ReflectionClass`, so is not actually a class

src/Identifier/IdentifierType.php
Ocramius: @asgrim add a `@TODO` then, please

test/unit/SourceLocator/StringSourceLocatorTest.php
```diff
@@ -0,0 +1,35 @@
...
+        ));
+
+        $this->assertSame($sourceCode, $locatedSource->getSource());
+        $this->assertNull($locatedSource->getFileName());
+    }
+
+    public function testConstructorThrowsExceptionIfEmptyStringGiven()
+    {
+        $this->setExpectedException(InvalidArgumentException::class);
+        new StringSourceLocator('');
```
Ocramius: This seems to be invalid to me. Why is an empty string considered invalid source?

test/unit/SourceLocator/SingleFileSourceLocatorTest.php
```diff
@@ -0,0 +1,36 @@
...
+        ));
+
+        $this->assertSame($expectedContent, $locatedSource->getSource());
+        $this->assertSame($fileName, $locatedSource->getFileName());
+    }
+
+    public function testConstructorThrowsExceptionIfEmptyFileGiven()
+    {
+        $this->setExpectedException(InvalidArgumentException::class);
+        new SingleFileSourceLocator('');
```
Ocramius: More like "invalid file location"

test/unit/SourceLocator/LocatedSourceTest.php
```diff
@@ -0,0 +1,54 @@
...
+
+        $this->assertSame($source, $locatedSource->getSource());
+        $this->assertSame($file, $locatedSource->getFileName());
+    }
+
+    public function testValuesWithNullFilename()
+    {
+        $source = '<?php echo "Hello world";';
+        $file = null;
+        $locatedSource = new LocatedSource($source, $file);
```
Ocramius: Is this for `eval()`'d code?

test/unit/SourceLocator/ComposerSourceLocatorTest.php
```diff
@@ -0,0 +1,67 @@
...
+            new IdentifierType(IdentifierType::IDENTIFIER_CLASS)
+        ));
+
+        $this->assertSame($expectedContent, $locatedSource->getSource());
+        $this->assertSame($fileName, $locatedSource->getFileName());
+    }
+
+    public function testInvokableThrowsExceptionWhenClassNotResolved()
+    {
+        $className = 'ClassWithNoNamespace';
```
Ocramius: `::class`?

src/Identifier/IdentifierType.php
Ocramius: Maybe rename the method to make that clear?

src/Identifier/IdentifierType.php
Ocramius: @asgrim add a `@TODO` then, please

test/unit/SourceLocator/ComposerSourceLocatorTest.php
asgrim: Using `::class` here would load the class, which we do not want

test/unit/SourceLocator/StringSourceLocatorTest.php
asgrim: There will be nothing in the AST if we load an empty string, so there is no point in even letting it past this initial point

test/unit/SourceLocator/LocatedSourceTest.php
asgrim: Noooooo it does not get `eval()`'d, it gets loaded directly into the AST (rather than being loaded from a file)

test/unit/SourceLocator/StringSourceLocatorTest.php
Ocramius: An empty string is still valid PHP source: it simply has no effect

test/unit/SourceLocator/LocatedSourceTest.php
Ocramius: I'm confused by the meaning of `LocatedSource` then. Why is `LocatedSource('contents', null)` valid? What's the use-case scenario?

test/unit/SourceLocator/ComposerSourceLocatorTest.php
Ocramius: Won't load, as per discussion. `::class` is a compile-time resolved expression that won't cause runtime side-effects

src/Identifier/IdentifierType.php
asgrim: I've removed this method, probably not required anyway

test/unit/Identifier/IdentifierTest.php
asgrim: Done

test/unit/Identifier/IdentifierTypeTest.php
asgrim: Done

test/unit/NodeCompiler/CompileNodeToValueTest.php
asgrim: Done

test/unit/SourceLocator/SingleFileSourceLocatorTest.php
asgrim: Added custom exception type

test/unit/SourceLocator/ComposerSourceLocatorTest.php
asgrim: Fixed

test/unit/SourceLocator/LocatedSourceTest.php
asgrim: For when you want to load source code directly from a string - the `StringSourceLocator` does exactly this, but the filename is allowed to be `null` (because you are loading from a string, rather than a file). It just means when you reflect from a class within a string, you'll not have a filename when you call `$classInfo->getFileName()` but then that would be expected (because you never loaded from file)

test/unit/SourceLocator/StringSourceLocatorTest.php
asgrim: Added custom exception and comment to hopefully clarify

test/unit/SourceLocator/LocatedSourceTest.php
Ocramius: @asgrim so why not using a  `StringSourceLocator` instead?

asgrim: Because this is a unit test, I'm just checking that the values are what we expect them to be when creating a `LocatedSource` VO

Ocramius: @asgrim still don't understand why `LocatedSource('content', null)` is valid: this means that the source was not "located", so the entire definition is borked IMO

asgrim: Created issue #43 and `@todo` comment

test/unit/SourceLocator/StringSourceLocatorTest.php
Ocramius: Still think this invalid. Should I report a separate issue?

asgrim: It would not make any sense to accept an empty string as source. Otherwise, nothing else will stop you doing it, and it just won't find the thing you're trying to reflect, and you'll be like "WHY DOES MY THING NOT FIND REFLECTION" and 3 hours later you find it's because you accidentally passed an empty string to the `StringSourceLocator` because of a misnamed variable or something, and then you'll be like



However, I did forget to expect `EmptyPhpSourceCode` not `InvalidArgumentException` so I've pushed again.

Ocramius: Makes sense for now, but I will hit this use-case scenario when traversing a directory that contains empty files :-)

Issue #30, opened 2015-07-03
Create context in FindTypeFromAst type finder


asgrim:
At the moment, no context happens in the `FindTypeFromAst` type finder.

This needs to be implemented!

asgrim:
Fixed via #46

Issue #29, opened 2015-07-03
Create context in FindParameterType type finder


asgrim:
At the moment, no context happens in the `FindParameterType` type finder.

This needs to be implemented!

asgrim:
Closed via #45

PR #33, opened 2015-07-03
ResolveTypes invokable now requires Context always

### LINKED ISSUE #31: Context should be required parameter in ResolveTypes type finder
Currently the signature is:

```php
public function __invoke($stringTypes, Context $context = null)
```

However `$context` should not be optional, so change to:

```php
public function __invoke($stringTypes, Context $context)
```

However, this implementation depends on #29 and #30 before we can remove the optional-ness.

asgrim:
Fixes #31. I managed to fix this without requiring #29 and #30 to be done by using a dummy context (which is what phpDocumentor does internally), which means we can fix the API early, and implement #29 and #30 independently (once this is merged at least).

### DIFF
--- a/src/TypesFinder/FindParameterType.php
+++ b/src/TypesFinder/FindParameterType.php
@@ -2,6 +2,7 @@
 
 namespace BetterReflection\TypesFinder;
 
+use phpDocumentor\Reflection\Types\Context;
 use PhpParser\Node\Param as ParamNode;
 use phpDocumentor\Reflection\DocBlock;
 use phpDocumentor\Reflection\Type;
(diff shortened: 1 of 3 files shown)


src/TypesFinder/FindParameterType.php
```diff
@@ -26,7 +27,7 @@ public function __invoke(ReflectionFunctionAbstract $function, ParamNode $node)
         foreach ($paramTags as $paramTag) {
             /* @var $paramTag \phpDocumentor\Reflection\DocBlock\Tag\ParamTag */
             if ($paramTag->getVariableName() === '$' . $node->name) {
-                return (new ResolveTypes())->__invoke($paramTag->getTypes());
+                return (new ResolveTypes())->__invoke($paramTag->getTypes(), new Context(''));
```
Ocramius: Empty context? Add a `@TODO comment`

src/TypesFinder/FindTypeFromAst.php
```diff
@@ -27,7 +28,7 @@ public function __invoke($astType)
             return null;
         }
 
-        $types = (new ResolveTypes())->__invoke([$typeString]);
+        $types = (new ResolveTypes())->__invoke([$typeString], new Context(''));
```
Ocramius: same

asgrim:
@Ocramius I have added `@todo`'s

Ocramius:
:+1:

PR #40, opened 2015-07-05
Compatibility for ReflectionFunctionAbstract and ReflectionFunction

asgrim:
~~Depends on #32 being merged first, possibly rebase after etc.~~

This introduces compatibility for #7 items `ReflectionFunction` and `ReflectionFunctionAbstract` where possible. Some items are not implement, but marked accordingly in the compatibility list (with corresponding GitHub issues)

PR #41, opened 2015-07-06
Add AutoloadSourceLocator enabling ReflectionClass and ReflectionFunction instantiation

### LINKED ISSUE #36: Implement "autoloading" source locator
In order to automatically reflect using the exact same API as the core `Reflection*` classes, we need to be able to essentially ignore the whole `SourceLocator` concept, despite it's flexibility.

So, I think it's possible to implement a new `AutoloadingSourceLocator` that takes no constructor dependencies, which, funnily enough, would probably need to rely on the internal `ReflectionClass` API, e.g.:

```php
// if it's a class we're loading...
$reflectionClass = new \ReflectionClass($identifier->getName());
$filename = $reflectionClass->getFileName();
$source = file_get_contents($filename);
// use these to make the LocatedSource...
```

Additionally, we'd need to create a "static" reflector, which would be something like:

```php
class StaticReflector
{
    public static function reflect(...)
    {
        return (new Reflector(new AutoloadingSourceLocator()))->reflect($identifier);
    }
}
```

Then we can use this reflector in the constructors or something to reflect (still a bit hazy on how to actually implement this). The downside of this is it means we do have to LOAD the classes. So I also propose making a `LoadSafeSourceLocator` interface, which doesn't have any methods, but purely indicates that it "by design" does NOT load up classes into memory. Then in the `..\Reflector\Generic` class, we can check if `$reflection implements LoadSafeSourceLocator`, and ensure that the class is not loaded beforehand, and is not loaded after reflection, to ensure the contract is obeyed.

### TL;DR

* Create new `AutoloadingSourceLocator` that uses core refleciton API to locate source code filename
* Cleanup constructors so they are empty
* Create `StaticReflector` that uses `AutoloadingSourceLocator`
* Update `Reflection*` classes to use `AutoloadingSourceLocator`
* Create `LoadSafeSourceLocator` interface, any source locator that implements this interface MUST ensure the class does not get loaded ever. These would be the `Composer`, `SingleFile` and `String` source locators, as these do not load file. `AutoloadingSourceLocator` would load the file, so MUST NOT implement the `LoadSafeSourceLocator` interface.
* This allows us to do the following (i.e. replicate almost _exactly_ the core reflection API instantiation technique):

```php
$reflectionClass = new \BetterReflection\Reflection\ReflectionClass('My\Foo\Bar');
```

asgrim:
Implemented the `AutoloadSourceLocator` which means we can now use "static" creation of `ReflectionClass` objects, e.g.:

```php
$reflection = ReflectionClass::createFromName('BetterReflectionTest\Fixture\ExampleClass');
```

There are caveats (we expect the autoloader to load classes from a file, i.e. using require/include), but it does actually work.

~~It also means this can actually implement the `LoadSafeSourceLocator` (to be implemented)~~

Merging this would fix #36 and would be a better solution that I initially proposed actually :grinning: 

### Todo

* [x] Rebase onto `master` when #40 is merged
* [x] Complete implementation for `ReflectionFunction`
* [x] Use static constructor not dodgy `__construct` nullable nonsense
* ~~Create a `LoadSafeSourceLocator` and write the logic to do checking~~
* ~~Implement the @todo in [Identifier.php#L45](https://github.com/Roave/BetterReflection/blob/master/src/Identifier/Identifier.php#L45)~~

### DIFF
--- a/src/Reflection/ReflectionClass.php
+++ b/src/Reflection/ReflectionClass.php
@@ -3,6 +3,8 @@
 namespace BetterReflection\Reflection;
 
 use BetterReflection\NodeCompiler\CompileNodeToValue;
+use BetterReflection\Reflector\ClassReflector;
+use BetterReflection\SourceLocator\AutoloadSourceLocator;
 use PhpParser\Node\Stmt\Namespace_ as NamespaceNode;
 use PhpParser\Node\Stmt\Class_ as ClassNode;
 use PhpParser\Node\Stmt\ClassConst as ConstNode;
@@ -18,34 +20,39 @@ class ReflectionClass implements Reflection
     /**
      * @var NamespaceNode
      */
-    private $declaringNamespace;
+    private $declaringNamespace = null;
 
     /**
      * @var ReflectionMethod[]
      */
-    private $methods;
+    private $methods = [];
 
     /**
      * @var mixed[]
      */
-    private $constants;
+    private $constants = [];
 
     /**
      * @var ReflectionProperty[]
      */
-    private $properties;
+    private $properties = [];
 
     /**
      * @var string
      */
     private $filename;
 
-    private function __construct()
+    public function __construct($className = null)
     {
-        $this->declaringNamespace = null;
-        $this->methods = [];
-        $this->constants = [];
-        $this->properties = [];
+        if ($className) {
+            $reflection = (new ClassReflector(new AutoloadSourceLocator()))->reflect($className);
+            $this->name = $reflection->name;
+            $this->declaringNamespace = $reflection->declaringNamespace;
+            $this->methods = $reflection->methods;
+            $this->constants = $reflection->constants;
+            $this->properties = $reflection->properties;
+            $this->filename = $reflection->filename;
+        }
     }
 
     /**
--- a/src/Reflection/ReflectionClass.php
+++ b/src/Reflection/ReflectionClass.php
@@ -42,17 +42,13 @@ class ReflectionClass implements Reflection
      */
     private $filename;
 
-    public function __construct($className = null)
+    private function __construct()
     {
-        if ($className) {
-            $reflection = (new ClassReflector(new AutoloadSourceLocator()))->reflect($className);
-            $this->name = $reflection->name;
-            $this->declaringNamespace = $reflection->declaringNamespace;
-            $this->methods = $reflection->methods;
-            $this->constants = $reflection->constants;
-            $this->properties = $reflection->properties;
-            $this->filename = $reflection->filename;
-        }
+    }
+
+    public static function createFromName($className)
+    {
+        return (new ClassReflector(new AutoloadSourceLocator()))->reflect($className);
     }
 
     /**
--- a/src/Reflection/ReflectionFunction.php
+++ b/src/Reflection/ReflectionFunction.php
@@ -2,11 +2,26 @@
 
 namespace BetterReflection\Reflection;
 
+use BetterReflection\Reflector\FunctionReflector;
+use BetterReflection\SourceLocator\AutoloadSourceLocator;
 use PhpParser\Node\Stmt\Function_ as FunctionNode;
 use PhpParser\Node\Stmt\Namespace_ as NamespaceNode;
 
 class ReflectionFunction extends ReflectionFunctionAbstract implements Reflection
 {
+    public function __construct($functionName = null)
+    {
+        if ($functionName) {
+            $reflection = (new FunctionReflector(new AutoloadSourceLocator()))->reflect($functionName);
+            $this->parameters = $reflection->parameters;
+            $this->name = $reflection->name;
+            $this->declaringNamespace = $reflection->declaringNamespace;
+            $this->docBlock = $reflection->docBlock;
+            $this->filename = $reflection->filename;
+            $this->node = $reflection->node;
+        }
+    }
+
     /**
      * @param FunctionNode $node
      * @param NamespaceNode|null $namespaceNode
@@ -18,7 +33,7 @@ class ReflectionFunction extends ReflectionFunctionAbstract implements Reflectio
         NamespaceNode $namespaceNode = null,
         $filename = null
     ) {
-        $method = new self($node);
+        $method = new self();
 
         $method->populateFunctionAbstract($node, $namespaceNode, $filename);
 
--- a/src/Reflection/ReflectionFunction.php
+++ b/src/Reflection/ReflectionFunction.php
@@ -9,35 +9,31 @@ use PhpParser\Node\Stmt\Namespace_ as NamespaceNode;
 
 class ReflectionFunction extends ReflectionFunctionAbstract implements Reflection
 {
-    public function __construct($functionName = null)
+    /**
+     * @param string $functionName
+     * @return ReflectionFunction
+     */
+    public static function createFromName($functionName)
     {
-        if ($functionName) {
-            $reflection = (new FunctionReflector(new AutoloadSourceLocator()))->reflect($functionName);
-            $this->parameters = $reflection->parameters;
-            $this->name = $reflection->name;
-            $this->declaringNamespace = $reflection->declaringNamespace;
-            $this->docBlock = $reflection->docBlock;
-            $this->filename = $reflection->filename;
-            $this->node = $reflection->node;
-        }
+        return (new FunctionReflector(new AutoloadSourceLocator()))->reflect($functionName);
     }
 
     /**
      * @param FunctionNode $node
      * @param NamespaceNode|null $namespaceNode
      * @param string|null $filename
-     * @return ReflectionMethod
+     * @return ReflectionFunction
      */
     public static function createFromNode(
         FunctionNode $node,
         NamespaceNode $namespaceNode = null,
         $filename = null
     ) {
-        $method = new self();
+        $function = new self();
 
-        $method->populateFunctionAbstract($node, $namespaceNode, $filename);
+        $function->populateFunctionAbstract($node, $namespaceNode, $filename);
 
-        return $method;
+        return $function;
     }
 
     /**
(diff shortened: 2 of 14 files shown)


src/Reflection/ReflectionClass.php
```diff
@@ -18,34 +20,39 @@ class ReflectionClass implements Reflection
...
     private $filename;
 
-    private function __construct()
+    public function __construct($className = null)
     {
-        $this->declaringNamespace = null;
-        $this->methods = [];
-        $this->constants = [];
-        $this->properties = [];
+        if ($className) {
```
Ocramius: Why is `null` allowed ?

src/Reflection/ReflectionFunction.php
```diff
@@ -2,11 +2,26 @@
...
+use BetterReflection\Reflector\FunctionReflector;
+use BetterReflection\SourceLocator\AutoloadSourceLocator;
 use PhpParser\Node\Stmt\Function_ as FunctionNode;
 use PhpParser\Node\Stmt\Namespace_ as NamespaceNode;
 
 class ReflectionFunction extends ReflectionFunctionAbstract implements Reflection
 {
+    public function __construct($functionName = null)
+    {
+        if ($functionName) {
```
Ocramius: Same as above

src/SourceLocator/AutoloadSourceLocator.php
```diff
@@ -0,0 +1,103 @@
...
+ */
+class AutoloadSourceLocator implements SourceLocator
+{
+    private static $autoloadLocatedFile;
+
+    public function __invoke(Identifier $identifier)
+    {
+        self::$autoloadLocatedFile = null;
+
+        if ($identifier->getType()->getName() == IdentifierType::IDENTIFIER_CLASS) {
```
Ocramius: This logic is hard to follow: split into private methods please

src/SourceLocator/AutoloadSourceLocator.php
```diff
@@ -0,0 +1,103 @@
...
+     * Our wrapper simply records which file we tried to load and returns
+     * boolean false indicating failure
+     *
+     * @param string $path
+     * @param string $mode
+     * @param int $options
+     * @param string $opened_path
+     * @return bool
+     */
+    public function stream_open($path, $mode, $options, &$opened_path)
```
Ocramius: <3 PHP internals non-interfaced stream wrappers' crap.

src/SourceLocator/AutoloadSourceLocator.php
```diff
@@ -0,0 +1,103 @@
...
+            'rdev' => 0,
+            'size' => 1,
+            'atime' => time(),
+            'mtime' => time(),
+            'ctime' => time(),
+            'blksize' => 4096,
+            'blocks' => 8,
+        ];
+
+        $x = array_merge(array_values($assoc), $assoc);
```
Ocramius: Return directly

src/Reflection/ReflectionClass.php
asgrim: Because that is how `::createFromNode` works.

doing this:

```php
$classInfo = new ReflectionClass('Foo');
```

creates autoload source locator > creates class reflector > calls reflect > locates class > loads source code > loads AST > calls `ReflectionClass::createFromNode()` > which instantiates a new `ReflectionClass` but with no parameters

src/Reflection/ReflectionFunction.php
asgrim: Same as above

src/SourceLocator/AutoloadSourceLocator.php
asgrim: I do not :P

src/SourceLocator/AutoloadSourceLocator.php
asgrim: ah yes, left this in for testing :+1:

src/Reflection/ReflectionClass.php
Ocramius: @asgrim you are changing public API to please package API? Seems wrong :-\

asgrim: I don't see another way around it, ideally I need two constructors.. one for internal (`::createFromNode`) use and one for public API :/

src/SourceLocator/AutoloadSourceLocator.php
asgrim: Refactored this, hopefully much nicer

src/SourceLocator/AutoloadSourceLocator.php
```diff
@@ -0,0 +1,147 @@
...
+
+    /**
+     * Attempts to locate the specified identifier and store the result in
+     * self::$autoloadLocatedFile
+     *
+     * @param Identifier $identifier
+     */
+    private function locateIdentifier(Identifier $identifier)
+    {
+        if ($identifier->getType()->getName() == IdentifierType::IDENTIFIER_CLASS) {
```
Ocramius: Cache the call to `$identifier->getType()->getName()`. You should also probably make a method that matches the constant, rather than getting the constant and then comparing outside the VO

Ocramius: Something like `$type->isClass()`

src/SourceLocator/AutoloadSourceLocator.php
```diff
@@ -0,0 +1,147 @@
...
+     * error handler temporarily.
+     *
+     * @param string $className#
+     */
+    private function locateClassByName($className)
+    {
+        if (class_exists($className, false)) {
+            $reflection = new \ReflectionClass($className);
+            self::$autoloadLocatedFile = $reflection->getFileName();
+        } else {
```
Ocramius: Consider returning early

src/SourceLocator/AutoloadSourceLocator.php
```diff
@@ -0,0 +1,147 @@
...
+     * have function autoloading. Therefore if it exists, we simply use the
+     * internal reflection API to find the filename. If it doesn't we can do
+     * nothing so throw an exception.
+     *
+     * @param string $functionName
+     * @throws Exception\FunctionUndefined
+     */
+    private function locateFunctionByName($functionName)
+    {
+        if (function_exists($functionName)) {
```
Ocramius: Reverse this conditional to throw in it, and return outside it

src/SourceLocator/AutoloadSourceLocator.php
Ocramius: Consider putting `@link` to the stream wrapper API pages here

src/SourceLocator/AutoloadSourceLocator.php
```diff
@@ -0,0 +1,147 @@
...
+
+/**
+ * Use PHP's built in autoloader to locate a class, without actually loading.
+ *
+ * There are some prerequisites...
+ *   - we expect the autoloader to load classes from a file (i.e. using require/include)
+ */
+class AutoloadSourceLocator implements SourceLocator
+{
+    private static $autoloadLocatedFile;
```
Ocramius: Docblock

src/SourceLocator/AutoloadSourceLocator.php
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    /**
+     * Must be implemented to return some data so that calls like is_file will work
+     *
+     * @param $path
+     * @param $flags
+     * @return mixed[]
+     */
+    public function url_stat($path, $flags)
```
Ocramius: Consider putting `@link` to the stream wrapper API pages here

src/SourceLocator/AutoloadSourceLocator.php
```diff
@@ -0,0 +1,147 @@
...
+    {
+        $assoc = [
+            'dev' => 2056,
```
Ocramius: WOOOOO, loads of magic constants. WTF is does it even mean? <_<

src/SourceLocator/AutoloadSourceLocator.php
```diff
@@ -0,0 +1,147 @@
...
+     * that it cannot find the file, so we squelch the errors by overriding the
+     * error handler temporarily.
+     *
+     * @param string $className#
+     */
+    private function locateClassByName($className)
+    {
+        if (class_exists($className, false)) {
+            $reflection = new \ReflectionClass($className);
+            self::$autoloadLocatedFile = $reflection->getFileName();
```
Ocramius: This could return `false`

src/SourceLocator/AutoloadSourceLocator.php
asgrim: It's not magic constant, it's just dummy data (see the docblock...)

src/SourceLocator/AutoloadSourceLocator.php
asgrim: that's okay, exception will be caught up ahead

src/SourceLocator/AutoloadSourceLocator.php
asgrim: yep

src/SourceLocator/AutoloadSourceLocator.php
asgrim: done

src/SourceLocator/AutoloadSourceLocator.php
asgrim: done

src/SourceLocator/AutoloadSourceLocator.php
asgrim: done

src/SourceLocator/AutoloadSourceLocator.php
asgrim: done

src/SourceLocator/AutoloadSourceLocator.php
asgrim: done

PR #45, opened 2015-07-07
Context for FindParameterType

asgrim:
By using the better availability of `LocatedSource` from PR #44, we are able to now determine parameter types with context, thus resolving #29 

* [x] Depends on #44 being merged - require rebase
* Required for #46 to be merged

### DIFF
--- a/src/TypesFinder/FindParameterType.php
+++ b/src/TypesFinder/FindParameterType.php
@@ -2,7 +2,9 @@
 
 namespace BetterReflection\TypesFinder;
 
+use BetterReflection\Reflection\ReflectionMethod;
 use phpDocumentor\Reflection\Types\Context;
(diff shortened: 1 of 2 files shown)


src/TypesFinder/FindParameterType.php
```diff
@@ -27,10 +30,25 @@ public function __invoke(ReflectionFunctionAbstract $function, ParamNode $node)
...
         return [];
     }
+
+    /**
+     * @param ReflectionFunctionAbstract $function
+     * @return Context
+     */
+    private function createContextForFunction(ReflectionFunctionAbstract $function)
+    {
+        if ($function instanceof ReflectionMethod) {
```
Ocramius: You are not covering all paths in this method, are you?

asgrim: Am now! :dancer:

PR #46, opened 2015-07-07
Context for FindTypeFromAst

asgrim:
By using the better availability of `LocatedSource` from PR #44, we are able to now determine parameter types with context, thus resolving #30 

* [x] Depends on #45 being merged - require rebase

### DIFF
--- a/src/Reflection/ReflectionParameter.php
+++ b/src/Reflection/ReflectionParameter.php
@@ -126,7 +126,15 @@ class ReflectionParameter implements \Reflector
         $param->isVariadic = (bool)$node->variadic;
         $param->isByReference = (bool)$node->byRef;
         $param->parameterIndex = (int)$parameterIndex;
-        $param->typeHint = (new FindTypeFromAst())->__invoke($node->type);
+        $param->typeHint = (new FindTypeFromAst())->__invoke(
+            $node->type,
+            $function->getLocatedSource(),
+            (
+                $function instanceof ReflectionMethod
+                ? $function->getDeclaringClass()->getNamespaceName()
+                : $function->getNamespaceName()
+            )
+        );
 
         if ($param->hasDefaultValue) {
             $param->parseDefaultValueNode($node->default);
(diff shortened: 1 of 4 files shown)


src/Reflection/ReflectionParameter.php
```diff
@@ -126,7 +126,15 @@ public static function createFromNode(
         $param->isVariadic = (bool)$node->variadic;
         $param->isByReference = (bool)$node->byRef;
         $param->parameterIndex = (int)$parameterIndex;
-        $param->typeHint = (new FindTypeFromAst())->__invoke($node->type);
+        $param->typeHint = (new FindTypeFromAst())->__invoke(
+            $node->type,
+            $function->getLocatedSource(),
+            (
+                $function instanceof ReflectionMethod
```
Ocramius: The check seems to be a bit too loose. Are you sure `getNamespaceName` is available if it's not a `ReflectionMethod`?

asgrim: yes - if it is not available, it will be an empty string `''`, which is expected by phpdocumentor type resolver

asgrim: This is actually how TypeResolver works internally as well: https://github.com/phpDocumentor/TypeResolver/blob/master/src/Types/ContextFactory.php#L43

Ocramius: What I mean is: "is `getNamespaceName` available as a method?"

asgrim: yes, `ReflectionFunctionAbstract` has `getNamespaceName()` (because a function can have a namespace). Oddly enough, PHP behaviour is that if it's a `ReflectionMethod` (that inherits `ReflectionFunctionAbstract`) then the namespace will all be empty, because that information *should* be retrieved from the declaring class - hence if it's a method, you call `$function->getDeclaringClass()->getNamespaceName()` and if it's just a function you directly call `$function->getNamespaceName()`

Ocramius: @asgrim that clears it up. Also, I hope we fix this horrible bug via `BetterReflection` :P

PR #56, opened 2015-07-10
Adds compatibility for ReflectionProperty

asgrim:
As part of #7, this PR completes method compatibility for the `ReflectionProperty` class, including test coverage.

Note that the API for the `FindPropertyType` invokable has changed, because we no longer need the AST node, due to implementing the `getDocComment()` method in `ReflectionProperty`.

### DIFF
--- a/src/Reflection/ReflectionProperty.php
+++ b/src/Reflection/ReflectionProperty.php
@@ -43,11 +43,6 @@ class ReflectionProperty implements \Reflector
      */
     private $docBlock = '';
 
-    /**
-     * @var bool
-     */
-    private $isRuntimeDeclared = false;
-
     private function __construct()
     {
     }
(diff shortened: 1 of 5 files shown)


src/Reflection/ReflectionProperty.php
```diff
@@ -38,6 +38,16 @@ class ReflectionProperty implements \Reflector
...
 
+    /**
+     * @var string
+     */
+    private $docBlock = '';
+
+    /**
+     * @var bool
+     */
+    private $isRuntimeDeclared = false;
```
Ocramius: Unused

asgrim: Yes, it's true, but it makes intent much clearer when reading the code in the `isDefault()` implementation.

Ocramius: @asgrim use a constant then

PR #59, opened 2015-07-12
Compatibility for ReflectionClass

### LINKED ISSUE #7: Need to become "mostly" compatible with PHP's Reflection API

asgrim:
This PR brings about "most" compatibility for `ReflectionClass`, thus merging this will finally resolve #7.

### General
* [x] Rebase onto master before merge (#60 may cause conflict)
* [x] getParentClass
* [x] getDefaultProperties
* [x] getDocComment
* [x] getStartLine
* [x] getEndLine
* [x] hasConstant
* [x] hasMethod
* [x] hasProperty
* [x] isInternal
* [x] isUserDefined
* [x] isAbstract
* [x] isFinal
* [x] getModifiers

### Trait-related
* [x] getTraits
* [x] getTraitNames
* [x] getTraitAliases - waiting for PR merge: phpDocumentor/TypeResolver#9
* [x] isTrait

### Interface-related
* [x] getInterfaceNames
* [x] getInterfaces
* [x] isInterface

### Needs parent/interface/trait hierarchy
* [x] isCloneable
* [x] isInstance
* [x] isInstantiable
* [x] isIterateable
* [x] isSubClassOf
* [x] implementsInterface

### Other stuff
* [x] Inject `BetterReflection\Reflector\Reflector` instead of a `SourceLocator` to things like `getTraits`, `getInterfaces`, `getParentClass` (and other related methods)

### DIFF
--- a/src/Reflection/ReflectionClass.php
+++ b/src/Reflection/ReflectionClass.php
@@ -205,6 +205,22 @@ class ReflectionClass implements Reflection
     }
 
     /**
+     * Does the class have the specified method method?
+     *
+     * @param $methodName
+     * @return bool
+     */
+    public function hasMethod($methodName)
+    {
+        try {
+            $this->getMethod($methodName);
+            return true;
+        } catch (\OutOfBoundsException $exception) {
+            return false;
+        }
+    }
+
+    /**
      * Get an array of the defined constants in this class.
      *
      * @return mixed[]
@@ -224,11 +240,22 @@ class ReflectionClass implements Reflection
      */
     public function getConstant($name)
     {
-        if (!isset($this->constants[$name])) {
+        if (!$this->hasConstant($name)) {
             return null;
         }
 
         return $this->constants[$name];
+    }
+
+    /**
+     * Does this class have the specified constant?
+     *
+     * @param $name
+     * @return bool
+     */
+    public function hasConstant($name)
+    {
+        return isset($this->constants[$name]);
     }
 
     /**
@@ -261,11 +288,22 @@ class ReflectionClass implements Reflection
      */
     public function getProperty($name)
     {
-        if (!isset($this->properties[$name])) {
+        if (!$this->hasProperty($name)) {
             return null;
         }
 
         return $this->properties[$name];
+    }
+
+    /**
+     * Does this class have the specified property?
+     *
+     * @param $name
+     * @return bool
+     */
+    public function hasProperty($name)
+    {
+        return isset($this->properties[$name]);
     }
 
     /**
--- a/src/Reflection/ReflectionClass.php
+++ b/src/Reflection/ReflectionClass.php
@@ -31,7 +31,7 @@ class ReflectionClass implements Reflection
     /**
      * @var NamespaceNode
      */
-    private $declaringNamespace = null;
+    private $declaringNamespace;
 
     /**
      * @var ReflectionMethod[]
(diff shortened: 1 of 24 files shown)


src/Reflection/ReflectionClass.php
```diff
@@ -182,6 +210,22 @@ public function getMethod($methodName)
     }
 
     /**
+     * Does the class have the specified method method?
+     *
+     * @param $methodName
```
Ocramius: string

src/Reflection/ReflectionClass.php
```diff
@@ -209,6 +253,17 @@ public function getConstant($name)
     }
 
     /**
+     * Does this class have the specified constant?
+     *
+     * @param $name
```
Ocramius: string

src/Reflection/ReflectionClass.php
```diff
@@ -246,6 +301,30 @@ public function getProperty($name)
     }
 
     /**
+     * Does this class have the specified property?
+     *
+     * @param $name
```
Ocramius: string

src/Reflection/ReflectionClass.php
```diff
@@ -260,4 +339,288 @@ public function getLocatedSource()
...
+    {
+        return (int)$this->node->getAttribute('startLine', -1);
+    }
+
+    /**
+     * Get the line number that this class ends on.
+     *
+     * @return int
+     */
+    public function getEndLine()
```
Ocramius: As suggested, getting the column may also be very interesting

asgrim: Added #68

Ocramius:
> * [ ] Inject `BetterReflection\Reflector\Reflector` instead of a `SourceLocator` to things like `getTraits`, `getInterfaces`, `getParentClass` (and other related methods)

Consider using just the internal reflector, or modifying the reflector at runtime (I don't see a use-case for it, though)

asgrim:
Note - `ReflectionClass` doesn't actually have access to an internal
`Reflector` as this never gets injected... so I think for this to work it
would have to be injected in the static constructor.

asgrim:
Note - `ReflectionClass` doesn't actually have access to an internal
`Reflector` as this never gets injected... so I think for this to work it
would have to be injected in the static constructor.

PR #63, opened 2015-07-15
Compile array expressions in CompileNodeToValue

### LINKED ISSUE #51: CompileNodeToValue: compile array correctly
At the moment if the node is an array, we just return an empty array `[]` as the value. We should traverse the nodes and determine the real values in the array.

asgrim:
Merging this should resolve #51

### DIFF
--- a/test/unit/NodeCompiler/CompileNodeToValueTest.php
+++ b/test/unit/NodeCompiler/CompileNodeToValueTest.php
@@ -23,6 +23,10 @@ class CompileNodeToValueTest extends \PHPUnit_Framework_TestCase
             ['[]', []],
             ['false', false],
             ['true', true],
+            ['[1,2,3]', [1, 2, 3]],
+            ['["foo","bar"]', ['foo', 'bar']],
+            ['[1 => "foo", 2 => "bar"]', [1 => 'foo', 2 => 'bar']],
+            ['["foo" => "bar"]', ['foo' => 'bar']],
         ];
     }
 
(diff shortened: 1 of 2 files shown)


test/unit/NodeCompiler/CompileNodeToValueTest.php
```diff
@@ -23,6 +23,10 @@ public function nodeProvider()
             ['[]', []],
             ['false', false],
             ['true', true],
+            ['[1,2,3]', [1, 2, 3]],
+            ['["foo","bar"]', ['foo', 'bar']],
+            ['[1 => "foo", 2 => "bar"]', [1 => 'foo', 2 => 'bar']],
+            ['["foo" => "bar"]', ['foo' => 'bar']],
```
Ocramius: What about stuff like `'["foo"."bar" => 2 * 3]'`?

asgrim: https://github.com/Roave/BetterReflection/issues/64

Ocramius: @asgrim add a `@todo` in the tests

asgrim: There's no point, there could be loads of tests we could add here. Issue will suffice for this, otherwise that test will be one massive block of `@todo` comments

Ocramius: Alrighty

Ocramius:
:+1:

Issue #64, opened 2015-07-15
CompileNodeToValue: compile complex expressions


asgrim:
e.g. stuff like `["foo"."bar" => 2 * 3]`

They are constant expressions, but currently not supported.

asgrim:
Also:

```php
<?php

class Foo
{
    const BAR = 123 * self::BAZ;
    const BAZ = 456 * self::BAR;
}

var_dump(Foo::BAR, Foo::BAZ);
```

asgrim:
@ircmaxell I'd love it if you had any feedback on what the best way to implement this may be.

To summarise, given that we have an AST node (that represents a value):

```php
public function foo($x = null) {
```

Currently we have this [current CompileNodeToValue.php](https://github.com/Roave/BetterReflection/blob/336d9aa95e703907dd8e2f17ea8d3947796ed1f6/src/NodeCompiler/CompileNodeToValue.php)

Which takes the AST node (representing `null` in the example above), and evaluating `null` is pretty simple (it is a `ConstNode` so we can just return the value `null`).

However, it starts to get a little more complicated if you have an expression, e.g.:

```php
5 + (self::FOO * 3)
```

The AST tree needs to essentially be "compiled" now (beyond the rudimentary switch we have currently). The current proposal we have is in [proposed CompileNodeToValue.php](https://github.com/Roave/BetterReflection/blob/a79162c1dd1079e6ac468d0a25e601c47e65aa68/src/NodeCompiler/CompileNodeToValue.php) (part of PR #67), but basically uses the `PhpParser` printer to render PHP code (for whitelisted nodes) and then `eval()`'s the statement...

The main issues at the moment with this approach is:

* `ClassConstFetch` and `ConstFetch` (that isn't `true|false|null`) are difficult to evaluate - contextual information is required to do this. We've discussed this in issue #19 and feel that the best solution at the moment is to only resolve `ClassConstFetch`, otherwise there's a big rabbit hole. Bear in mind we could potentially be very limited to the scope we are able to look for defined constants for example (e.g. if we only have the source code of a single file, but the constant defined in a different file to that which we are reflecting from, then we would not be able to evaluate the constant).
* This code in #67 uses `eval()`... despite using a white list, I still feel uncomfortable using `eval()` to do this, although I'm not sure if there are any other approaches...

If you have any moments spare, I'd love it if we could pick your brains on this one! Thanks so much :)

asgrim:
Closed in #95

Issue #19, opened 2015-07-02
CompileNodeToValue does not evaluate the value of constant expressions


asgrim:
If you have a parameter with a default value of a constant (e.g. `self::WHATEVER`), we currently do not evaluate the *value* of `self::WHATEVER`, we just return `self::WHATEVER` as a string. This is incorrect, we need to (wherever possible) determine the actual value.

asgrim:
Failing test case:

```php
public function testClassConstantUsedAsDefaultValue()
{
    $php = '<?php
        class MyClass {
            const FOO = 123;

            public function bar($baz = self::FOO);
        }
    ';

    $tree = (new Parser(new Lexer))->parse($php);

    $actualValue = (new CompileNodeToValue())
        ->__invoke($tree[0]->stmts[1]->params[0]->default);
    $this->assertSame(123, $actualValue);
}
```

Ocramius:
Few notes:

 - constants in a class can refer to other constants in a class (recursive evaluation)
 - constants may be undefined (php does trigger a notice at runtime in those cases: http://3v4l.org/rDiY1)

asgrim:
Also note that you could have something like:

```php
<?php
define("LOL", $argv[0]);

function foo($foo = LOL) {
 var_dump($foo);
}
```

which you can never determine statically, because $argv[0] is out of context. Note that `define` is a `Node\Expr\FuncCall`, and the second parameter is a node something like:

```
Node\Expr\ArrayDimFetch
  ->var = Node\Expr\Variable
    ->name = argv
  ->dim = Node\Scalar\LNumber
    ->value = 0
```

However, pretty much exactly the same but we **can** work this one out:

```php
<?php
define("LOL", 123);

function foo($foo = LOL) {
 var_dump($foo);
}
```

As you'd just find all the constants defined as "LOL" and see it is a scalar expression. It gets a little more complicated when you have:

```php
<?php
$foo = [1, 2, 3];
define("LOL", $foo[0]);

function foo($foo = LOL) {
 var_dump($foo);
}
```

With this you have almost no more information in the specific node, so you MUST look around in the context (the code...) and try to evaluate the expression.

```
Node\Expr\ArrayDimFetch
  ->var = Node\Expr\Variable
    ->name = foo
  ->dim = Node\Scalar\LNumber
    ->value = 0
```

[...]

PR #67, opened 2015-07-15
[WIP] Improved node compilation in CompileNodeToValue (eval solution)

asgrim:
This PR aims to solve both #64 and #19 

An alternative solution exists in #95

### DIFF
--- a/src/NodeCompiler/CompileNodeToValue.php
+++ b/src/NodeCompiler/CompileNodeToValue.php
@@ -3,6 +3,7 @@
 namespace BetterReflection\NodeCompiler;
 
 use PhpParser\Node;
+use PhpParser\PrettyPrinter\Standard as Printer;
 
 class CompileNodeToValue
 {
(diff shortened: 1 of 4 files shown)


asgrim:
@Ocramius thoughts? :P

src/NodeCompiler/CompileNodeToValue.php
```diff
@@ -15,51 +16,67 @@ class CompileNodeToValue
...
-                    return true;
-                default:
-                    // @todo this should evaluate the VALUE, not the name
-                    /* @see https://github.com/Roave/BetterReflection/issues/19 */
-                    return $firstName;
-            }
+        $printer = new Printer();
+        $code = $printer->prettyPrint([$node]);
+
+        eval('$x = ' . $code);
```
nikic: A constant access could be nested in an expression (`foo::bar + 1`), in which case eval won't work (unless you actually want a runtime constant lookup).

Ocramius: Traversal to replace such nodes should indeed happen upfront :-(

asgrim: Aye, ConstFetch and ClassConstFetch have dedicated handlers above (although I haven't dealt with them properly yet still). It's mainly my concern using `eval` tbh...

Ocramius: @asgrim you can probably use `eval()` if you first validate that ALL the nodes are within a given whitelist (WHITELIST! No blacklists, plox :P )

asgrim: thanks for feedbacks folks :)

PR #73, opened 2015-07-19
`ReflectionClass#isSubclassOf()` support

Ocramius:
Required for #59

### DIFF
--- a/src/Reflection/ReflectionClass.php
+++ b/src/Reflection/ReflectionClass.php
@@ -5,6 +5,7 @@ namespace BetterReflection\Reflection;
 use BetterReflection\NodeCompiler\CompileNodeToValue;
 use BetterReflection\Reflection\Exception\NoParent;
 use BetterReflection\Reflection\Exception\NotAnObject;
+use BetterReflection\Reflection\Exception\NotAString;
 use BetterReflection\Reflector\ClassReflector;
 use BetterReflection\SourceLocator\AutoloadSourceLocator;
 use BetterReflection\SourceLocator\LocatedSource;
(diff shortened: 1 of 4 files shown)


src/Reflection/ReflectionClass.php
```diff
@@ -627,13 +628,13 @@ public function getTraitAliases()
...
-     * @param SourceLocator|null $sourceLocator a source locator - if none is provided, an autoloader-based locator
-     *                                          will be used
+     * @param SourceLocator $sourceLocator a source locator - if none is provided, an autoloader-based locator
+     *                                     will be used
      *
      * @return ReflectionClass[] An associative array of interfaces, with keys as interface names and the array
      *                           values as {@see ReflectionClass} objects.
      */
-    public function getInterfaces(SourceLocator $sourceLocator = null)
+    public function getInterfaces(SourceLocator $sourceLocator)
```
Ocramius: This was changed because the previous logic didn't account for a null source locator. This should be handled again when we have an idea of what we want when we deal with the injected reflector.

src/Reflection/ReflectionClass.php
```diff
@@ -688,6 +689,34 @@ public function isInstance($object)
     }
 
     /**
+     * Checks whether the given object is an instance
```
asgrim: Is this correct docs?

asgrim: I think it would be worth clarifying exactly the difference between this and `isInstanceOf`

Ocramius: Indeed. I'll change this.

PR #75, opened 2015-07-20
`ReflectionClass#implementsInterface()` support

Ocramius:
Required for #59


asgrim:
@Ocramius Hmm, I merged, but should this not traverse up the parent classes to see if the class implements an interface (but further up the inheritance tree)?

asgrim:
@Ocramius sorry, ignore me, I see it uses `getInterfaceNames()` which already goes up the tree ;)

PR #78, opened 2015-07-20
AggregateSourceLocator for combining multiple SourceLocators

### LINKED ISSUE #53: TraversableSourceLocator or something
The end goal of this is to come up with something that can locate from multiple files / sources. @Ocramius suggested using something traversable, so you pass in for example a `DirectoryIterator` which the locate would use to find the class. I like the flexibility of this approach, but the only question mark for me at the moment is how to determine the filename, for example if an `ArrayIterator` was given (which was an array of source code items etc.)

asgrim:
Add AggregateSourceLocator which is now used internally if the source locator specified is not already an aggregate.

This fixes #53 which means in the future we can add multiple sources to source locators.

### DIFF
--- a/src/Reflector/Generic.php
+++ b/src/Reflector/Generic.php
@@ -40,7 +40,7 @@ class Generic
             $aggregate = new AggregateSourceLocator([$this->sourceLocator]);
         }
 
-        foreach ($aggregate->__invoke($identifier) as $locatedSource) {
+        foreach ($aggregate($identifier) as $locatedSource) {
             try {
                 return $this->reflectFromLocatedSource($identifier, $locatedSource);
             }
--- a/src/Reflector/Generic.php
+++ b/src/Reflector/Generic.php
@@ -5,6 +5,7 @@ namespace BetterReflection\Reflector;
 use BetterReflection\Identifier\Identifier;
 use BetterReflection\Identifier\IdentifierType;
 use BetterReflection\Reflection\ReflectionFunction;
+use BetterReflection\SourceLocator\AggregateSourceLocator;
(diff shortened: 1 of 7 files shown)


src/Reflector/Generic.php
```diff
@@ -34,10 +35,20 @@ public function __construct(SourceLocator $sourceLocator)
...
-        return $this->reflectFromLocatedSource(
-            $identifier,
-            $this->sourceLocator->__invoke($identifier)
-        );
+        $aggregate = $this->sourceLocator;
+        if (!($aggregate instanceof AggregateSourceLocator)) {
+            $aggregate = new AggregateSourceLocator([$this->sourceLocator]);
+        }
+
+        foreach ($aggregate->__invoke($identifier) as $locatedSource) {
```
Ocramius: `$aggregate($identifier)`. If you use `__invoke` explicitly, you lose the advantage of type-hinting `callable`

src/Reflector/Generic.php
```diff
@@ -34,10 +35,20 @@ public function __construct(SourceLocator $sourceLocator)
...
+            try {
+                return $this->reflectFromLocatedSource($identifier, $locatedSource);
+            }
+            catch (Exception\IdentifierNotFound $identifierNotFound) {
+            }
+        }
+
+        throw Exception\IdentifierNotFound::fromIdentifier($identifier);
```
Ocramius: Should probably encapsulate the previous exceptions, in order to show tested paths. Overkill?

asgrim: Possibly in future, if someone requires it, otherwise YAGNI

PR #79, opened 2015-07-20
Source locator parameter removal when traversing reflection objects

Ocramius:


src/Reflection/ReflectionClass.php
```diff
@@ -370,22 +380,17 @@ public function getEndLine()
...
         }
 
-        $fqsen = $this->extendsClassType->__toString();
-
-        if (null !== $sourceLocator) {
-            $parent = (new ClassReflector($sourceLocator))->reflect($fqsen);
-        } else {
-            $parent = self::createFromName($fqsen);
-        }
+        // @TODO use actual `ClassReflector` or `FunctionReflector`?
```
asgrim: is this `@todo` still needed?

src/Reflection/ReflectionClass.php
```diff
@@ -540,18 +544,12 @@ private function getFqsenFromNamedNode(Node\Name $node)
...
+    private function reflectClassForNamedNode(Node\Name $node)
     {
-        $fqsen = $this->getFqsenFromNamedNode($node);
-
-        if (null !== $sourceLocator) {
-            return (new ClassReflector($sourceLocator))->reflect($fqsen);
-        }
-
-        return self::createFromName($fqsen);
+        // @TODO use actual `ClassReflector` or `FunctionReflector`?
```
asgrim: same

src/Reflection/ReflectionClass.php
```diff
@@ -370,22 +380,17 @@ public function getEndLine()
...
-        $fqsen = $this->extendsClassType->__toString();
-
-        if (null !== $sourceLocator) {
-            $parent = (new ClassReflector($sourceLocator))->reflect($fqsen);
-        } else {
-            $parent = self::createFromName($fqsen);
-        }
+        // @TODO use actual `ClassReflector` or `FunctionReflector`?
+        /* @var $parent self */
+        $parent = $this->reflector->reflect((string) $this->extendsClassType);
```
asgrim: Can you not just call `$this->reflectClassForNamedNode((string)$this->extendsClassType);`?

src/Reflection/ReflectionClass.php
Ocramius: I am still not sure, that's why it's still there. I have no specific `ClassReflector` hinting, therefore the usage is unsafe in here

src/Reflection/ReflectionClass.php
Ocramius: same as above. `@TODO` stands imo

src/Reflection/ReflectionClass.php
asgrim: Some kind of carnage...

PR #80, opened 2015-07-20
Internal class reflections

### LINKED ISSUE #38: Create ReflectionFunctions/ReflectionClasses from internal definitions
We currently can't create `ReflectionFunction`s and `ReflectionClass`es from internal functions/class, as the source code is not available. The only way I see is to use the core reflection API, which is a bit of a fudge, and would rely on #36 anyway (would it even be possible to get type hints? maybe not)

_Todo_
- [x] Internal classes (see #80)
- [x] Internal functions

asgrim:
This PR fixes #38

* `AggregateSourceLocator`s can be nested now (`Reflector\Generic` will recurse through them in a flat ordered way)
* I had to modify the behaviour of the `SourceLocator`s slightly so that they don't throw exceptions, but merely return null. This is so that when we loop through the `AggregateSourceLocator`, we don't break the loop.
* By default `PhpInternalSourceLocator` gets added to all calls to `reflect()`. Can be disabled by a paramter
* A new magic const is defined as a flag to notify `ReflectionClass` that the `LocatedSource` originated from the `PhpInternalSourceLocator` (thus `isInternal() === true`)

### Todo
* [x] Needs the rest of the internal classes implemented...
* [x] ~~~Need to implement functions too...~~~
* [x] More path coverage for `PhpInternalSourceLocator`

### DIFF
--- a/src/SourceLocator/AggregateSourceLocator.php
+++ b/src/SourceLocator/AggregateSourceLocator.php
@@ -30,7 +30,13 @@ class AggregateSourceLocator implements SourceLocator
     public function __invoke(Identifier $identifier)
     {
         foreach ($this->sourceLocators as $sourceLocator) {
-            yield $sourceLocator->__invoke($identifier);
+            if ($sourceLocator instanceof self) {
+                foreach ($sourceLocator->__invoke($identifier) as $value) {
+                    yield $value;
+                }
+                continue;
+            }
+            yield $sourceLocator($identifier);
         }
     }
 }
--- a/src/SourceLocator/LocatedSource.php
+++ b/src/SourceLocator/LocatedSource.php
@@ -7,8 +7,6 @@ namespace BetterReflection\SourceLocator;
  */
 class LocatedSource
 {
-    const INTERNAL_SOURCE_MAGIC_CONST = 'BetterReflectionPhpInternalSourceStub';
-
     /**
      * @var string
      */
@@ -19,6 +17,10 @@ class LocatedSource
      */
     private $filename;
 
+    /**
+     * @param string      $source
+     * @param string|null $filename
+     */
     public function __construct($source, $filename)
     {
         if (!is_string($source) || empty($source)) {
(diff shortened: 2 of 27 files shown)


src/SourceLocator/AggregateSourceLocator.php
```diff
@@ -30,7 +30,13 @@ public function __construct(array $sourceLocators = [])
     public function __invoke(Identifier $identifier)
     {
         foreach ($this->sourceLocators as $sourceLocator) {
-            yield $sourceLocator->__invoke($identifier);
+            if ($sourceLocator instanceof self) {
```
Ocramius: Why not just `if ($sourceLocator instanceof Traversable) {`?

src/SourceLocator/LocatedSource.php
```diff
@@ -31,7 +33,7 @@ public function __construct($source, $filename)
             );
         }
 
-        if (null !== $filename) {
+        if (null !== $filename && self::INTERNAL_SOURCE_MAGIC_CONST !== $filename) {
```
Ocramius: This is REALLY fishy and hard to comprehend

src/SourceLocator/PhpInternalSourceLocator.php
```diff
@@ -0,0 +1,30 @@
...
+            return null;
+        }
+
+        $methodName = str_replace('\\', '__', $identifier->getName());
+        if (!method_exists($this, $methodName)) {
+            return null;
+        }
+
+        return new LocatedSource(
+            '<?php ' . $this->$methodName(),
```
Ocramius: As suggested, stubbing would probably be better here...

src/SourceLocator/AggregateSourceLocator.php
asgrim: 'cuz `AggregateSourceLocator` is not `Traversable`?

src/SourceLocator/LocatedSource.php
asgrim: yeah, I wasn't sure how to do it, but it's basically to pass the flag down to `ReflectionClass->isInternal()` method so it knows internal or not

src/SourceLocator/PhpInternalSourceLocator.php
asgrim: $methodName returns the stub

Issue #34, opened 2015-07-03
Implement lazy/runtime reflection


asgrim:
Currently, we read the AST and populate the `Reflection*` objects at reflection-creation-time. This could potentially cause issues at some point, where somewhere down the line, an incomplete `Reflection*` could be used (case in point, see #27 - had to move the `declaringClass` to before the parent setup so that `getDeclaringClass()` was available when examining the method).

We could potentially simply store the AST at creation-time, and then when methods are called at run-time, perform required reflections as and when.

* [x] ReflectionProperty - #82 
* [x] ReflectionParameter - #83 
* [x] ReflectionMethod - #88 
* [x] ReflectionFunctionAbstract - #84 
* [x] ReflectionFunction (does not have anything to make lazy currently) - #84 
* [x] ReflectionClass #89

GeeH:
YES!

We should simply grab the AST when the class is reflected and perform some basic reflection, and lazily do the rest when the getters are called **as long as the AST is immutable** once the class is created.

asgrim:
Yeah, I'm thinking this is going to be much better overall, I'm going to put this into the 1.0.0 milestone, just so we don't have to worry about any BC breaks as a result of implementation.

PR #82, opened 2015-07-21
Made ReflectionProperty load lazily

asgrim:
Part of #34 - this PR makes `ReflectionProperty` load "lazily"


Ocramius:
As per review, requires testing internal state (greybox testing) in order to verify that lazy loading is properly happening.

PR #90, opened 2015-07-26
Increase coverage again

asgrim:
Improved coverage. There's a couple of uncovered things still:

* CompileNodeToValue - because implementation incomplete (bugs already open)
* `ClassReflector->reflect` - not sure why, the exit path seems uncovered for me, very odd o_O
* Similar thing with `Generic->reflectFromNamespace`, the parameter `LocatedSource $locatedSource` on L141 is uncovered for me

cov


Ocramius:
:+1:

PR #92, opened 2015-07-31
ReflectionObject

### LINKED ISSUE #16: Implement ReflectionObject
It's not supported currently because it requires an instance of an existing class, which means we can't currently implement...

asgrim:
Fixes #16 by implementing `ReflectionObject` which can reflect run-time properties.

### DIFF
--- a/src/Reflection/ReflectionClass.php
+++ b/src/Reflection/ReflectionClass.php
@@ -414,6 +414,18 @@ class ReflectionClass implements Reflection, \Reflector
     }
 
     /**
+     * Protected API to add properties at runtime (used in ReflectionObject)
+     *
+     * @param ReflectionProperty $property
+     */
+    protected function addProperty(ReflectionProperty $property)
+    {
+        // Ensure the properties are already cached
+        $this->getProperties();
+        $this->cachedProperties[$property->getName()] = $property;
+    }
+
+    /**
      * Return an array with default properties (properties that were defined at
      * compile-time rather than at run time).
      *
--- a/src/Reflection/ReflectionClass.php
+++ b/src/Reflection/ReflectionClass.php
@@ -67,6 +67,12 @@ class ReflectionClass implements Reflection, \Reflector
     {
     }
 
+    /**
+     * Create a reflection and return the string representation of a named class
+     *
+     * @param string $className
+     * @return string
+     */
     public static function export($className = null)
     {
         if (null === $className) {
@@ -77,6 +83,14 @@ class ReflectionClass implements Reflection, \Reflector
         return $reflection->__toString();
     }
 
+    /**
+     * Get a string representation of this reflection
+     *
+     * @todo Refactor this
+     * @see https://github.com/Roave/BetterReflection/issues/94
+     *
+     * @return string
+     */
     public function __toString()
     {
         $isObject = $this instanceof ReflectionObject;
--- a/src/Reflection/ReflectionClass.php
+++ b/src/Reflection/ReflectionClass.php
@@ -79,12 +79,15 @@ class ReflectionClass implements Reflection, \Reflector
 
     public function __toString()
     {
-        $format  = "Class [ <user> class %s%s%s ] {\n";
+        $isObject = $this instanceof ReflectionObject;
+
+        $format  = "%s [ <user> class %s%s%s ] {\n";
         $format .= "  @@ %s %d-%d\n\n";
         $format .= "  - Constants [%d] {%s\n  }\n\n";
         $format .= "  - Static properties [%d] {%s\n  }\n\n";
         $format .= "  - Static methods [%d] {%s\n  }\n\n";
         $format .= "  - Properties [%d] {%s\n  }\n\n";
+        $format .= ($isObject ? "  - Dynamic properties [%d] {%s\n  }\n\n" : '%s%s');
         $format .= "  - Methods [%d] {%s\n  }\n";
         $format .= "}\n";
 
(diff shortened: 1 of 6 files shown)


src/Reflection/ReflectionClass.php
```diff
@@ -331,6 +331,18 @@ public function hasProperty($name)
     }
 
     /**
+     * Protected API to add properties at runtime (used in ReflectionObject)
+     *
+     * @param ReflectionProperty $property
+     */
+    protected function addProperty(ReflectionProperty $property)
```
Ocramius: Wait... you can add properties at runtime? WADDAFAK? 

If this is allowed, then it should happen in `ReflectionObject` internals, by overriding parent class properties rather than overwriting the `ReflectionClass` internal state.

asgrim: Yes:

```php
$foo = new \stdClass();
$foo->bar = 'baz';
```
If you create a `ReflectionObject` using `$foo` then `bar` is a runtime property.

`ReflectionObject` can't access `ReflectionClass` properties as they're private. However, can modify them to simply append them as we overwrite all the methods anyway.

Ocramius: @asgrim what about having "cachedProperties" (parent class state) and "runtimeProperties" (child class state)?

asgrim:
@Ocramius updated & rebased

src/Reflection/ReflectionClass.php
```diff
@@ -79,12 +79,15 @@ public static function export($className = null)
...
-        $format  = "Class [ <user> class %s%s%s ] {\n";
+        $isObject = $this instanceof ReflectionObject;
+
+        $format  = "%s [ <user> class %s%s%s ] {\n";
         $format .= "  @@ %s %d-%d\n\n";
         $format .= "  - Constants [%d] {%s\n  }\n\n";
         $format .= "  - Static properties [%d] {%s\n  }\n\n";
         $format .= "  - Static methods [%d] {%s\n  }\n\n";
         $format .= "  - Properties [%d] {%s\n  }\n\n";
+        $format .= ($isObject ? "  - Dynamic properties [%d] {%s\n  }\n\n" : '%s%s');
```
Ocramius: The parallelism between line 90 and line [153~154](https://github.com/Roave/BetterReflection/pull/92/files#diff-0b62f5919f325c15e41b6b4c09690ff7R151) makes this quite hard to follow. Maybe just hardcode the values in this particular line?

src/Reflection/ReflectionObject.php
```diff
@@ -0,0 +1,465 @@
+<?php
+
+namespace BetterReflection\Reflection;
+
+use BetterReflection\Reflector\ClassReflector;
+use BetterReflection\SourceLocator\EvaledCodeSourceLocator;
+use PhpParser\Node\Stmt\Property as PropertyNode;
+use PhpParser\Builder\Property as PropertyNodeBuilder;
+
+class ReflectionObject extends ReflectionClass
```
Ocramius: In order to avoid future breakages (since you designed this class as a decorator, but then used inheritance due to how Reflection works in internals) you need a test that verifies that ALL public methods of this class are declared in this class.

test/unit/Fixture/ClassForHinting.php
```diff
@@ -4,4 +4,5 @@
 
 class ClassForHinting
 {
+    public $someProperty;
```
Ocramius: This class is likely used somewhere else. Consider using a separate test asset for this particular test scenario

src/Reflection/ReflectionClass.php
Ocramius: Actually, logic should be replaced in the child class with some duplication. Horrible but correct solution :( 

I'm fine with having a `@TODO` instead of that

src/Reflection/ReflectionProperty.php
```diff
@@ -36,7 +41,7 @@ public function __toString()
     {
         return sprintf(
             'Property [%s %s%s $%s ]',
-            $this->isStatic() ? '' : ($this->isDefault() ? ' <default>' : ' <implicit>'),
+            $this->isStatic() ? '' : ($this->isDefault() ? ' <default>' : ' <dynamic>'),
```
Ocramius: Where did `<implicit>` go?

src/Reflection/ReflectionObject.php
asgrim: This is already done in [ReflectionObjectTest.php#126](https://github.com/Roave/BetterReflection/pull/92/files#diff-d4ef1e7f3bf562f17cbf82b1a40fa729R126)

src/Reflection/ReflectionProperty.php
asgrim: Apparently never used o_O not sure where I got that from initially

test/unit/Fixture/ClassForHinting.php
asgrim: Doesn't affect any other tests... why create another file for no reason?

src/Reflection/ReflectionClass.php
asgrim: Added `@todo`

src/Reflection/ReflectionObject.php
Ocramius: Oh, I missed that D:

I would probably just have done:

```php
$methods = (new ReflectionClass(ReflectionObject::class))->getMethods();
array_walk(
    $methods,
    function (ReflectionMethod $method) {
        if (! $method->isPublic()) {
            return;
        }
        
        $this->assertSame(ReflectionObject::class, $method->getDeclaringClass()->getName());
    }
);
```

asgrim: Added some good old commenting to describe the intent of the existing tests that cover this off

PR #93, opened 2015-08-02
Added some example stubs for internal classes

asgrim:
For #87
Just a proof of concept really.

### DIFF
--- a/src/SourceLocator/PhpInternalSourceLocator.php
+++ b/src/SourceLocator/PhpInternalSourceLocator.php
@@ -69,12 +69,12 @@ class PhpInternalSourceLocator implements SourceLocator
      * Determine the stub name
      *
      * @param string $className
-     * @return string
+     * @return string|null
      */
     private function buildStubName($className)
     {
         if (!preg_match('/^[a-zA-Z]+$/', $className)) {
-            throw new \InvalidArgumentException('Not a valid class name.');
+            return null;
         }
 
         return __DIR__ . '/../../stub/' . $className . '.stub';
@@ -88,9 +88,9 @@ class PhpInternalSourceLocator implements SourceLocator
      */
     public function hasStub($className)
     {
-        try {
-            $expectedStubName = $this->buildStubName($className);
-        } catch (\InvalidArgumentException $e) {
+        $expectedStubName = $this->buildStubName($className);
+
+        if (null === $expectedStubName) {
             return false;
         }
 
--- a/src/SourceLocator/PhpInternalSourceLocator.php
+++ b/src/SourceLocator/PhpInternalSourceLocator.php
@@ -17,6 +17,10 @@ class PhpInternalSourceLocator implements SourceLocator
             return null;
         }
 
+        if ($stub = $this->getStub($name)) {
+            return new InternalLocatedSource($stub);
+        }
+
         return new InternalLocatedSource(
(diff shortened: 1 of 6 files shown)


stub/Exception.stub.php
```diff
@@ -0,0 +1,17 @@
+<?php
```
Ocramius: May want to remove the `.php` suffix to avoid some fool from importing these

src/SourceLocator/PhpInternalSourceLocator.php
```diff
@@ -43,4 +47,15 @@ private function getInternalReflectionClassName(Identifier $identifier)
 
         return $reflection->isInternal() ? $reflection->getName() : null;
     }
+
+    private function getStub($name)
+    {
+        $expectedStubName = __DIR__ . '/../../stub/' . $name . '.stub.php';
```
Ocramius: This introduces an LFI security issue. Need to check the `$name` first.

Ocramius: In general, simply matching against `[A-Za-z]+` is enough right now, as we are dealing mostly with non-namespaced stubs only.

Ocramius:
This diff needs a change in the tests that ensures that stubbed classes are not skipped in test data providers

asgrim:
The tests simply test everything in `get_declared_classes|get_declared_interfaces|get_declared_traits` anyway, so they shouldn't be skipped? Maybe I misunderstood?

Ocramius:
@asgrim there is a functional test for this.

[this test](https://github.com/Roave/BetterReflection/blob/3b28cc655b4df8e4524030f30bceded68051d038/test/unit/SourceLocator/PhpInternalSourceLocatorTest.php#L58) and similar ones must be adapted to NOT skip in case of a stubbed class (even in case of exceptions).

asgrim:
@Ocramius it already does not skip the test if the class is stubbed - that's the whole point... simply by stubbing these classes, more tests are now passing instead of being marked incomplete...

Ocramius:
@asgrim the test should simply fail if an exception is thrown *AND* there is a stub. The test is currently not designed for that.

asgrim:
@Ocramius I see. There's no current external way to tell if a stub was found for a class, do we want to expose that functionality just to do a test?

src/SourceLocator/PhpInternalSourceLocator.php
```diff
@@ -43,4 +47,24 @@ private function getInternalReflectionClassName(Identifier $identifier)
...
+     * Get the stub source code for an internal class.
+     *
+     * Returns null if nothing is found.
+     *
+     * @param string $name Should only contain [A-Za-z]
+     * @return string|null
+     */
+    private function getStub($name)
+    {
+        $name = preg_replace('[^A-Za-z]', '', $name);
```
Ocramius: This should simply be 

```php
if (! preg_match('^[a-zA-Z]+$', $name)) {
    return null;
}
```

Ocramius:
> @Ocramius I see. There's no current external way to tell if a stub was found for a class, do we want to expose that functionality just to do a test? 

We'd probably just add a whitelist or reproduce the check against file existence in the try-catch.

asgrim:
@Ocramius updated like

Ocramius:
@asgrim :+1: 

Waiting for travis

PR #95, opened 2015-08-03
Improved node compilation in CompileNodeToValue (compile solution)

asgrim:
This PR aims to solve both #64 and #19.

An alternate PR exists in #67

~~Depends on PR #96 being merged~~

### DIFF
--- a/src/NodeCompiler/CompileNodeToValue.php
+++ b/src/NodeCompiler/CompileNodeToValue.php
@@ -3,7 +3,6 @@
 namespace BetterReflection\NodeCompiler;
 
 use BetterReflection\Reflection\ReflectionClass;
-use BetterReflection\Reflector\Reflector;
 use PhpParser\Node;
 
 class CompileNodeToValue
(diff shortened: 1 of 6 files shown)


src/NodeCompiler/CompileNodeToValue.php
```diff
@@ -27,44 +27,91 @@ public function __invoke(Node $node)
...
+        $firstName = reset($constNode->name->parts);
+        switch ($firstName) {
+            case 'null':
+                return null;
+            case 'false':
+                return false;
+            case 'true':
+                return true;
+            default:
+                throw new Exception\UnableToCompileNode('Unable to compile constant expressions');
```
Ocramius: May want to at least try casting the constant expression to string in here, or else there may be rage

Ocramius: `UnableToCompileNode::fromConstFetch($constNode)` maybe?

asgrim: Previously it was just returning a string with the const name, (e.g. `FOO` would compile to `"FOO"`), but surely this could cause undesired behaviour?

Ocramius: This could indeed cause undesired behavior :-(

asgrim:
@Ocramius I'm proposing this solution (over the `eval` solution in #67) to be accepted as the preferred for #19 and #64

asgrim:
#98 created to improve `compileBinaryOp` method

Issue #13, opened 2015-07-02
Support ReflectionParameter->getClass()


asgrim:
This issue is a result of the discussion here: https://github.com/Roave/BetterReflection/pull/9/files#r33761101

There are potential thorny issues surrounding recursion / not being able to access the `Reflector` instance etc. here, that prevent us from implementing `getClass` in a clean way.

Ocramius:
Implemented in #96

PR #96, opened 2015-08-03
Added ReflectionParameter->getClass() method

asgrim:
Fixes issue #13 by implementing `ReflectionParameter::getClass()` method

### DIFF
--- a/src/Reflection/ReflectionParameter.php
+++ b/src/Reflection/ReflectionParameter.php
@@ -366,4 +366,19 @@ class ReflectionParameter implements \Reflector
 
         return $this->defaultValueConstantName;
     }
+
+    /**
+     * Gets a ReflectionClass for the type hint (returns null if not a class)
+     *
+     * @return ReflectionClass|null
+     */
+    public function getClass()
+    {
+        $hint = $this->getTypeHint();
+        if (!($hint instanceof Types\Object_)) {
+            return null;
+        }
+
+        return ReflectionClass::createFromName($hint->getFqsen()->__toString());
+    }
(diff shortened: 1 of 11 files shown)


src/Reflection/ReflectionParameter.php
```diff
@@ -366,4 +366,19 @@ public function getDefaultValueConstantName()
...
+     * @return ReflectionClass|null
+     */
+    public function getClass()
+    {
+        $hint = $this->getTypeHint();
+        if (!($hint instanceof Types\Object_)) {
+            return null;
+        }
+
+        return ReflectionClass::createFromName($hint->getFqsen()->__toString());
```
Ocramius: Note that we should consistently pass around the same locators

asgrim: Rebased & updated

src/Reflection/ReflectionParameter.php
```diff
@@ -366,4 +375,23 @@ public function getDefaultValueConstantName()
...
+     * @return ReflectionClass|null
+     */
+    public function getClass()
+    {
+        $hint = $this->getTypeHint();
+        if (!($hint instanceof Types\Object_)) {
+            return null;
+        }
+
+        if (!$this->reflector instanceof ClassReflector) {
```
asgrim: This becomes an issue now because if this is a parameter that belongs to a function, the `$this->reflector` will be a `FunctionReflector` and thus we'll never be able to reflect on classes that are typehinted in a `ReflectionFunction`... :/

PR #99, opened 2015-08-13
Implement adapters for core reflection classes

### LINKED ISSUE #18: Extend core \ReflectionClass etc.
Adapter for core Reflection classes, e.g.

```php
<?php

namespace BetterReflection\CompatibleReflection;

use \ReflectionClass as CoreReflectionClass;
use BetterReflection\Reflection\ReflectionClass as BetterReflectionClass;

class ReflectionClass extends CoreReflectionClass
{
    /**
     * @var BetterReflectionClass
     */
    private $betterReflection;

    public function __construct()
    {
        throw new \Exception("Cannot instantiate like this...");
    }

    public static function createFromNode(
        ClassNode $node,
        NamespaceNode $namespace = null,
        $filename = null
    ) {
        $reflection = new self();
        $reflection->betterReflection = BetterReflectionClass::createFromNode($node, $namespace, $filename);
    }

    // Implement all the core methods
}
```

Alternatively, just make `\BetterReflection\Reflection\ReflectionClass` extend `\ReflectionClass` and ensure every method is implemented...

Why? So that something that is (incorrectly) type hinting for `\ReflectionClass` instead of `\Reflector` can use this library...

asgrim:
Would resolve #18 when complete.

### DIFF
--- a/src/Reflection/Adapter/ReflectionClass.php
+++ b/src/Reflection/Adapter/ReflectionClass.php
@@ -300,7 +300,7 @@ class ReflectionClass extends CoreReflectionClass
     /**
      * {@inheritDoc}
      */
-    public function newInstance(...$args)
+    public function newInstance($args = null, $_ = null)
     {
         throw new Exception\NotImplemented('Not implemented');
     }
--- a/src/Reflection/Adapter/ReflectionClass.php
+++ b/src/Reflection/Adapter/ReflectionClass.php
@@ -1,0 +1,24 @@
+<?php
+
+namespace BetterReflection\Reflection\Adapter;
+
+use ReflectionClass as CoreReflectionClass;
+use BetterReflection\Reflection\ReflectionClass as BetterReflectionClass;
+
+class ReflectionClass extends CoreReflectionClass
+{
+    /**
+     * @var BetterReflectionClass
+     */
+    private $betterReflectionClass;
+
+    public function __construct(BetterReflectionClass $betterReflectionClass)
+    {
+        $this->betterReflectionClass = $betterReflectionClass;
+    }
+
+    public function __call($method, $args)
+    {
+        $this->betterReflectionClass->$method(...$args);
+    }
+}
--- a/src/Reflection/Adapter/ReflectionClass.php
+++ b/src/Reflection/Adapter/ReflectionClass.php
@@ -372,13 +372,7 @@ class ReflectionClass extends CoreReflectionClass
      */
     public function getDefaultProperties()
     {
-        $properties = $this->betterReflectionClass->getDefaultProperties();
-
-        $wrappedProperties = [];
-        foreach ($properties as $key => $property) {
-            $wrappedProperties[$key] = new ReflectionProperty($property);
-        }
-        return $wrappedProperties;
+        throw new Exception\NotImplemented('Not implemented');
     }
 
     /**
--- a/src/Reflection/Adapter/ReflectionClass.php
+++ b/src/Reflection/Adapter/ReflectionClass.php
@@ -300,7 +300,7 @@ class ReflectionClass extends CoreReflectionClass
     /**
      * {@inheritDoc}
      */
-    public function newInstance($args = null, ...$args)
+    public function newInstance(...$args)
     {
         throw new Exception\NotImplemented('Not implemented');
     }
--- a/src/Reflection/Adapter/ReflectionClass.php
+++ b/src/Reflection/Adapter/ReflectionClass.php
@@ -307,7 +307,7 @@ class ReflectionClass extends CoreReflectionClass
     /**
      * @throws \Exception
      */
-    public function newInstance($args = null, $_ = null)
+    public function newInstance($args = null, ...$args)
     {
         throw new Exception\NotImplemented('Not implemented');
     }
--- a/src/Reflection/Adapter/ReflectionClass.php
+++ b/src/Reflection/Adapter/ReflectionClass.php
@@ -98,7 +98,7 @@ class ReflectionClass extends CoreReflectionClass
     }
 
     /**
-     * @return int
+     * @return string
      */
     public function getDocComment()
     {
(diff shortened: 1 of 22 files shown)


asgrim:
@Ocramius how do we feel about using `__call` to reduce boilerplate for these (before I go ahead an implement the rest...)?

Ocramius:
> @Ocramius how do we feel about using `__call` to reduce boilerplate for these (before I go ahead an implement the rest...)?

Better keep it ugly than using `__call`

src/Reflection/Adapter/ReflectionClass.php
```diff
@@ -0,0 +1,24 @@
...
+     * @var BetterReflectionClass
+     */
+    private $betterReflectionClass;
+
+    public function __construct(BetterReflectionClass $betterReflectionClass)
+    {
+        $this->betterReflectionClass = $betterReflectionClass;
+    }
+
+    public function __call($method, $args)
```
Ocramius: As discussed, this won't work

asgrim:
:grin:

src/Reflection/Adapter/ReflectionClass.php
```diff
@@ -0,0 +1,442 @@
...
+     */
+    private $betterReflectionClass;
+
+    public function __construct(BetterReflectionClass $betterReflectionClass)
+    {
+        $this->betterReflectionClass = $betterReflectionClass;
+    }
+
+    /**
+     * @return string
```
Ocramius: Use `{@inheritDoc}` plz :)

src/Reflection/Adapter/ReflectionClass.php
```diff
@@ -0,0 +1,442 @@
...
+     */
+    public function isInstance($object)
+    {
+        return $this->betterReflectionClass->isInstance($object);
+    }
+
+    /**
+     * @throws \Exception
+     */
+    public function newInstance($args = null, $_ = null)
```
Ocramius: That `$_` is simply a variadic. Simply use `func_get_args()`?

Ocramius: Also, this is runtime reflection: you will need to rely on the actual internal reflection at this point (or rely on `doctrine/instantiator`)

src/Reflection/Adapter/ReflectionClass.php
```diff
@@ -0,0 +1,442 @@
...
+    /**
+     * @return ReflectionProperty[]
+     */
+    public function getDefaultProperties()
+    {
+        $properties = $this->betterReflectionClass->getDefaultProperties();
+
+        $wrappedProperties = [];
+        foreach ($properties as $key => $property) {
+            $wrappedProperties[$key] = new ReflectionProperty($property);
```
Ocramius: You know that you can simply `array_map` this, right? <_<

src/Reflection/Adapter/ReflectionClass.php
```diff
@@ -0,0 +1,442 @@
...
+     */
+    public function newInstanceWithoutConstructor()
+    {
+        throw new Exception\NotImplemented('Not implemented');
+    }
+
+    /**
+     * @throws \Exception
+     */
+    public function newInstanceArgs(array $args = null)
```
Ocramius: Variadic

src/Reflection/Adapter/ReflectionFunction.php
```diff
@@ -0,0 +1,257 @@
...
+
+    /**
+     * @return ReflectionParameter[]
+     */
+    public function getParameters()
+    {
+        $parameters = $this->betterReflectionFunction->getParameters();
+
+        $wrappedParameters = [];
+        foreach ($parameters as $key => $parameter) {
```
Ocramius: Same here: can simply `array_map()`

src/Reflection/Adapter/ReflectionClass.php
asgrim: We don't implement `newInstance`, so implementation is irrelevant, see:
https://github.com/Roave/BetterReflection/blob/master/docs/compatibility.md#reflectionclass

src/Reflection/Adapter/ReflectionClass.php
asgrim: `array_map` does not maintain keys, and this is more readable

src/Reflection/Adapter/ReflectionFunction.php
asgrim: `array_map` does not maintain keys, and this is more readable

src/Reflection/Adapter/ReflectionClass.php
```diff
@@ -0,0 +1,433 @@
...
+     */
+    public function isInstance($object)
+    {
+        return $this->betterReflectionClass->isInstance($object);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function newInstance($args = null, $_ = null)
```
asgrim: @Ocramius it seems PHP7 complains about method signature not matching (see [build 347.2](https://travis-ci.org/Roave/BetterReflection/jobs/76068447#L241)). I've changed this back to avoid errors.

src/Reflection/Adapter/ReflectionClass.php
```diff
@@ -0,0 +1,433 @@
...
+    {
+        throw new Exception\NotImplemented('Not implemented');
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function getParentClass()
+    {
+        return new ReflectionClass($this->betterReflectionClass->getParentClass());
```
Ocramius: `$this->betterReflectionClass->getParentClass()` may be `null`

src/Reflection/Adapter/ReflectionClass.php
```diff
@@ -0,0 +1,433 @@
...
+     */
+    public function getDefaultProperties()
+    {
+        $properties = $this->betterReflectionClass->getDefaultProperties();
+
+        $wrappedProperties = [];
+        foreach ($properties as $key => $property) {
+            $wrappedProperties[$key] = new ReflectionProperty($property);
+        }
+        return $wrappedProperties;
```
Ocramius: IIRC, `getDefaultProperties()` returns the actual values. This code is currently causing `var_export does not handle circular references` in ProxyManager sources.

asgrim: You are actually correct, I implemented this incorrectly :/

Issue #100, opened 2015-08-13
Can compile predefined constants


asgrim:
We can actually reflect on probably "most" constants, certainly internal constants.

Simply check `defined('blah')` first, and then replace the value :)

Ocramius:
Approach:

 1. download shitload of stuff from https://svn.php.net/repository/phpdoc/en/trunk
 2. grep -rl "<refname>$CLASS::$METHOD</refname>"
 3. `xmllint --html --nowarning --xpath '//methodparam[@choice="opt"]//initializer' language/predefined/exception/construct.xml 2>/dev/null` (XPATH still to be cleaned)
 4. patch `Zend\Code` to actually catch `ReflectionException` when calling `$reflectionParameter->getDefaultValue()` and point it at the 3 steps above instead.

Ocramius:
Here's the `.sh` script to find the method parameters:

```sh
FINDMETHOD="mysqli_result::fetch_all"
FINDPARAMETER="resulttype"

find . -type f -iname "*.xml" | \
    xargs -I '{}' xmllint --noblanks --html --nowarning --xpath \
    "//refsect1[@role=\"description\" and //methodname[text()=\"$FINDMETHOD\"]]//methodsynopsis[@role=\"oop\"]//methodparam[//parameter[text()=\"$FINDPARAMETER\"]]//initializer/text()" '{}' \
    2>/dev/null
```

Ocramius:
Reporting progress.

I used following patched ZF2 class to compile classes with optional parameters via reflection + documentation:

```php
<?php
/**
 * Zend Framework (http://framework.zend.com/)
 *
 * @link      http://github.com/zendframework/zf2 for the canonical source repository
 * @copyright Copyright (c) 2005-2015 Zend Technologies USA Inc. (http://www.zend.com)
 * @license   http://framework.zend.com/license/new-bsd New BSD License
 */

namespace Zend\Code\Generator;

use BetterReflection\NodeCompiler\CompileNodeToValue;
use BetterReflection\NodeCompiler\CompilerContext;
use BetterReflection\Reflector\ClassReflector;
use BetterReflection\SourceLocator\StringSourceLocator;
use PhpParser\Node\Expr\Array_;
use PhpParser\Node\Expr\BinaryOp;
use PhpParser\Node\Expr\ClassConstFetch;
use PhpParser\Node\Expr\ConstFetch;
use PhpParser\Node\Stmt\Return_;
use PhpParser\ParserFactory;
[...]

PR #102, opened 2015-08-17
Evaluate the value of already-existing constants

asgrim:
#100 boom

### DIFF
--- a/test/unit/NodeCompiler/CompileNodeToValueTest.php
+++ b/test/unit/NodeCompiler/CompileNodeToValueTest.php
@@ -129,6 +129,8 @@ class CompileNodeToValueTest extends \PHPUnit_Framework_TestCase
             ['1 <= 2', true],
             ['2 <= 2', true],
             ['3 <= 2', false],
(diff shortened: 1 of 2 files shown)


test/unit/NodeCompiler/CompileNodeToValueTest.php
```diff
@@ -173,15 +175,28 @@ public function testExceptionThrownWhenSpaceshipOperatorUsed()
...
             UnableToCompileNode::class,
-            'Unable to compile constant expressions'
+            'Constant "FOO" has not been defined'
         );
         (new CompileNodeToValue())->__invoke(new ConstFetch(new Name('FOO')), $this->getDummyContext());
     }
 
+    public function testConstantValueCompiled()
+    {
+        define('BETTER_REFLECTION_TEST_CONST', 123);
```
Ocramius: This needs to pass through a `defined(...) || define(...)` to avoid issues if the test is executed twice. Consider using `uniqid()` instead?

Issue #10, opened 2015-07-01
Discover return types from DocBlocks


asgrim:
Using the DocBlocks, we can at least attempt to resolve the return types for functions and methods.

```php
/**
 * @return int|null
 */
public function aMethodReturningInt() { /* ... */ }
```

Implement a new method `getDocBlockReturnTypes()` which would return an array of `Type` objects. In this case, it would return something like a `[Types\Integer,Types\Null_]`

Ocramius:
Handled in #104

PR #104, opened 2015-08-18
Discover return types for functions and methods

asgrim:
See #10

### DIFF
--- a/src/Reflection/ReflectionFunctionAbstract.php
+++ b/src/Reflection/ReflectionFunctionAbstract.php
@@ -391,7 +391,7 @@ abstract class ReflectionFunctionAbstract implements \Reflector
      *
      * @return Type[]
      */
-    public function getReturnTypes()
+    public function getDocblockReturnTypes()
     {
         return  (new FindReturnType())->__invoke($this);
     }
--- a/src/Reflection/ReflectionFunctionAbstract.php
+++ b/src/Reflection/ReflectionFunctionAbstract.php
@@ -4,10 +4,12 @@ namespace BetterReflection\Reflection;
 
 use BetterReflection\Reflector\Reflector;
 use BetterReflection\SourceLocator\LocatedSource;
+use BetterReflection\TypesFinder\FindReturnType;
 use PhpParser\Node;
 use PhpParser\Node\Stmt as MethodOrFunctionNode;
 use PhpParser\Node\Stmt\Namespace_ as NamespaceNode;
 use PhpParser\Node\Expr\Yield_ as YieldNode;
+use phpDocumentor\Reflection\Type;
 
 abstract class ReflectionFunctionAbstract implements \Reflector
 {
@@ -379,5 +381,18 @@ abstract class ReflectionFunctionAbstract implements \Reflector
     public function returnsReference()
     {
         return (bool)$this->node->byRef;
+    }
+
+    /**
+     * Get the return types defined in the DocBlocks. This returns an array because
+     * the parameter may have multiple (compound) types specified (for example
+     * when you type hint pipe-separated "string|null", in which case this
+     * would return an array of Type objects, one for string, one for null.
+     *
+     * @return Type[]
+     */
+    public function getReturnTypes()
+    {
+        return  (new FindReturnType())->__invoke($this);
     }
 }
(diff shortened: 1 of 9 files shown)


src/TypesFinder/FindReturnType.php
```diff
@@ -0,0 +1,56 @@
...
+            )
+        );
+
+        $returnTags = $docBlock->getTagsByName('return');
+
+        foreach ($returnTags as $returnTag) {
+            /* @var $returnTag \phpDocumentor\Reflection\DocBlock\Tag\ReturnTag */
+            return (new ResolveTypes())->__invoke($returnTag->getTypes(), $context);
+        }
+        return [];
```
Ocramius: Should it be `mixed` by default?

src/Reflection/ReflectionFunctionAbstract.php
```diff
@@ -380,4 +382,17 @@ public function returnsReference()
...
+
+    /**
+     * Get the return types defined in the DocBlocks. This returns an array because
+     * the parameter may have multiple (compound) types specified (for example
+     * when you type hint pipe-separated "string|null", in which case this
+     * would return an array of Type objects, one for string, one for null.
+     *
+     * @return Type[]
+     */
+    public function getReturnTypes()
```
Ocramius: `getDocblockReturnTypes()`?

test/unit/Reflection/ReflectionFunctionTest.php
```diff
@@ -107,4 +108,22 @@ public function testStringCast($functionName, $expectedStringValue)
 
         $this->assertStringMatchesFormat($expectedStringValue, (string)$functionInfo);
     }
+
+    public function testGetReturnTypes()
+    {
+        $php = '<?php
```
Ocramius: You can use `<<<'PHP'`

test/unit/Reflection/ReflectionFunctionAbstractTest.php
```diff
@@ -276,4 +277,22 @@ public function testGetLocatedSource()
 
         $this->assertSame($locatedSource, $functionInfo->getLocatedSource());
     }
+
+    public function testGetReturnTypes()
+    {
+        $php = '<?php
```
Ocramius: You can use `<<<'PHP'`

test/unit/Reflection/ReflectionMethodTest.php
```diff
@@ -142,6 +143,18 @@ public function testMethodNameWithNamespace()
         $this->assertSame('someMethod', $methodInfo->getShortName());
     }
 
+    public function testGetReturnTypes()
+    {
+        $classInfo = $this->reflector->reflect('\BetterReflectionTest\Fixture\ExampleClass');
```
Ocramius: Backslash?

Ocramius: You can also use `::class`

src/TypesFinder/FindReturnType.php
```diff
@@ -0,0 +1,56 @@
...
+     * Given a function, attempt to find the return type.
+     *
+     * @param ReflectionFunctionAbstract $function
+     * @return Type[]
+     */
+    public function __invoke(ReflectionFunctionAbstract $function)
+    {
+        $context = $this->createContextForFunction($function);
+
+        $docBlock = new DocBlock(
```
Ocramius: Avoid assigning to a variable?

test/unit/Fixture/ExampleClass.php
```diff
@@ -30,6 +30,9 @@ public function __construct()
         {
         }
 
+        /**
+         * @return int
```
Ocramius: I suggest using a separate class. This asset is already used in other tests, and I indeed see some differences in other tests because of this.

src/TypesFinder/FindReturnType.php
asgrim: I don't think so, I don't think we should make any assumptions about return types. If they're there, we return them, if not, we don't. Plain and simple that way - if applications using Better Refletion want to make that assumption, they can do so simply.

test/unit/Fixture/ExampleClass.php
asgrim: That's okay, no? The tests still pass ;)

nclundsten:
@asgrim you trolling with assignees ?  :P

asgrim:
@nclundsten sorry - clicked the wrong name ^_^

nclundsten:
no prob, was just surprised when i started getting notifications

PR #115, opened 2015-08-30
Hotfix - evaled source locator should load only evaled classes

Ocramius:


Ocramius:
Depends on #116, #117, #118

Issue #43, opened 2015-07-07
StringSourceLocator doesn't "locate" source, but returns a LocatedSource object


asgrim:
Some conceptual "naming" issue that @Ocramius has... `StringSourceLocator` doesn't "locate" source, but still returns a `LocatedSource` object.

Solutions for this would be to either create a `Source` interface that `LocatedSource` and something else (that the `StringSourceLocator` would return) implement, or create an abstract parent.

Just need to clarify naming and intent

For reference, original discussion at https://github.com/Roave/BetterReflection/pull/32#discussion_r34014981

asgrim:
Will be fixed by resolving #123

Ocramius:
Handled in #123

PR #118, opened 2015-08-31
Hotfix - generic reflector should not care about locator implementation

Ocramius:
This is a partial fix for #43

This patch disables a test, specifically `BetterReflectionTest\Reflector\GenericTest#testReflectWithAggregateSourceLocatorFindsClass()`

That test uses multiple string locators on one line.

I believe it may be up to the `StringSourceLocator` to actually filter whether a particular symbol is included in the reflected source.

The `StringSourceLocator` may internally re-use the source locator logic somehow.

### DIFF
--- a/src/Reflector/ClassReflector.php
+++ b/src/Reflector/ClassReflector.php
@@ -5,6 +5,10 @@ namespace BetterReflection\Reflector;
 use BetterReflection\Identifier\Identifier;
 use BetterReflection\Identifier\IdentifierType;
 use BetterReflection\Reflector\Generic as GenericReflector;
+use BetterReflection\SourceLocator\AggregateSourceLocator;
+use BetterReflection\SourceLocator\AutoloadSourceLocator;
+use BetterReflection\SourceLocator\EvaledCodeSourceLocator;
+use BetterReflection\SourceLocator\PhpInternalSourceLocator;
 use BetterReflection\SourceLocator\SourceLocator;
 
 class ClassReflector implements Reflector
@@ -17,6 +21,15 @@ class ClassReflector implements Reflector
     public function __construct(SourceLocator $sourceLocator)
     {
         $this->reflector = new GenericReflector($sourceLocator);
+    }
+
+    public static function buildDefaultReflector()
+    {
+        return new self(new AggregateSourceLocator([
+            new PhpInternalSourceLocator(),
+            new EvaledCodeSourceLocator(),
+            new AutoloadSourceLocator(),
+        ]));
     }
 
     /**
--- a/src/Reflector/ClassReflector.php
+++ b/src/Reflector/ClassReflector.php
@@ -18,11 +18,17 @@ class ClassReflector implements Reflector
      */
     private $reflector;
 
+    /**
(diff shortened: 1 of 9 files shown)


src/Reflector/ClassReflector.php
```diff
@@ -14,12 +18,27 @@ class ClassReflector implements Reflector
...
+     */
     public function __construct(SourceLocator $sourceLocator)
     {
         $this->reflector = new GenericReflector($sourceLocator);
     }
 
     /**
+     * @return self
+     */
+    public static function buildDefaultReflector()
```
asgrim: Just create a new `DefaultSourceLocator` for this purpose imo

test/unit/Reflector/GenericTest.php
```diff
@@ -160,6 +160,8 @@ public function testReflectWithAggregateSourceLocatorWhenIdentifierDoesNotExist(
 
     public function testReflectWithAggregateSourceLocatorFindsClass()
     {
+        $this->markTestIncomplete('See https://github.com/Roave/BetterReflection/issues/43');
```
asgrim: Please add further information to #43 as unsure as to the relevance? #43 is (currently) about naming as I see it

asgrim: Also I created #123 - this may be more relevant imo

src/Reflector/ClassReflector.php
Ocramius: @asgrim it would extend `AggregateSourceLocator` without any other difference with it. I think the factory approach is fine: maybe move it to a separate class? Couldn't find any though

Ocramius: `ClassReflectorBuilder`?

test/unit/Reflector/GenericTest.php
Ocramius: The point is that string source locators don't "locate", but "provide". If they were to "locate" (and therefore return `null` or a `LocatedSource`) then this would be fine.

Ocramius: Indeed, I'll link #123 here

src/Reflector/Generic.php
```diff
@@ -41,33 +41,18 @@ public function __construct(SourceLocator $sourceLocator)
...
      *
      * @param Identifier $identifier
-     * @param bool $autoReflectInternals
+     *
      * @return Reflection
      */
-    public function reflect(Identifier $identifier, $autoReflectInternals = true)
+    public function reflect(Identifier $identifier)
     {
         $aggregate = $this->sourceLocator;
```
asgrim: If the "aggregate" logic has been abstracted into the `AggregateSourceLocator` (which is better) this should be named more appropriately (it is not definitely an `Aggregate` any more)

Ocramius: Agreed, fixing.

src/Reflector/ClassReflector.php
asgrim: meh, leave it for now, if it becomes a requirement later we can revisit

PR #120, opened 2015-08-31
Hotfix - #119 - workaround for stubbing internal and evaluated sources

### LINKED ISSUE #119: `PhpInternalSourceLocator` and `EvaledCodeSourceLocator` can only stub classes
As it stands, `PhpInternalSourceLocator` and `EvaledCodeSourceLocator` can only stub classes.

This is a limitation as per zendframework/zend-code#7

To fix it (for now), a simple solution is to apply a transformation to the output of `zendframework/zend-code`, but it is not sufficiently solid/stable.

Ocramius:
Should fix #119 for now.

### DIFF
--- a/src/SourceLocator/SourceStubber.php
+++ b/src/SourceLocator/SourceStubber.php
@@ -46,11 +46,23 @@ final class SourceStubber
             return $stubCode;
         }
 
-        $statements = $this->parser->parse('<?php ' . $stubCode);
+        return $this->prettyPrinter->prettyPrint(
+            $this->replaceNodesRecursively($this->parser->parse('<?php ' . $stubCode), $interface)
+        );
+    }
 
+    /**
+     * @param \PhpParser\Node[] $statements
+     * @param bool              $interfaceOrTrait (true => interface, false => trait)
+     * @param bool              $trait
+     *
+     * @return \PhpParser\Node[]
+     */
+    private function replaceNodesRecursively(array $statements, $interfaceOrTrait)
+    {
         foreach ($statements as $key => $statement) {
             if ($statement instanceof Class_) {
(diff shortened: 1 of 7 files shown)


src/SourceLocator/SourceStubber.php
```diff
@@ -0,0 +1,85 @@
...
+
+    /**
+     * @param ClassReflection $reflection
+     *
+     * @return string
+     */
+    public function __invoke(ClassReflection $reflection)
+    {
+        $stubCode   = ClassGenerator::fromReflection($reflection)->generate();
+        $interface  = $reflection->isInterface();
```
asgrim: Naming: imo, booleans should be `$isInterface` as communicates to the reader this should be `bool`

src/SourceLocator/SourceStubber.php
```diff
@@ -0,0 +1,85 @@
...
+        $trait      = $reflection->isTrait();
```
asgrim: as above, `$isTrait`

src/SourceLocator/SourceStubber.php
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    /**
+     * @param \PhpParser\Node[] $statements
+     * @param bool              $interfaceOrTrait (true => interface, false => trait)
+     * @param bool              $trait
+     *
+     * @return \PhpParser\Node[]
+     */
+    private function replaceNodesRecursively(array $statements, $interfaceOrTrait)
```
asgrim: Naming here - this is actually `$isInterface`

Ocramius: I meant to make it `$isInterfaceOfTrait`

PR #122, opened 2015-08-31
Hotfix - #121 - reflectionclass getmethods should return only visibile methods

### LINKED ISSUE #121: `ReflectionClass#getMethods()` should only return methods visible to the current class
See https://3v4l.org/ZgpTn

```php
<?php

class A {private function b() {}}
class C {private function d() {}}

var_dump((new ReflectionClass(C::class))->getMethods());
```

```
array(1) {
  [0]=>
  object(ReflectionMethod)#2 (2) {
    ["name"]=>
    string(1) "d"
    ["class"]=>
    string(1) "C"
  }
}
```

The current behavior (incorrect) returns all methods from any ancestor, and with any visibility.

Ocramius:
Should fix #121


Ocramius:
Invalid test case.

PR #124, opened 2015-08-31
Add DefaultSourceLocator

asgrim:
Suggestion to merge into `hotfix/generic-reflector-should-not-care-about-locator-implementation` PR #118

### DIFF
--- a/src/SourceLocator/DefaultSourceLocator.php
+++ b/src/SourceLocator/DefaultSourceLocator.php
@@ -1,0 +1,20 @@
+<?php
+
+namespace BetterReflection\SourceLocator;
+
+/**
+ * A simplification to avoid having to define the same three source locators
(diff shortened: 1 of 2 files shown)


src/SourceLocator/DefaultSourceLocator.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+namespace BetterReflection\SourceLocator;
+
+/**
+ * A simplification to avoid having to define the same three source locators
+ * all over the place. This should be able to locate most things in common
+ * environments.
+ */
+class DefaultSourceLocator extends AggregateSourceLocator
```
Ocramius: Disagree on giving a default locator a new type, but if we go down this road:

 - implement `LocatorInterface`
 - don't inherit
 - build an `AggregateSourceLocator` internally

PR #125, opened 2015-09-02
Add the package license for composer

stof:
the license file is available in the repo, but the info was missing in the composer.json, marking it as an unknown license on Packagist


Ocramius:
:+1:

PR #129, opened 2015-09-17
added identifier accessor for IdentifierNotFoundException

digitalkaoz:
EOM

### DIFF
--- a/src/Reflector/Exception/IdentifierNotFound.php
+++ b/src/Reflector/Exception/IdentifierNotFound.php
@@ -6,12 +6,32 @@ use BetterReflection\Identifier\Identifier;
 
 class IdentifierNotFound extends \RuntimeException
 {
+    /**
+     * @var Identifier
+     */
+    private $identifier;
+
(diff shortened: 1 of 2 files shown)


src/Reflector/Exception/IdentifierNotFound.php
```diff
@@ -6,12 +6,32 @@
 
 class IdentifierNotFound extends \RuntimeException
 {
+    /**
+     * @var Identifier
+     */
+    private $identifier;
+
+    public function __construct($message, $identifier)
```
asgrim: please type hint `$identifier`

digitalkaoz: done

Ocramius:
What's the exact use-case for this, in a consumer?

digitalkaoz:
yes i use this to get the "unloadable" class back from the Exception to do further things. parsing it from the message felt a bit ugly

asgrim:
@Ocramius I have no problem with this, I'm gonna merge

Issue #130, opened 2015-10-11
Fetch AST or Code for a method/function


asgrim:
Add new methods to `ReflectionFunctionAbstract` which fetch AST or code representation:
* `ReflectionFunctionAbstract::getAst()` - returns the AST for the code within the function
* `ReflectionFunctionAbstract::getBodyCode()` returns the code of the function body itself (using `PhpParser`'s printer functionality)

Ocramius:
:+1: VERY (***VERY***) useful!

asgrim:
Note the `getBodyCode` would not get the surrounding syntax, it would only get the actual body of the function. See #133 for exporting a whole reflection. For example, given you have a reflection of this method:

```php
private function add($a, $b)
{
    return $a + $b;
}
```

Then calling `getBodyCode()` would result in just:

```php
return $a + $b;
```

asgrim:
Fixed by #135

PR #135, opened 2015-10-18
Add returning AST methods

### LINKED ISSUE #133: Allow exporting a reflection to PHP
A pretty powerful feature of BR would be to export a reflection to PHP code (using PhpParser's printing functionality. This would pretty much take the whole AST stored within a reflection, feed it to PhpParser's printer, and return it.

Something like:

```php
echo $reflectedClass->getMethod('add')->asPhp();
```

would result in a string with the following:

```php
/**
 * Some addition method
 *
 * @param int $a
 * @param int $a
 * @return int
 */
private function add($a, $b)
{
    return $a + $b;
}
```

Note that because we exported the method only, it is not valid PHP by itself (it should exist within the context of a class).

Exporting a class would be more "complete":

```php
echo $reflectedClass->asPhp();
```

might result in something like:

```
class MathematicalStuff
{
    /**
     * Some addition method
     *
     * @param int $a
     * @param int $a
     * @return int
     */
    public function add($a, $b)
    {
        return $a + $b;
    }
}
```

*Note: not sure if the docblocks output is possible, but it would be great to include this*

asgrim:
Would resolve #133 and #130 

This PR adds new methods:
 - `getAst` for `ReflectionClass` and `ReflectionFunctionAbstract` (#133)
 - `getBodyCode` and `getBodyAst` for `ReflectionFunctionAbstract` (#130)

### DIFF
--- a/docs/ast-extraction.md
+++ b/docs/ast-extraction.md
@@ -1,0 +1,47 @@
+# Extracting AST from reflections
+
+## Method and Function body AST
+
+Given a `ReflectionMethod` or `ReflectionFunction`, you can extract the content
+(within curly braces) as an Abstract Syntax Tree, or as unparsed AST (i.e. as
+normal code). This allows you to easily grab the AST from a specific method with
+very few lines of code:
+
+```php
+<?php
+
+use BetterReflection\Reflector\ReflectionClass;
+
+$classInfo = ReflectionClass::createFromName('Foo\Bar\MyClass');
+
+// Retrieves the AST statements array *within* the method's curly braces
+$ast = $classInfo->getMethod('foo')->getBodyAst();
+$php = $classInfo->getMethod('foo')->getBodyCode();
+```
+
+### Note on extracting code using `getBodyCode()`
+
+The code returned by the `getBodyCode()` method is pretty-printed using the
+AST unparser in the `PhpParser` library. This means the actual code returned
+here may be laid out differently to the original function or method. See the
+code documentation for details on how to provide a custom printer to override
+the default behaviour.
+
+## Fetching AST representation of a class or function
+
+It is possible to fetch an AST representation of a class or function using the
+`getAst()` method on a `ReflectionClass` and `ReflectionFunction`.
+
+```php
+<?php
+
+use BetterReflection\Reflector\ReflectionClass;
+
+$classInfo = ReflectionClass::createFromName('Foo\Bar\MyClass');
+
+// Retrieves AST nodes for the entire class (including the class definition)
+$ast = $classInfo->getAst();
+```
+
+Note that the node returned will be the top level of the class, thus including
+the actual class definition itself.
--- a/docs/ast-extraction.md
+++ b/docs/ast-extraction.md
@@ -12,7 +12,7 @@ very few lines of code:
 
 use BetterReflection\Reflector\ReflectionClass;
 
-$classInfo = ReflectionClass::createFromName('Foo\Bar\MyClass');
+$classInfo = ReflectionClass::createFromName(\Foo\Bar\MyClass::class);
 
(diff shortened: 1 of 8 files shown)


docs/ast-extraction.md
```diff
@@ -0,0 +1,47 @@
...
+(within curly braces) as an Abstract Syntax Tree, or as unparsed AST (i.e. as
+normal code). This allows you to easily grab the AST from a specific method with
+very few lines of code:
+
+```php
+<?php
+
+use BetterReflection\Reflector\ReflectionClass;
+
+$classInfo = ReflectionClass::createFromName('Foo\Bar\MyClass');
```
Ocramius: Consider using `::class` in examples

docs/ast-extraction.md
```diff
@@ -0,0 +1,47 @@
...
+
+It is possible to fetch an AST representation of a class or function using the
+`getAst()` method on a `ReflectionClass` and `ReflectionFunction`.
+
+```php
+<?php
+
+use BetterReflection\Reflector\ReflectionClass;
+
+$classInfo = ReflectionClass::createFromName('Foo\Bar\MyClass');
```
Ocramius: Consider using `::class` in examples

src/Reflection/ReflectionFunctionAbstract.php
```diff
@@ -406,4 +408,46 @@ public function __clone()
     {
         throw Exception\Uncloneable::fromClass(__CLASS__);
     }
+
+    /**
+     * Retrieves the body of this function as AST nodes
+     *
+     * @return array
```
Ocramius: `Node[]`?

src/Reflection/ReflectionClass.php
```diff
@@ -1036,4 +1036,14 @@ public function __clone()
     {
         throw Exception\Uncloneable::fromClass(__CLASS__);
     }
+
+    /**
+     * Retrieve the AST node for this class
+     *
+     * @return ClassNode|InterfaceNode|TraitNode|ClassLikeNode
```
Ocramius: `ClassLikeNode` already provides all the others

src/Reflection/ReflectionFunctionAbstract.php
```diff
@@ -406,4 +408,46 @@ public function __clone()
...
+     * a default will be used.
+     *
+     * Note that the formatting of the code may not be the same as the original
+     * function. If specific formatting is required, you should provide your
+     * own implementation of a PrettyPrinter to unparse the AST.
+     *
+     * @param PrettyPrinterAbstract|null $printer
+     * @return string
+     */
+    public function getBodyCode(PrettyPrinterAbstract $printer = null)
```
Ocramius: Is this one actually needed? I wouldn't pretty-print: the consumer can do it, no?

Ocramius:
@asgrim needs rebase

src/Reflection/ReflectionFunctionAbstract.php
asgrim: It's just a helper, I think it's useful sometimes

docs/ast-extraction.md
asgrim: Fixed in 27025f3

docs/ast-extraction.md
asgrim: Fixed in 27025f3

src/Reflection/ReflectionFunctionAbstract.php
asgrim: Fixed in 27025f3

src/Reflection/ReflectionClass.php
asgrim: Fixed in 27025f3

PR #139, opened 2015-10-27
More PHPT tests

### LINKED ISSUE #12: Add core reflection .phpt tests to ensure compatibility
For compatibility, we could run tests vs the internal reflection to ensure the same results are achieved for maximum awesometowns.

asgrim:
This aims to implement as many PHPT tests as possible to resolve #12

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ before_script:
 
 script:
   - vendor/bin/phpunit --coverage-text --coverage-clover=clover.xml --colors
+  - vendor/bin/phpunit test/compat || true
   - vendor/bin/phpunit test/core || true
 
 after_script:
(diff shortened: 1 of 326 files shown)


Ocramius:
@asgrim still `[WIP]`?

asgrim:
@Ocramius yes, there's loads more tests to add....

asgrim:
@Ocramius Right, most of the compat tests are failing, but they're in now. There's numerous issues to go through, I've created just a couple of bugs for now, but this is a huge undertaking... Getting these `.phpt` tests to all be green for a `1.0.0` release is totally unrealistic imo...

asgrim:
(I removed the .phpt test from running on normal test suite, so you must run manually using `vendor/bin/phpunit test/compat` btw)

PR #140, opened 2015-10-28
Fixed composer.json

GrahamCampbell:
`@dev` does absolutely nothing because of your prefer-stable directive, and `dev-master` should be avoided at all costs.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -3,11 +3,11 @@
     "description": "Better Reflection - an improved code reflection API",
     "license": "MIT",
     "require": {
-        "php": ">= 5.6",
-        "nikic/php-parser": "dev-master",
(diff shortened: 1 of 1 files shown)


asgrim:
We're fine with `dev-master` for now thanks, we are not stable yet ;)

GrahamCampbell:
It's not about being stable. It's about dev-master being the same as an unbounded dependency, and my vesion constraints load exactly the same version of the code you want.

asgrim:
Yes, I am fully aware of that.

Ocramius:
@GrahamCampbell we simply will have to tell users (for now) to use a `"minimum-stability": "dev"` until we reach stability at all.

Issue #134, opened 2015-10-12
Add methods to modify reflections


asgrim:
It would be pretty cool to be able to modify a reflection, for example:

```php
$newProperty = ReflectionProperty::create('bar', \ReflectionProperty::IS_PRIVATE | \ReflectionProperty::IS_STATIC);
$reflectionClass->addProperty($newProperty);
```

This is a powerful feature when combined with the ability to export a reflection to PHP code (see #133).

asgrim:
*Note: this sort of modification would modify the attached AST*

Possible ideas for modifications:

* Class::add/removeProperty
* FunctionAbstract::add/removeParameter
* Property::setVisibility
* (add more ideas here later)

olimorris:
The `addMethod` idea would be very special. Unless you install runkit on your server, there's no neat way of actually monkey patching a method at runtime in PHP.

In terms of use cases, I'm using a third party package and need to add some specific functionality to the core of the package (namely two methods). A pull request would be too opinionated and a general fork would take me out of sync from the rest of the package.

olimorris:
@asgrim this is brilliant. Great job.

PR #141, opened 2015-10-28
Method body replacement

asgrim:
Solves part of #134  (ability to replace the statements within a function or method)

~~Note this requires PR #135 to be merged first, so do not merge please.~~

~~Also depends on jeremeamia/super_closure#70~~ (stable in 2.2.0+ now, which is used by our master now anyway)

### DIFF
--- a/src/Reflection/ReflectionFunctionAbstract.php
+++ b/src/Reflection/ReflectionFunctionAbstract.php
@@ -492,39 +492,41 @@ abstract class ReflectionFunctionAbstract implements \Reflector
 
     /**
      * Override the method or function's body of statements with an entirely new
-     * body of statements (!) within the reflection.
-     *
-     * You may either pass a string containing statements, or a closure:
+     * body of statements within the reflection.
      *
      * @example
-     * // The following are identical...
-     * $reflectionFunction->setBody('return true');
-     * $reflectionFunction->setBody(function () { return true; });
+     * $reflectionFunction->setBodyFromClosure(function () { return true; });
      *
-     * @param string|Closure $newBody
+     * @param \Closure $newBody
      */
-    public function setBody($newBody)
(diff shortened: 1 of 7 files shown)


composer.json
```diff
@@ -7,7 +7,8 @@
         "nikic/php-parser": "dev-master",
         "phpdocumentor/reflection-docblock": "^2.0",
         "phpdocumentor/type-resolver": "dev-master",
-        "zendframework/zend-code": "^2.5@dev"
+        "zendframework/zend-code": "^2.5@dev",
+        "jeremeamia/superclosure": "dev-php-parser"
```
asgrim: Needs replacing before merge

asgrim: (the version, not the whole dependency)

src/Reflection/ReflectionFunctionAbstract.php
```diff
@@ -398,4 +404,84 @@ public function getDocBlockReturnTypes()
...
+     * body of statements (!) within the reflection.
+     *
+     * You may either pass a string containing statements, or a closure:
+     *
+     * @example
+     * // The following are identical...
+     * $reflectionFunction->setBody('return true');
+     * $reflectionFunction->setBody(function () { return true; });
+     *
+     * @param string|Closure $newBody
```
Ocramius: I would disallow mixed here, and instead have two separate endpoints

Ocramius: Would also be awesome to just allow setting AST :-P

Ocramius:
Note: I would not merge this until we first have a stable (tagged) API with the read-only part of better-reflection

asgrim:
@Ocramius agreed. I just created `1.1.0` milestone for future releases

asgrim:
@asgrim worth rebasing and looking at comments after #151 is merged

PR #147, opened 2015-11-05
Added demo folder

asgrim:
Just a folder with code that should work with current version of the library that demonstrates usage in an isolated way


demo/basic-reflection/index.php
```diff
@@ -0,0 +1,46 @@
+<?php
+
+// Load a standard (internal) class
```
Ocramius: These should be separate files IMO

asgrim: done, yo

Ocramius: @asgrim merged, yo

asgrim:
thanks

PR #149, opened 2015-11-07
Use branch-alias instead of dev-master

EmanueleMinotto:
The `dev-master` could change in every moment, instead the `branch-alias` shouldn’t change as fast as the previous.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -4,9 +4,9 @@
     "license": "MIT",
     "require": {
         "php": ">= 5.6",
-        "nikic/php-parser": "dev-master",
+        "nikic/php-parser": "^2.0@dev",
         "phpdocumentor/reflection-docblock": "^2.0",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -4,9 +4,9 @@
     "license": "MIT",
     "require": {
         "php": ">= 5.6",
-        "nikic/php-parser": "dev-master",
+        "nikic/php-parser": "^2.0@dev",
```
Ocramius: We should probably rely on the beta

EmanueleMinotto:
@Ocramius I'm not fully sure about versions after the `-[stability]` part, so I've added `^2.0-beta` considering these results: http://semver.mwl.be/#?package=nikic%2Fphp-parser&version=%5E2.0-beta&minimum-stability=dev

Ocramius:
@EmanueleMinotto thanks!

Issue #61, opened 2015-07-14
Reflection* classes should be uncloneable


asgrim:
Reflection* classes should prevent being able to be cloned.

```php
--TEST--
Reflection class can not be cloned
--CREDITS--
Stefan Koopmanschap <stefan@phpgg.nl>
TestFest PHP|Tek
--SKIPIF--
<?php
if (!extension_loaded('reflection')) print 'skip';
?>
--FILE-- 
<?php
require 'vendor/autoload.php';
use BetterReflection\Reflector\ClassReflector;
use BetterReflection\SourceLocator\SingleFileSourceLocator;
$reflector = new ClassReflector(new SingleFileSourceLocator(__FILE__));

class Foo {
}

$rc = $reflector->reflect("Foo");
$rc2 = clone($rc);
?>
--EXPECTF--
Fatal error: Uncaught Error: Trying to clone an uncloneable object of class ReflectionClass in %s:%d
Stack trace:
#0 {main}
  thrown in %s on line %d
```

Ocramius:
@asgrim should they be? :|

asgrim:
seems that way...

asgrim:
Handled by #150

PR #150, opened 2015-11-14
Make reflections uncloneable

asgrim:
As identified in #61, the Reflections should be uncloneable...

### DIFF
--- a/src/Reflection/ReflectionClass.php
+++ b/src/Reflection/ReflectionClass.php
@@ -1034,6 +1034,6 @@ class ReflectionClass implements Reflection, \Reflector
      */
     public function __clone()
     {
-        throw new Exception\Uncloneable('Trying to clone an uncloneable object of class ' . __CLASS__);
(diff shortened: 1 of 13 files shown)


src/Reflection/ReflectionClass.php
```diff
@@ -1028,4 +1028,12 @@ function (Node\Name $interfaceName) {
             )
         );
     }
+
+    /**
+     * @throws Exception\Uncloneable
+     */
+    public function __clone()
+    {
+        throw new Exception\Uncloneable('Trying to clone an uncloneable object of class ' . __CLASS__);
```
Ocramius: Each of these strings are duplicate: instead, create a public static constructor on `Uncloneable` and call it from all around the various `__clone` methods

asgrim: @Ocramius sorted

Ocramius:
:+1:

Issue #37, opened 2015-07-05
Add ability to create a ReflectionFunction from a closure


asgrim:
Currently, because we don't load "live" (loaded-into-memory) code, we can't create a `ReflectionFunction` using a closure (because we actually don't have source code for it, thus can't create the AST from it). #36 may help to implement this, but because we don't have source code for closures, I'm not even sure if this is possible (without using core Reflection API and hacking it in...?)

asgrim:
can use https://github.com/jeremeamia/super_closure for this

asgrim:
though depends on jeremeamia/super_closure#70 to be compatible

asgrim:
Handled by #151

PR #151, opened 2015-11-26
Reflect from Closures

asgrim:
Handles issue #37 by adding ability to reflect on anonymous functions / closures.

### DIFF
--- a/src/Reflection/ReflectionFunction.php
+++ b/src/Reflection/ReflectionFunction.php
@@ -5,8 +5,8 @@ namespace BetterReflection\Reflection;
 use BetterReflection\Reflector\FunctionReflector;
 use BetterReflection\Reflector\Reflector;
 use BetterReflection\SourceLocator\Type\AutoloadSourceLocator;
-use BetterReflection\SourceLocator\Located\LocatedSource;
 use BetterReflection\SourceLocator\Type\ClosureSourceLocator;
+use BetterReflection\SourceLocator\Located\LocatedSource;
 use PhpParser\Node\FunctionLike as FunctionNode;
 use PhpParser\Node\Stmt\Namespace_ as NamespaceNode;
 
--- a/src/Reflection/ReflectionFunction.php
+++ b/src/Reflection/ReflectionFunction.php
@@ -6,7 +6,7 @@ use BetterReflection\Reflector\FunctionReflector;
 use BetterReflection\Reflector\Reflector;
 use BetterReflection\SourceLocator\Type\AutoloadSourceLocator;
 use BetterReflection\SourceLocator\Located\LocatedSource;
-use PhpParser\Node\Stmt\Function_ as FunctionNode;
+use PhpParser\Node\FunctionLike as FunctionNode;
 use PhpParser\Node\Stmt\Namespace_ as NamespaceNode;
 
 class ReflectionFunction extends ReflectionFunctionAbstract implements Reflection
--- a/src/Reflection/ReflectionFunction.php
+++ b/src/Reflection/ReflectionFunction.php
@@ -6,6 +6,7 @@ use BetterReflection\Reflector\FunctionReflector;
 use BetterReflection\Reflector\Reflector;
 use BetterReflection\SourceLocator\Type\AutoloadSourceLocator;
 use BetterReflection\SourceLocator\Located\LocatedSource;
+use BetterReflection\SourceLocator\Type\ClosureSourceLocator;
 use PhpParser\Node\FunctionLike as FunctionNode;
 use PhpParser\Node\Stmt\Namespace_ as NamespaceNode;
 
@@ -18,6 +19,15 @@ class ReflectionFunction extends ReflectionFunctionAbstract implements Reflectio
     public static function createFromName($functionName)
     {
         return (new FunctionReflector(new AutoloadSourceLocator()))->reflect($functionName);
+    }
+
+    /**
+     * @param \Closure $closure
+     * @return ReflectionFunction
+     */
+    public static function createFromClosure(\Closure $closure)
+    {
+        return (new FunctionReflector(new ClosureSourceLocator($closure)))->reflect('{closure}');
     }
 
     /**
--- a/src/SourceLocator/Type/ClosureSourceLocator.php
+++ b/src/SourceLocator/Type/ClosureSourceLocator.php
@@ -78,6 +78,7 @@ class ClosureSourceLocator implements SourceLocator
 
     public function locateIdentifiersByType(Reflector $reflector, IdentifierType $identifierType)
     {
-        // @TODO: Implement locateIdentifiersByType() method.
+        // @todo https://github.com/Roave/BetterReflection/issues/152
+        throw new \LogicException('Not implemented - Unable to reflect closures in this way currently');
     }
 }
--- a/src/SourceLocator/Type/ClosureSourceLocator.php
+++ b/src/SourceLocator/Type/ClosureSourceLocator.php
@@ -9,6 +9,8 @@ use BetterReflection\Reflector\Reflector;
 use BetterReflection\SourceLocator\Ast\Strategy\AstConversionStrategy;
 use BetterReflection\SourceLocator\Ast\Strategy\NodeToReflection;
 use BetterReflection\SourceLocator\Located\LocatedSource;
+use PhpParser\Node\Name;
+use PhpParser\Node\Stmt\Namespace_;
 use SuperClosure\Analyzer\AstAnalyzer;
 use SuperClosure\Analyzer\ClosureAnalyzer;
 use PhpParser\Node\Expr\Closure as ClosureNode;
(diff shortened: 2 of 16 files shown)


src/Reflection/ReflectionFunction.php
```diff
@@ -21,6 +22,15 @@ public static function createFromName($functionName)
     }
 
     /**
+     * @param \Closure $closure
+     * @return ReflectionFunction
+     */
+    public static function createFromClosure(\Closure $closure)
+    {
+        return (new FunctionReflector(new ClosureSourceLocator($closure)))->reflect('{closure}');
```
Ocramius: Does the fact that a closure source locator is used limit reflection of type hints?

As in:
`$c = function (Foo $foo) : Bar {}`

Can I get reflection for `$foo` and for the return parameter even if a `ClosureSourceLocator` was used?

src/SourceLocator/Type/ClosureSourceLocator.php
```diff
@@ -0,0 +1,83 @@
...
+            $reflector,
+            $closureData['ast'],
+            $locatedSource,
+            $namespaceNode
+        );
+    }
+
+    public function locateIdentifiersByType(Reflector $reflector, IdentifierType $identifierType)
+    {
+        // @TODO: Implement locateIdentifiersByType() method.
```
Ocramius: What about this TODO? Maybe throw an exception for now? Also, is all the interface covered? If so: `final`

src/SourceLocator/Type/ClosureSourceLocator.php
```diff
@@ -0,0 +1,83 @@
...
+use BetterReflection\SourceLocator\Ast\Strategy\AstConversionStrategy;
+use BetterReflection\SourceLocator\Ast\Strategy\NodeToReflection;
+use BetterReflection\SourceLocator\Located\LocatedSource;
+use PhpParser\Node\Name;
+use PhpParser\Node\Stmt\Namespace_;
+use SuperClosure\Analyzer\AstAnalyzer;
+use SuperClosure\Analyzer\ClosureAnalyzer;
+use PhpParser\Node\Expr\Closure as ClosureNode;
+
+class ClosureSourceLocator implements SourceLocator
```
Ocramius: Is all the interface covered? If so: `final`

test/unit/Reflection/ReflectionFunctionAbstractTest.php
```diff
@@ -42,7 +43,7 @@ public function testPopulateFunctionAbstractThrowsExceptionWithInvalidNode()
 
         $this->setExpectedException(
             \InvalidArgumentException::class,
-            'Node parameter must be ClassMethod or Function_'
+            'Node parameter must be ClassMethod or FunctionLike'
```
Ocramius: I suggest using `sprintf` and `::class` here

Ocramius: Probably better if FQCN is expected here

test/unit/Reflection/ReflectionFunctionTest.php
```diff
@@ -81,13 +81,22 @@ public function testIsUserDefined()
...
 
-    public function testStaticCreation()
+    public function testStaticCreationFromName()
     {
         require_once(__DIR__ . '/../Fixture/Functions.php');
         $reflection = ReflectionFunction::createFromName('BetterReflectionTest\Fixture\myFunction');
         $this->assertSame('myFunction', $reflection->getShortName());
     }
 
+    public function testStaticCreationFromClosure()
```
Ocramius: Question: what happens with two closures on the same line? Can you please try adding a test case with that scenario (I know it will fail, but at least we'd have a documented limitation)

test/unit/SourceLocator/Type/ClosureSourceLocatorTest.php
```diff
@@ -0,0 +1,44 @@
...
+            $this->getMockReflector(),
+            new Identifier(
+                'Foo',
+                new IdentifierType(IdentifierType::IDENTIFIER_FUNCTION)
+            )
+        );
+
+        $this->assertSame('{closure}', $reflection->getShortName());
+        $this->assertContains('Hello world!', $reflection->getLocatedSource()->getSource());
+    }
```
Ocramius: Question: what happens with two closures on the same line? Can you please try adding a test case with that scenario (I know it will fail, but at least we'd have a documented limitation)

test/unit/SourceLocator/Type/ClosureSourceLocatorTest.php
```diff
@@ -0,0 +1,44 @@
...
+{
+    /**
+     * @return Reflector|\PHPUnit_Framework_MockObject_MockObject
+     */
+    private function getMockReflector()
+    {
+        return $this->getMock(Reflector::class);
+    }
+
+    public function testClosureSourceLocator()
```
Ocramius: Misses tests for the incomplete method

test/unit/SourceLocator/Type/ClosureSourceLocatorTest.php
asgrim: It should not matter, in the constructor you pass the actual `\Closure` instance you want to reflect on, so it'll never actually pull back the wrong closure, unless I misunderstood?

src/Reflection/ReflectionFunction.php
asgrim: Fixed in 78c3f78

src/SourceLocator/Type/ClosureSourceLocator.php
asgrim: Fixed in b0e899d

src/SourceLocator/Type/ClosureSourceLocator.php
asgrim: Fixed in 2a2668f

test/unit/Reflection/ReflectionFunctionAbstractTest.php
asgrim: Fixed in d3e2e66

test/unit/SourceLocator/Type/ClosureSourceLocatorTest.php
asgrim: Fixed in b0e899d

src/Reflection/ReflectionFunction.php
asgrim: Fixed in 78c3f78

src/SourceLocator/Type/ClosureSourceLocator.php
asgrim: Fixed in b0e899d

src/SourceLocator/Type/ClosureSourceLocator.php
asgrim: Fixed in 2a2668f

test/unit/Reflection/ReflectionFunctionAbstractTest.php
asgrim: Fixed in d3e2e66

test/unit/SourceLocator/Type/ClosureSourceLocatorTest.php
asgrim: Fixed in b0e899d

test/unit/SourceLocator/Type/ClosureSourceLocatorTest.php
Ocramius: It will actually break, because superclosure uses the definition line to recognize a closure (php internal API does not provide the column)

asgrim: Damnit @Ocramius why are you always right...

Added an expectation test in d5b4034

Ocramius: @asgrim I really suck at making things easy for implementors :+1:

test/unit/SourceLocator/Type/ClosureSourceLocatorTest.php
```diff
@@ -0,0 +1,77 @@
...
+        );
+    }
+
+    public function testTwoClosuresSameLineFails()
+    {
+        $closure1 = function () {}; $closure2 = function () {};
+
+        $locator = new ClosureSourceLocator($closure1);
+
+        $this->setExpectedException(ClosureAnalysisException::class);
```
Ocramius: Should probably bubble up as an exception in our own namespace?

asgrim: Is there much benefit? The exception message is very clear, we'd only be wrapping it anyway - plus if this issue goes away, then this test will start failing anyway (alerting us that we can change this expectation...)

Ocramius: That is true, but making an exception from a different namespace bubble up (for a known case) will cause the consumers to type-hint against the third-party lib (without importing it manually, which is a mistake), and we would have to document all `@throws` with third-party lib classes (again bad)

asgrim: Fixed in fbbbe7c.

Ocramius:
Dang, totally forgot to merge this, sorry :|

asgrim:
Don't worry, I did it :D

PR #158, opened 2016-01-31
Run against lowest dependencies as well

Ocramius:


asgrim:
Possibly need to increase the minimum required version of Zend\Code? Only briefly looked but looks related to that.

PR #165, opened 2016-02-06
Fix coding standards with fix-cs.sh script

asgrim:
I took out `phpdoc_short_description` fixer because reasons.


Ocramius:
`because reasons` LGTM :+1:

PR #166, opened 2016-02-06
Fix locateIdentifiersByType to correctly aggregate source

AydinHassan:
Hi, 

It seems the aggregate source locator was not working as `+=` ignores keys that already exist in left hand array:

> The + operator returns the right-hand array appended to the left-hand array; for keys that exist in both arrays, the elements from the left-hand array will be used, and the matching elements from the right-hand array will be ignored.

refs: http://php.net/manual/en/language.operators.array.php


asgrim:
LGTM :+1:

asgrim:
Thanks @AydinHassan

PR #167, opened 2016-02-06
Failing test for class name resolution in class constants

AydinHassan:
Hi, 

Class name resolution in constants doesn't seem to work, I've provided a failing test case but I have no idea how to fix it :smile: 

If you could give some pointers I can have a pop at fixing it?

### DIFF
--- a/test/unit/NodeCompiler/CompileNodeToValueTest.php
+++ b/test/unit/NodeCompiler/CompileNodeToValueTest.php
@@ -251,4 +251,37 @@ class CompileNodeToValueTest extends \PHPUnit_Framework_TestCase
 
         $this->assertSame('baz', $paramInfo->getDefaultValue());
     }
+
+    public function testClassConstantClassNameResolution()
+    {
+        $phpCode = '<?php
+
+        class Foo {
+        }
+        class Bat {
+            const QUX = Foo::class;
+        }
+        ';
+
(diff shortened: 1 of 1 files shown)


test/unit/NodeCompiler/CompileNodeToValueTest.php
```diff
@@ -251,4 +251,37 @@ public function method($param = static::BAR) {}
...
+
+        class Foo {
+        }
+        class Bat {
+            const QUX = Foo::class;
+        }
+        ';
+
+        $reflector = new ClassReflector(new StringSourceLocator($phpCode));
+        $classInfo = $reflector->reflect('Bat');
```
Ocramius: Shouldn't this be `Bar\Bat`?

Ocramius: @asgrim should this method throw an exception on not found class?

AydinHassan: @Ocramius the other test should have been `Bar\Bat` this one isn't namespaced. Fixed that now.

asgrim:
Thanks @AydinHassan - I've opened a new PR with your test and the implementation in #168 - much appreciated

AydinHassan:
Awesome, thanks @asgrim !

PR #168, opened 2016-02-07
Implemented ::class constant resolution 

asgrim:
Completed implementation for failing test case from PR #167 (thanks @AydinHassan!)


Ocramius:
@asgrim is this to be backported to `1.0.1`?

asgrim:
I don't think so, it's a new feature still, so should aim for 1.1.0

Ocramius:
@asgrim if this isn't supposed to work in `1.0.x`, and is added, then the PR should probably include a change in the changelog/docs/compat docs?

Issue #178, opened 2016-02-20
Look up code unit by filename and line number


sebastianbergmann:
I recently needed functionality to [look up a code unit by filename and line number](https://github.com/sebastianbergmann/code-unit-reverse-lookup/blob/master/src/Wizard.php#L39) and put it into a new component that is now a dependency of php-code-coverage.

I am not happy with the names that I have chosen for this functionality and begin to think that this functionality may be better kept in a component with similar functionality. I was wondering if you already have this in BetterReflection or would consider adding it if not?

asgrim:
Yes; this can be done quite trivially; there's a couple of ways I can think of doing this but I need to have a think through them to make sure it's exactly what you're after (whether we need to change/add functionality, or whether it can be done already with minimal fuss for example)

sebastianbergmann:
Thanks for looking into it, @asgrim. Let me know if there is anything I can help with.

asgrim:
Alright, I've had a look, and we can pretty easily add new methods to our `ClassReflector` and `FunctionReflector` to `reflectOnLine`. So the usage would look something like this:

```
function findClass($file, $line)
{
    return (new ClassReflector(new SingleFileSourceLocator($file)))->reflectOnLine($line);
}
function findFunct($file, $line)
{
    return (new FunctionReflector(new SingleFileSourceLocator($file)))->reflectOnLine($line);
}
```

This would work for interfaces, classes, traits and functions. You could combine those two calls to get "anything" on a line.

If you're happy with this sort of usage, I can finish the implementation and get this merged, we'd target the 1.1.0 release for this.

sebastianbergmann:
It would be great if there was a `findAll($file, $line)` method as it would be inconvenient to first try `findClass()` and then `findFunct()` if the first call did not yield anything.

asgrim:
@sebastianbergmann feedback welcome on PR #179 - from what I understand this fulfils your needs now?

> As per discussion in feature request #178, this introduces a new `CombinedReflector` API that allows reflecting on a specific line.
> 
> Usage would look like this:
> 
> ```php
> $reflector = new CombinedReflector(new SingleFileSourceLocator($file));
> 
> // Will return a ReflectionFunction, ReflectionClass or null, depending on
> // what is on line 7 in $file
> $item = $reflector->reflectOnLine(7);
> ```

asgrim:
[...]

PR #179, opened 2016-02-21
[WIP] Added CombinedReflector, a helper API

asgrim:
As per discussion in feature request #178, this introduces a new `CombinedReflector` API that allows reflecting on a specific line.

Usage would look like this:

```php
$reflector = new CombinedReflector(new SingleFileSourceLocator($file));

// Will return a ReflectionFunction, ReflectionClass or null, depending on
// what is on line 7 in $file
$item = $reflector->reflectOnLine(7);
```

### DIFF
--- a/src/Reflector/CombinedReflector.php
+++ b/src/Reflector/CombinedReflector.php
@@ -1,0 +1,87 @@
+<?php
+
+namespace BetterReflection\Reflector;
+
+use BetterReflection\Identifier\Identifier;
+use BetterReflection\Identifier\IdentifierType;
+use BetterReflection\Reflection\Reflection;
+use BetterReflection\Reflection\ReflectionClass;
+use BetterReflection\Reflection\ReflectionFunction;
+use BetterReflection\SourceLocator\Type\SourceLocator;
+
+class CombinedReflector implements Reflector
+{
+    /**
+     * @var SourceLocator
+     */
+    private $sourceLocator;
+
+    public function __construct(SourceLocator $sourceLocator)
+    {
+        $this->sourceLocator = $sourceLocator;
+    }
+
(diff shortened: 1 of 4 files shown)


src/Reflector/CombinedReflector.php
```diff
@@ -0,0 +1,87 @@
...
+     *
+     * @param int $lineNumber
+     * @return ReflectionFunction|ReflectionClass|null
+     */
+    public function reflectOnLine($lineNumber)
+    {
+        $reflections = $this->getAll();
+        foreach ($reflections as $reflection) {
+            if (method_exists($reflection, 'getStartLine')
+                && $reflection->getStartLine() === $lineNumber) {
```
sebastianbergmann: We need a range check here: `$lineNumber` has to be `>=` `$reflection->getStartLine()` and `<=` `$reflection->getEndLine()`.

Ocramius: @sebastianbergmann can you clarify on why a range check is needed here?

asgrim: @Ocramius Presumably so you can pick any line within the class/function and have it find the class or function

sebastianbergmann: Exactly. Given a filename and a linenumber I want to know which unit of code, if any, this line is part of. That is what https://github.com/sebastianbergmann/code-unit-reverse-lookup/blob/master/src/Wizard.php#L39 currently does and what I like BetterReflection to do.

asgrim: Hang on, this won't pick up `ReflectionMethod`s currently, looks like your `Wizard` does this. /me does more thinking

Issue #162, opened 2016-02-06
Ability to add/remove type declarations and return type declarations


asgrim:
Depends on #164 (PHP 7 types compatibility) being implemented first, but once done, I'd like the ability to:

 * Add/Remove type declarations to parameters
 * Add/Remove return type declarations from methods/functions

Proposed API additions:

* `ReflectionParameter::setType(ReflectionType $newType)`
* `ReflectionParameter::removeType()`
* `ReflectionFunctionAbstract::setReturnType(ReflectionType $newType)`
* `ReflectionFunctionAbstract::removeReturnType()`

asgrim:
Closed by #180

PR #180, opened 2016-02-21
Set/remove PHP 7 types

asgrim:
For feature request in issue #162, this adds feature and documentation for setting/removing PHP 7 type declarations, for both parameters and return types.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ features than PHP's built-in [reflection API](http://php.net/manual/en/book.refl
 * Ability to extract AST from methods and functions
 * Ability to return AST representation of a class or function
 * Fetch return type declaration and parameter type declarations in PHP 7 code (even when running PHP 5!)
-* Change or remove PHP 7 return type declarations from methods and functions
+* Change or remove PHP 7 parameter type and return type declarations from methods and functions
 * Change the body of a function or method to do something different
 * *Moar stuff coming soon!*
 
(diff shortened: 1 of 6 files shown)


Ocramius:
I just realized that we now have (since #141, which I should *NOT* have merged, given this comment) reflection API for *reading* reflection data, and reflection API for writing reflection data... all smashed together.

Yeah, that sounds messy as heck.

@asgrim could you look at how we could separate the read and the write side, or make reflections immutable?

I'd say we should hold on this until we know what the best approach for this and #141 is... And we should really deprecate/move (to a separate unit) the functionality that was introduced in #141

asgrim:
If we want to be compatible with core reflection, they can't be immutable anyway (setValue, setAccessible etc.). The only thing I could suggest is using 'with*` methods, but that would cause copying the entire AST each time which isn't very efficient... Especially with a large reflected class.

Ocramius:
> that would cause copying the entire AST each time which isn't very efficient... Especially with a large reflected class.

True, but I already see a lot of foot-gunning going on when dealing with the mutable approach :-)

asgrim:
I'm not sure of a nicer way of doing this without breaking compatibility with core reflection API :|

asgrim:
@Ocramius we gonna merge this or find an alternate solution, Mr Immutable?

Ocramius:
My complaints are moved to #182.

:+1: for now!

PR #181, opened 2016-02-22
Added FindReflectionOnLine helper utility

asgrim:
Attempt number 2 and solving #178 in a nicer way.

This provides an API as per the documentation written in `docs/find-reflection-on-line.md`.

### DIFF
--- a/docs/find-reflection-on-line.md
+++ b/docs/find-reflection-on-line.md
@@ -1,0 +1,18 @@
+# Find a reflection on specified line
+
+Purely an ease-of-use API, the `FindReflectionOnLine` helper simplifies usage
+when you want to find a reflection on a specific line. Usage is very simple, as
+the class is an invokable class:
+
+```php
+<?php
+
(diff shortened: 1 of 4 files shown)


asgrim:
ping @sebastianbergmann - have you had a chance to look over this change, and whether it's suitable for your needs, please?

sebastianbergmann:
Not yet, sorry. Hopefully tomorrow.

sebastianbergmann:
I had a chance to look at this and it does what I need. Thanks!

Can you explain, though, why you chose to implement this as a functor (using `__invoke()`) instead of a straight-forward public method?

asgrim:
The class has a single purpose, so no need for any other public API really. Can be used throw-away, for example in PHP 7, something like this should work (didn't check syntax btw):

```php
$reflection = (new FindReflectionOnLine)($file, $line);
```

Ocramius:
This is good to go: merging!

PR #184, opened 2016-03-16
Add demo tests to Travis build

asgrim:


test/demo/check-demo.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+
+cd "$(dirname "$0")"
+cd ../..
+
+check () {
```
Ocramius: `set -e` does the same...

test/demo/check-demo.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+
+cd "$(dirname "$0")"
```
Ocramius: Unreliable: use `THIS_SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"`

test/demo/check-demo.sh
asgrim: The script may not actually error, it may just have different output, so this tests the output is exactly as expected

test/demo/check-demo.sh
asgrim: pfft, OSX has to be different :P

Ocramius: Not just osx...

test/demo/check-demo.sh
Ocramius: Ah, alright then :-)

asgrim:
@Ocramius this ready for review when you have some time please :)

Ocramius:
:+1:

PR #185, opened 2016-03-16
Use prefer-dist in travis now (see composer/composer#4884)

asgrim:


.travis.yml
```diff
@@ -11,7 +11,7 @@ env:
     - DEPENDENCIES="--prefer-lowest --prefer-stable"
 
 before_script:
-  - composer update --prefer-source $DEPENDENCIES
+  - composer update --prefer-dist $DEPENDENCIES
```
Ocramius: This will cause travis to choke if under high load: avoid.

asgrim: Nope, GitHub changed their API limits, so this shouldn't be necessary any more, see: composer/composer#4884

Ocramius:
Fair enough then :-)

PR #188, opened 2016-04-02
New phpt suite

### LINKED ISSUE #155: Write .phpt test suite
Ref; #146 the `.phpt` tests are mostly unusable at the moment because they mostly use `var_dump`, and `var_dump` on a Better Reflection is like a black hole. Alter the `.phpt` to be more sensible about stuff (`var_dump` on a BR is mostly useless anyway, you just dump `Zend\Code` and the AST stuff...)

asgrim:
Fixes #155 by kicking off a new .phpt test suite that demonstrates most basic functionalities.

Could not test the "reflect from closure" type stuff, because closures can't be located (due to issue discovered in #12 - the phpunit issue sebastianbergmann/phpunit#1783)


test/compat/GetFunctionBody.phpt
```diff
@@ -0,0 +1,76 @@
...
+$functionInfo = $reflector->reflect('adder');
+var_dump($functionInfo->getName());
+
+var_dump($functionInfo->getBodyAst());
+var_dump($functionInfo->getBodyCode());
+
+?>
+--EXPECTF--
+string(5) "adder"
+array(1) {
```
Ocramius: This test seems too fragile, and may break if PHP-Parser changes internal properties representation. `getBodyCode` is sufficient.

asgrim: Updated

Issue #189, opened 2016-04-02
Call to a member function isInterface() on null


Danack:
Even if this isn't supposed to work, having a better exception than an EngineException would be good.

```
<?php

use BetterReflection\SourceLocator\Type\StringSourceLocator;
use BetterReflection\Reflector\ClassReflector;

$autoloader = require __DIR__.'/../vendor/autoload.php';

$code = <<< 'CODE'
<?php 

namespace Abraham\TwitterOAuth;

class HmacSha1 extends SignatureMethod
{
    /**
     * {@inheritDoc}
     */
    public function getName()
    {
        return "HMAC-SHA1";
    }

    /**
     * {@inheritDoc}
     */
    public function buildSignature(Request $request, Consumer $consumer, Token $token = null)
    {
        $signatureBase = $request->getSignatureBaseString();

        $parts = [$consumer->secret, null !== $token ? $token->secret : ""];

        $parts = Util::urlencodeRfc3986($parts);
        $key = implode('&', $parts);

        return base64_encode(hash_hmac('sha1', $signatureBase, $key, true));
    }
}

CODE;

$sourceLocator = new StringSourceLocator($code);

$reflector = new ClassReflector($sourceLocator);
$reflClass = $reflector->reflect('Abraham\TwitterOAuth\HmacSha1');

$properties = $reflClass->getProperties();
$methods = $reflClass->getMethods();


```

asgrim:
@Danack all your problems are caused by trying to find classes that don't exist within the scope of the SourceLocator (in this case, `SignatureMethod` does not exist within scope).

I'm not sure the best way of handling it; there's a couple of options I guess;

1. throw an exception to say "can't reflect XXX because not found" or whatever (useful feedback)
1. make a "placeholder" reflection or something (probably not so useful, obscures the problem I think)

unless you can think of someting else? cc @Ocramius

Ocramius:
> throw an exception to say "can't reflect XXX because not found" or whatever (useful feedback)

That's what reflection does, indeed.

An alternative would be to create an aggregate source locator that wraps around your string + your existing classes.

asgrim:
Exception would look like:

```
[...]

PR #193, opened 2016-04-05
Throw exceptions when ClassReflector cannot locate identifiers

asgrim:
As reported in #189, we are creating calls on `null` objects when identifiers are not found. This is expected behaviour, but not clearly reported on. This change introduces the `IdentifierNotFound` exception, but at the `ClassReflector` level (thus allowing exception to bubble up, rather than simply returning `null`).


Ocramius:
:+1:

Issue #192, opened 2016-04-03
Need to report which files are unparseable.


Danack:
Scenario - I'm trying to analyze a large library and all the classes in it. One (or more) of the files that are being included in the parsing are not valid PHP files. In this case because I'm accidentally including a test file.

Currently what happens is that a PHPParser exception escapes, and has no useful information about what file contains the problem.

It would be good to catch all PHPParser exceptions inside BetterReflection, and to throw a more specific BetterReflectionException, that contains the information about which file was being parsed or other specific problem, so that it can be fixed, without using the force.

Current output, doesn't say which file caused the problem:
```
PHP Fatal error:  Uncaught PhpParser\Error: Syntax error, unexpected T_DEC, expecting '[' on line 14 in /home/github/John/John/vendor/nikic/php-parser/lib/PhpParser/ParserAbstract.php:277
Stack trace:
#0 /home/github/John/John/vendor/nikic/php-parser/lib/PhpParser/Parser/Multiple.php(58): PhpParser\ParserAbstract->parse('--FILE--\n<?php\n...')
#1 /home/github/John/John/vendor/nikic/php-parser/lib/PhpParser/Parser/Multiple.php(29): PhpParser\Parser\Multiple->tryParse(Object(PhpParser\Parser\Php7), '--FILE--\n<?php\n...')
#2 /home/github/John/John/vendor/roave/better-reflection/src/SourceLocator/Ast/Locator.php(67): PhpParser\Parser\Multiple->parse('--FILE--\n<?php\n...')
#3 /home/github/John/John/vendor/roave/better-reflection/src/SourceLocator/Ast/Locator.php(49): BetterReflection\SourceLocator\Ast\Locator->findReflectionsOfType(Object(BetterReflection\Reflector\ClassReflector), Object(BetterReflection\SourceLocator\Located\LocatedSource), Object(BetterReflection\Identifier\IdentifierType))
#4 /home/github/John/John/vendor/roa in /home/github/John/John/vendor/nikic/php-parser/lib/PhpParser/ParserAbstract.php on line 277
```

Repo code:
```
<?php

use BetterReflection\Reflector\ClassReflector;
use BetterReflection\SourceLocator\Type\SingleFileSourceLocator;
use BetterReflection\SourceLocator\Type\AggregateSourceLocator;


ini_set('memory_limit','768M');

$autoloader = require __DIR__.'/../vendor/autoload.php';

[...]

PR #194, opened 2016-04-05
Rethrow any exception encountered whilst finding reflections

asgrim:
Reported in #192 this new feature rethrows exceptions encountered whilst finding reflections in the `Ast\Locator`, and tries it's best to give useful output based on the `LocatedSource` given.

### DIFF
--- a/src/SourceLocator/Ast/Exception/AstParse.php
+++ b/src/SourceLocator/Ast/Exception/AstParse.php
@@ -1,0 +1,25 @@
+<?php
+
+namespace BetterReflection\SourceLocator\Ast\Exception;
+
+use BetterReflection\SourceLocator\Located\LocatedSource;
+
+class AstParse extends \RuntimeException
+{
+    public static function fromLocatedSource(LocatedSource $locatedSource, \Exception $previous = null)
+    {
+        $additionalInformation = '';
+        if (null !== $locatedSource->getFileName()) {
+            $additionalInformation = sprintf(' (in %s)', $locatedSource->getFileName());
+        }
+
+        if ($additionalInformation === '') {
+            $additionalInformation = sprintf(' (first 20 characters: %s)', substr($locatedSource->getSource(), 0, 20));
+        }
+
+        return new self(sprintf(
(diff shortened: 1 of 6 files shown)


src/SourceLocator/Ast/Exception/AstParse.php
```diff
@@ -0,0 +1,25 @@
+<?php
+
+namespace BetterReflection\SourceLocator\Ast\Exception;
+
+use BetterReflection\SourceLocator\Located\LocatedSource;
+
+class AstParse extends \RuntimeException
```
Ocramius: The exception name does not tell anything on its own: better name required

src/SourceLocator/Ast/Locator.php
```diff
@@ -59,15 +60,20 @@ public function findReflection(Reflector $reflector, LocatedSource $locatedSourc
...
-            $locatedSource
-        );
+        try {
+            return $this->findReflectionsInTree->__invoke(
+                $reflector,
+                $this->parser->parse($locatedSource->getSource()),
+                $identifierType,
+                $locatedSource
+            );
+        } catch (\Exception $exception) {
```
Ocramius: is `\Exception` intended here? Are there no more specific exceptions? What about `\Throwable`, if your intent is to catch everything?

src/SourceLocator/Ast/Exception/AstParse.php
Ocramius: Also: AST is already AST, while parsing happens before AST. Some terminology confusion here, IMO.

Is this about parsing or is it about AST structure being invalid?

asgrim: Agree - updated

src/SourceLocator/Ast/Locator.php
asgrim: Yes, intentional - anything beyond here we just rethrow, so I added the catch for `\Throwable` too

Ocramius:
:+1:

PR #196, opened 2016-04-05
Added addParameter and removeParameter methods

asgrim:
As per #134, this adds new methods `ReflectionFunctionAbstract::addParameter` and `ReflectionFunctionAbstract::removeParameter`

### DIFF
--- a/src/Reflection/ReflectionFunctionAbstract.php
+++ b/src/Reflection/ReflectionFunctionAbstract.php
@@ -585,7 +585,7 @@ abstract class ReflectionFunctionAbstract implements \Reflector
      * Remove a parameter from the method/function.
      *
      * @param string $parameterName
-     * @return bool
+     * @return void
      */
     public function removeParameter($parameterName)
     {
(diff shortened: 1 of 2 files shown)


src/Reflection/ReflectionFunctionAbstract.php
```diff
@@ -569,4 +570,33 @@ public function setBodyFromAst(array $nodes)
...
+    {
+        $lowerName = strtolower($parameterName);
+
+        foreach ($this->node->params as $key => $paramNode) {
+            if (strtolower($paramNode->name) === $lowerName) {
+                unset($this->node->params[$key]);
+            }
+        }
+
+        return false;
```
Ocramius: Why do you always return `false`? If at least one matching node was found, you should return `true`, no?

Ocramius: Also: is a return value useful at all here?

asgrim: Yes, that was left in by mistake, nice catch - I've removed now :)

PR #197, opened 2016-05-07
Added ReflectionProperty::setVisibility method

asgrim:
Final part of #134 adds `ReflectionProperty::setVisibility()` method to change the default visibility (note; this is distinct from `setAccessible` which operates on runtime visibility)

### DIFF
--- a/src/Reflection/ReflectionProperty.php
+++ b/src/Reflection/ReflectionProperty.php
@@ -6,6 +6,7 @@ use BetterReflection\NodeCompiler\CompileNodeToValue;
 use BetterReflection\NodeCompiler\CompilerContext;
 use BetterReflection\Reflector\Reflector;
 use BetterReflection\TypesFinder\FindPropertyType;
+use PhpParser\Node\Stmt\Class_;
 use PhpParser\Node\Stmt\Property as PropertyNode;
 use phpDocumentor\Reflection\Type;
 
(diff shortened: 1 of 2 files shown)


src/Reflection/ReflectionProperty.php
```diff
@@ -118,6 +119,33 @@ private function getVisibilityAsString()
...
 
     /**
+     * Set the default visibility of this propery. Use the core \ReflectionProperty::IS_* values as parameters, e.g.:
+     *
+     * @param int $newVisibility
+     * @throws \InvalidArgumentException
+     */
+    public function setVisibility($newVisibility)
+    {
+        $this->node->type = ~Class_::MODIFIER_PRIVATE;
```
Ocramius: These three assignments are overwriting the type: seems wrong

asgrim: they're unsetting those bits, then we set the new visibility bit instead; bitwise isn't my strong point

Ocramius: Could you point me at where this is happening? This can be simply reduced into a single bitwise operation:

`$this->node->type = ~Class_::MODIFIER_PRIVATE & ~Class_::MODIFIER_PROTECTED & ~Class_::MODIFIER_PUBLIC`

As it currently stands, your last assignment (line 131) is setting the modifier to "PUBLIC", and that's it.

asgrim: Hmm, I understood this to be:
 * remove the "private" bit, if set
 * remove the "protected" bit, if set
 * remove the "public" bit, if set
 * then in the switch statement, set the appropriate visibility bit depending on `$newVisibility`
Have I misunderstood?

Ocramius: That would be a `&=` then, not just a `=`

src/Reflection/ReflectionProperty.php
```diff
@@ -118,6 +119,33 @@ private function getVisibilityAsString()
     }
 
     /**
+     * Set the default visibility of this propery. Use the core \ReflectionProperty::IS_* values as parameters, e.g.:
```
pamil: Set the default visibility of this proper**t**y. :)

src/Reflection/ReflectionProperty.php
asgrim: okay; updated I think; the tests worked, but I guess because there weren't any "other" bits set :)

asgrim: Hmm, no; tests failing using `$this->node->type &= ~Class_::MODIFIER_PRIVATE;` etc.

asgrim: Ah got it; the "setting" the new bit after was wrong, should've been using `|=` :) tests should hopefully pass now

src/Reflection/ReflectionProperty.php
asgrim: Thanks!

asgrim:
Test failure here is irrelevant to this PR, and fixed properly in #199

Ocramius:
👍

PR #199, opened 2016-05-13
Fix unit test broken by PhpParser 2.1.0 change

asgrim:
Referring to https://github.com/nikic/PHP-Parser/releases/tag/v2.1.0 a change was introduced that distinguishes between "- and '-enclosed strings, therefore proved this test was flaky (the conversion from " to ' was assumped implicit).

### DIFF
--- a/test/unit/Reflection/ReflectionFunctionAbstractTest.php
+++ b/test/unit/Reflection/ReflectionFunctionAbstractTest.php
@@ -440,11 +440,11 @@ class ReflectionFunctionAbstractTest extends \PHPUnit_Framework_TestCase
 
     public function testGetBodyCode()
     {
-        $php = '<?php
+        $php = "<?php
             function foo() {
-                echo "Hello world";
+                echo 'Hello world';
             }
-        ';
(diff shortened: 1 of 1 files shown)


Ocramius:
👍

Ocramius:
@asgrim seems like installing with newer deps generates newer code. Test expectations on generated code break therefore (newer php-parser generates code with `'` or `"`, old php-parser just generates one of the two)

Ocramius:
One simple workaround is to compare the parsed code, rather than the generated code.

asgrim:
The point of the function is to generate the code however; we *could* check the AST but the properties of the nodes also differ between parser version... but the point of this test is to test the getBodyCode ;)

Ocramius:
Yeah, but the failure is due to the AST now recording additional info, which also leads to differences in generated code. Comparing the AST rather than the code will actually lead to the same results (for 2.0.x and 2.1.x), while it won't for the generated code.

You'd basically have to feed the dumped code to the parser again, then compare the two AST trees.

PR #206, opened 2016-07-03
Added ReflectionFunctionAbstract::getReturnStatementsAst() method

### LINKED ISSUE #198: Add ReflectionFunctionAbstract->getReturnStatementAst()
Add `ReflectionFunctionAbstract->getReturnStatementAst()` which, naturally, should return the AST nodes for the `return` statement for a function, if it exists.

asgrim:
Fixes #198 

Adds ability to fetch all `return` statement node trees from a reflection of a function/method.

### DIFF
--- a/src/Reflection/ReflectionFunctionAbstract.php
+++ b/src/Reflection/ReflectionFunctionAbstract.php
@@ -603,6 +603,9 @@ abstract class ReflectionFunctionAbstract implements \Reflector
     /**
      * Fetch an array of all return statements found within this function.
      *
+     * Note that return statements within smaller scopes contained (e.g. anonymous classes, closures) are not returned
+     * here as they are not within the immediate scope.
+     *
      * @return Node\Stmt\Return_[]
      */
     public function getReturnStatementsAst()
--- a/src/Reflection/ReflectionFunctionAbstract.php
+++ b/src/Reflection/ReflectionFunctionAbstract.php
@@ -6,6 +6,7 @@ use BetterReflection\Reflector\Reflector;
 use BetterReflection\SourceLocator\Located\LocatedSource;
 use BetterReflection\TypesFinder\FindReturnType;
 use BetterReflection\TypesFinder\FindTypeFromAst;
+use BetterReflection\Util\Visitor\ReturnNodeVisitor;
 use Closure;
 use PhpParser\Node;
 use PhpParser\Node\Stmt\Namespace_ as NamespaceNode;
@@ -13,6 +14,7 @@ use PhpParser\Node\Expr\Yield_ as YieldNode;
 use PhpParser\Node\Expr\Closure as ClosureNode;
 use PhpParser\Node\Param as ParamNode;
 use phpDocumentor\Reflection\Type;
+use PhpParser\NodeTraverser;
 use PhpParser\ParserFactory;
 use PhpParser\PrettyPrinter\Standard as StandardPrettyPrinter;
 use PhpParser\PrettyPrinterAbstract;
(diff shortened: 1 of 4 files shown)


src/Reflection/ReflectionFunctionAbstract.php
```diff
@@ -597,4 +599,21 @@ public function removeParameter($parameterName)
             }
         }
     }
+
+    /**
+     * Fetch an array of all return statements found within this function.
```
Ocramius: What about return statements in closures and anonymous classes defined in this method?

asgrim: Without checking, I imagine they'd be returned too. Do we want to prevent that?

Ocramius: Yes. Here's my reasoning about it:

```php
function foo()
{
    if (theWorldEnds()) {
        return true; // this return statement "belongs" to "foo"
    }

    return function () { // this return statement "belongs" to "foo" too
        endTheWorld(); // this return statement "belongs" to the closure inside "foo"
    }
}
```

Ocramius: Renaming the API to clarify "ownership" could also be useful

asgrim: I've not renamed the API but added a doc note. This no longer will return `return` statements found within closures or anon classes.

src/Util/Visitor/ReturnNodeVisitor.php
```diff
@@ -0,0 +1,51 @@
...
+    public function leaveNode(Node $node)
+    {
+        if ($this->isScopeChangingNode($node)) {
+            $this->scopeDepth--;
+        }
+    }
+
+    private function isScopeChangingNode(Node $node)
+    {
+        return $node instanceof Node\Expr\Closure || $node instanceof Node\Stmt\Class_;
```
nikic: PHP also allows you to nest real (non-closure) functions. A more robust variant would be to check against the FunctionLike interface here.

src/Util/Visitor/ReturnNodeVisitor.php
```diff
@@ -0,0 +1,51 @@
...
+     */
+    private $returnNodes = [];
+
+    private $scopeDepth = 1;
+
+    public function enterNode(Node $node)
+    {
+        if ($this->isScopeChangingNode($node)) {
+            $this->scopeDepth++;
+        }
```
nikic: An alternative to managing the scope depth would be to return `NodeTraverser::DONT_TRAVERSE_CHILDREN` here.

src/Util/Visitor/ReturnNodeVisitor.php
asgrim: Thanks for the advice - updated in b79e30d :)

src/Util/Visitor/ReturnNodeVisitor.php
asgrim: Thanks for the advice - updated in b79e30d :)

src/Reflection/ReflectionFunctionAbstract.php
Ocramius: @asgrim before we merge this: do you know if this code works also against the closure in the example above? Basically, if I reflect the closure above, will I be able to find `        return endTheWorld(); // this return statement "belongs" to the closure inside "foo"` ?

asgrim: No, it shouldn't be returned - see test case https://github.com/Roave/BetterReflection/pull/206/commits/b79e30ddc2d76ba2bb3344ca21b164d0fc083a0e#diff-01e91f664ebf580f72f9ffa4bad97307R24

Ocramius: @asgrim I mean if I'm explicitly reflecting on that one closure

asgrim: Yes, if you reflect on the closure you'll only see the return statement *inside* the closure

Ocramius:
Not sure why I didn't merge this one earlier! 👍

PR #208, opened 2016-07-03
Replaced fabpot/php-cs-fixer with friendsofphp/php-cs-fixer

asgrim:
Because `fabpot/php-cs-fixer` is moved to new package name.


Ocramius:
:+1:

PR #209, opened 2016-07-03
Added test for `void` return type for PHP 7.1

### LINKED ISSUE #203: Review PHP 7.1 feature: void return type
https://wiki.php.net/rfc/void_return_type

Ensure we support the new `void` return type in PHP 7.1.

asgrim:
Fixes #203 - `void` is already supported, so just needed to add test to ensure.

### DIFF
--- a/test/unit/Reflection/ReflectionFunctionAbstractTest.php
+++ b/test/unit/Reflection/ReflectionFunctionAbstractTest.php
@@ -349,6 +349,7 @@ class ReflectionFunctionAbstractTest extends \PHPUnit_Framework_TestCase
             ['returnsString', 'string'],
             ['returnsNull', 'null'],
             ['returnsObject', '\stdClass'],
+            ['returnsVoid', 'void'],
         ];
     }
 
@@ -523,7 +524,7 @@ class ReflectionFunctionAbstractTest extends \PHPUnit_Framework_TestCase
                 new String_('Hello world!')
             ]),
         ]);
-        
+
         $this->assertSame("echo 'Hello world!';", $function->getBodyCode());
     }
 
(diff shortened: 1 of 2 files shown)


test/unit/Reflection/ReflectionFunctionAbstractTest.php
```diff
@@ -349,6 +349,7 @@ public function returnTypeFunctionProvider()
             ['returnsString', 'string'],
             ['returnsNull', 'null'],
             ['returnsObject', '\stdClass'],
+            ['returnsVoid', 'void'],
```
Ocramius: Is this the only endpoint that is affected? Also, I see no new public API?

asgrim: Yes, nothing needs to change. phpDocumentor/TypeResolver already supported `void` so nothing needs to change for us, just a new test

Ocramius: So I can now call `ReflectionFunctionAbstract#returnsVoid()`? Note that the method seems to be defined via magic, so some sort of documentation (even via docblock) would make things easier

asgrim: No, there's no method called that, that's just the test case name :P no magic here, just the `getTypeHint` method

asgrim: `getReturnType` rather

asgrim: Oh I see where your confusion may have arisen; `returnsVoid` is the name of the function to be reflected in the test. The test is testing the `ReflectionFunctionAbstract#getReturnType()` method, and we expect, when cast to `string`, the value to be `void`.

Ocramius: Ah, makes sense! :-)

Ocramius:
👍 

As discussed on IRC, no further action needed here

PR #212, opened 2016-09-16
Updated deprecated phpunit API calls in test suite

### LINKED ISSUE #170: Use expectException instead of setExpectedException in tests
see: https://thephp.cc/news/2016/02/questioning-phpunit-best-practices

For forwards compatibility, we need to modify calls to `setExpectedException` to use `expectException` and `expectExceptionMessage` instead. 

i.e.

instead of 
```php
$this->setExpectedException(
    InvalidArgumentException::class,
    'foo is not a valid identifier type'
);
```

we need to:

```php
$this->expectException(InvalidArgumentException::class);
$this->expectExceptionMessage('foo is not a valid identifier type');
```

asgrim:
Fixes #170 and also replaces uses of `getMock` with `createMock`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -6,13 +6,13 @@
         "php": "^5.6|^7.0",
         "nikic/php-parser": "^2.0",
         "phpdocumentor/reflection-docblock": "^2.0",
-        "phpdocumentor/type-resolver": "^0.1.6",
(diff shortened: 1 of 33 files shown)


Ocramius:
:+1:

PR #216, opened 2016-10-04
Autoload interruption for monkey patching

### LINKED ISSUE #205: Autoload interruption for dynamic monkey patching
A suggestion I had at DPC was to use a similar stream overriding technique to perform monkey-patched reflection loading. Investigate and see if this is possible!

/cc @Ocramius

asgrim:
Fixes #205


asgrim:
@Ocramius could you give this a quick look over please? It's missing tests, but it's a "mostly working" PoC

demo/monkey-patching/index.php
```diff
@@ -4,19 +4,18 @@
 
 use BetterReflection\Reflector\ClassReflector;
 use BetterReflection\SourceLocator\Type\SingleFileSourceLocator;
-use PhpParser\PrettyPrinter\Standard as CodePrinter;
+use BetterReflection\Util\Autoload;
+
+Autoload::initialise();
```
Ocramius: So british

demo/monkey-patching/index.php
```diff
@@ -4,19 +4,18 @@
...
 
 // Create the reflection first (without loading)
 $classInfo = (new ClassReflector(new SingleFileSourceLocator(__DIR__ . '/MyClass.php')))->reflect('MyClass');
+Autoload::addClass($classInfo);
```
Ocramius: IMO, the autoloader shouldn't be static. We can add a way to get a singleton reference to it (it is a singleton by PHP's design, after all), but not a direct static method (Larval "facade-non-facade" approach)

src/Util/Autoload.php
```diff
@@ -0,0 +1,110 @@
...
+use PhpParser\Node\Name;
+use PhpParser\Node\Stmt\Namespace_;
+use PhpParser\PrettyPrinter\Standard as CodePrinter;
+
+class Autoload
+{
+    /**
+     * @var ReflectionClass[]
+     */
+    private static $reflections = [];
```
Ocramius: Make all these instance methods, as per above. Will also serve for easier testing later on :-)

src/Util/Autoload.php
```diff
@@ -0,0 +1,110 @@
...
+
+        self::$reflections[$reflectionClass->getName()] = $reflectionClass;
+    }
+
+    /**
+     * Register our own autoloader
+     * @throws \LogicException
+     * @return void
+     */
+    public static function initialise()
```
Ocramius: It's probably `register`

src/Util/Autoload.php
```diff
@@ -0,0 +1,110 @@
...
+    {
+        if (self::$initialiseCalled) {
```
Ocramius: br00tcn

src/Util/Autoload.php
```diff
@@ -0,0 +1,110 @@
...
+        spl_autoload_register([__CLASS__, 'autoload'], true, true);
+
+        self::$initialiseCalled = true;
+    }
+
+    /**
+     * Reset our autoloader, unregistering it and returning it to a clean state
+     * @return void
+     */
+    public static function reset()
```
Ocramius: This is something we don't really need to provide, IMO. If this class becomes a callable, we can simply use `spl_autoload_unregister()` with the instance

src/Util/Autoload.php
```diff
@@ -0,0 +1,110 @@
...
+        self::$initialiseCalled = false;
+        self::$reflections = [];
+    }
+
+    /**
+     * @param string $classToLoad
+     * @return bool
+     * @throws \RuntimeException
+     *
+     * @todo what about internal classes? evaled classes? likely we can't modify them...
```
Ocramius: No, we can't, since they are not autoloaded.

src/Util/Autoload.php
```diff
@@ -0,0 +1,110 @@
...
+
+        if ($classInfo->inNamespace()) {
+            $nodes[] = new Namespace_(new Name($classInfo->getNamespaceName()));
+        }
+
+        // @todo need to work out if we need to add `use` imports too...
+
+        $nodes[] = $classInfo->getAst();
+
+        eval((new CodePrinter())->prettyPrint($nodes));
```
Ocramius: This should be deferred to a code loader object, which would then simply decide whether to write things to FS (like a cache) or not. The reason why this is relevant is that we may want to sign code generated via BR.

src/Util/Autoload.php
```diff
@@ -0,0 +1,110 @@
...
+
+        if (!class_exists($classToLoad, false)) {
```
Ocramius: It may be an interface or a trait.

test/unit/Util/AutoloadTest.php
```diff
@@ -0,0 +1,87 @@
...
+        Autoload::addClass($reflection);
+
+        $reflection->getMethod('getValue')->setBodyFromClosure(function () {
+            return 'this is the expected value - we changed it';
+        });
+
+        $c = new TestClassForAutoloader();
+        $resultingValue = $c->getValue();
+
+        Autoload::reset();
```
Ocramius: All this staticness must disappear from the tests as well

src/Util/Autoload.php
asgrim: Can't make `private function __invoke`  - has to be public, so it can't be callable unless we want to expose the autoloader (not sure if that's a good idea?)

src/Util/Autoload.php
asgrim: Avoid registering the autoloader more than once... otherwise can get a bit of a mess...

src/Util/Autoload.php
asgrim: Renamed

src/Util/Autoload.php
asgrim: Done

demo/monkey-patching/index.php
asgrim: Done, though I haven't singleton'd it yet

demo/monkey-patching/index.php
asgrim: FTW

src/Util/Autoload.php
asgrim: Quite

src/Util/Autoload.php
asgrim: Done, though I've only implemented the `EvalLoader` for now

test/unit/Util/AutoloadTest.php
asgrim: Fxied

src/Util/Autoload.php
Ocramius: Not a problem, IMO. Could also simply unregister and re-register itself instead.

src/Util/Autoload/ClassLoaderMethod/FileCacheLoader.php
```diff
@@ -0,0 +1,46 @@
...
+        $this->cacheDirectory = $cacheDirectory;
+        $this->classPrinter = $classPrinter;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function __invoke(ReflectionClass $classInfo)
+    {
+        // @todo seems reasonable, any better approaches?
```
asgrim: probably need to rely on Roave/Signature#3 for this when it's ready

asgrim: We'd need to bump min to 7.0 for that anyway. Gonna look into that separately.

src/Util/Autoload/ClassLoaderMethod/Exception/SignatureCheckFailed.php
```diff
@@ -0,0 +1,21 @@
+<?php
+declare(strict_types = 1);
```
malukenho: Remove `declare(strict_types = 1);` as its supports php `5.6`

asgrim: Good point; except to use `Roave\Signature` library, we need 7.0+ anyway, so I still need to bump the min version here (see #221)

asgrim:
Thoughts appreciated at this point @Ocramius - no need to merge yet, just need some code review punishment I think.

src/Util/Autoload/ClassLoaderMethod/FileCacheLoader.php
```diff
@@ -0,0 +1,88 @@
...
+        /** @noinspection PhpIncludeInspection */
+        require_once $filename;
+    }
+
+    public static function defaultFileCacheLoader($cacheDirectory)
+    {
+        return new self(
+            $cacheDirectory,
+            new PhpParserPrinter(),
+            new FileContentSigner(new Base64Encoder()),
```
asgrim: This should really be a `sha` or `md5` or something, but needs implementation in `Roave\Signature` /cc @malukenho

composer.json
```diff
@@ -8,7 +8,8 @@
         "phpdocumentor/reflection-docblock": "^3.1.1",
         "phpdocumentor/type-resolver": "^0.2.1",
         "zendframework/zend-code": "^3.1",
-        "jeremeamia/superclosure": "^2.3"
+        "jeremeamia/superclosure": "^2.3",
+        "roave/signature": "dev-master"
```
Ocramius: Having this here may cause a bi-directional dependency between this packages, or is signature never going to use BR?

asgrim: I don't think Signature should use BR; no reason it should IMO

src/Util/Autoload/ClassLoaderMethod/FileCacheLoader.php
asgrim: Created issue roave/signature#8

asgrim:
@Ocramius @malukenho this is now ready for review / merge process kthxbai. Evil things in here :D

malukenho:
@asgrim seems good to merge for me 👍

PR #218, opened 2016-10-26
Upgrade phpdocumentor reflection docblock

### LINKED ISSUE #169: Investigate upgrading to phpdocumentor/reflection-docblock 3.1
Currently we're pegged on 2.0, but 3.0 is now out.

There's some changes in 3.0 which seem to break unit tests (mostly seems to be `Error: Class 'phpDocumentor\Reflection\DocBlock\Context' not found`, it may've just been renamed, or maybe something fundamental has changed so that we'd need to refactor.

Additionally, phpunit/phpunit 5 requires phpspec/prophecy which requires 2.0, which conflicts, so we'd have to alias it (`3.0.0 as 2.0.0` or something silly), or get them to allow `^2.0|^3.0`.

asgrim:
Fixes #169

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -5,7 +5,7 @@
     "require": {
         "php": "^5.6|^7.0",
         "nikic/php-parser": "^2.0",
-        "phpdocumentor/reflection-docblock": "^2.0",
+        "phpdocumentor/reflection-docblock": "^3.0",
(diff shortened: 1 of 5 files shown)


Ocramius:
👍

Issue #22, opened 2015-07-02
isOptional determination is a mess


asgrim:
It _does_ work but there is a mess of code in [ReflectionFunctionAbstract::setNodeOptionalFlag](https://github.com/Roave/BetterReflection/blob/11800e85f8c37c537e514f06cb929e49202d397c/src/Reflection/ReflectionFunctionAbstract.php#L67)

``` php
/**
 * We must determine if params are optional or not ahead of time, but we
 * must do it in reverse...
 */
private function setNodeOptionalFlag()
{
    $overallOptionalFlag = true;
    $lastParamIndex = (count($this->node->params) - 1);
    for ($i = $lastParamIndex; $i >= 0; $i--) {
        $hasDefault = ($this->node->params[$i]->default !== null);
        // When we find the first parameter that does not have a default,
        // flip the flag as all params for this are no longer optional
        // EVEN if they have a default value
        if (!$hasDefault) {
            $overallOptionalFlag = false;
        }
        $this->node->params[$i]->isOptional = $overallOptionalFlag;
    }
}
```

A simple solution would be to say something like (pseudo code) `if parameterindex > numrequiredparams, this param is optional` - however, on checking `getNumberOfRequiredParameters()` this has a cyclic dependency on `isOptional()` calls... which means they wouldn't be resolved at creation-time. Solve this somehow.

asgrim:
Incidentally, I think #34 may provide some kind of a workaround for this?

Ocramius:
This does not need fixing for now: closing as `wontfix`

PR #220, opened 2016-11-20
Fixed issues found by static analysis

ondrejmirtes:
These are fixes for issues found by my own static analysis tool [phpstan/phpstan](https://github.com/phpstan/phpstan). It's quite early in its development but it already proves to be really useful in my own projects. I can help you integrate it in BetterReflection if you'd like to.

By the way, there's one more error reported I wasn't sure how to fix:

```
 ------ ------------------------------------------------------------------------- 
  Line   src/Reflection/ReflectionParameter.php                                   
 ------ ------------------------------------------------------------------------- 
  259    Access to an undefined property PhpParser\Node\Param::$isOptional.       
 ------ ------------------------------------------------------------------------- 
```

It can be just a problem in a phpDoc or it can possibly be a real bug.

### DIFF
--- a/src/Reflection/ReflectionType.php
+++ b/src/Reflection/ReflectionType.php
@@ -86,7 +86,7 @@ class ReflectionType
                 return 'bool';
             case $this->type instanceof Types\Float_:
                 return 'float';
-            case $this->type instanceof Types\Void:
+            case $this->type instanceof Types\Void_:
                 return 'void';
             default:
                 return (string)$this->type;
--- a/src/SourceLocator/Type/EvaledCodeSourceLocator.php
+++ b/src/SourceLocator/Type/EvaledCodeSourceLocator.php
@@ -32,8 +32,7 @@ final class EvaledCodeSourceLocator extends AbstractSourceLocator
         $stubber = $this->stubber;
 
         return new EvaledLocatedSource(
-            "<?php\n\n" . $stubber(new ClassReflection($name)),
-            null
+            "<?php\n\n" . $stubber(new ClassReflection($name))
         );
     }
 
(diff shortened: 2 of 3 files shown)


src/Reflection/ReflectionType.php
```diff
@@ -86,7 +86,7 @@ public function __toString()
                 return 'bool';
             case $this->type instanceof Types\Float_:
                 return 'float';
-            case $this->type instanceof Types\Void:
+            case $this->type instanceof Types\Void_:
```
Ocramius: This kind of change probably needs a test. When would this crash, and does the coverage reach it at the moment?

ondrejmirtes: There's probably a test missing for return type void. The Types\Void class does not exist in phpDocumentor anymore.

asgrim: There is a test see [ReflectionTypeTest::testImplicitCastToString](https://github.com/Roave/BetterReflection/blob/master/test/unit/Reflection/ReflectionTypeTest.php#L45-L59), but types in phpDocumentor handles this failure case. The `default:` case below attempts to cast it to string, and types in phpDocumentor all implement `__toString()` so the test is testing the wrong thing. Therefore, it's pretty difficult to make this test fail unless we make a type that can't be cast to a string, so not sure what to suggest here :s

ondrejmirtes: OK, maybe it would be better to remove the `case`? And maybe others that are handled by the `default` branch?

asgrim:
Regarding the missing `isOptional` issue, we actually have an open bug report for that (see #22) where we set a runtime property to avoid a chicken/egg situation. It's pretty nasty indeed, but we'll fix eventually.

asgrim:
Thanks!

Issue #222, opened 2016-11-25
Retrieve in-scope variables for method?


dantleech:
Would it be reasonable to retrieve the local variables which are available in a reflection method?

e.g.

```
class Foobar
{
    public function barfoo(NumberOne $one, NumberTwo $two)
    {
         $numberThree = 3;
    }
}
```

```
$vars = $reflection->getMethod('barfoo')->getLocalVariables();

// [
//     'one' => Variable(type=NumberOne, declaredAt=2)
//     'two  => Variable(type=NumberTwo, declaredAt=2)
//     'numberThree' => Variable(type=int, declaredAt=4
// ];
```

asgrim:
I love this idea! We can definitely consider this one :)

Ocramius:
Oh hell yeah, this would be awesome!

Could detect things like:

 * local variable definition
 * accessed instance properties
 * accessed static properties
 * accessed globals (`global` keyword)
 * accessed super-globals (unsure about this one)

dantleech:
My motivation here is writing an auto-complete engine, as such the main problem is inferring the types of the local variables - we would need to resolve types from ASSIGN nodes the best we can (e.g. `gettype($bar = $boo->baz->bad()) == 'string'`).

The only technical problem, I think, is how we determine the types of class properties so that we can resolve the expression chain:

1. Use `@var` annotations to infer member property types.
2. Fallback to guessing from the `$this` assignations in the `__construct`.

> accessed instance properties
> accessed static properties
> accessed globals (global keyword)
> accessed super-globals (unsure about this one - probably sufficient to do a diff between locally defined variables and accessed variables)

Makes sense - I guess these could be collected in the same visitor/thing.

PR #224, opened 2016-11-27
[WIP] Variable collector

dantleech:
This is an initial attempt at a variable collector (#222), which can be used by the `ReflectionMethod` to determine the variables that are used. The bulk of this code is concerned with resolving types so that:

```php
/**
 * @var Foobar
 */
public function something()
{
     $bar = $this->foobar->getSomeClass()->getFoo();
}
```

If the `getSomeClass` method has the return type `SomeClass` class and `SomeClass` a method `getFoo` that returns type `int`, then `$bar` will be resolved to `int`.

```php
$reflection = $reflector->reflect(SomeClass::class);
$vars = $reflection->getMethod('something')->getVariables();

echo $vars[0]->getName(); // bar
echo $vars[0]->getType(); // int
echo $vars[0]->getDeclaredAt(); // 6
```

The variable collector visitor can return locally assigned and parameter variables available in the AST it is traversing (regardless of scope).  In the case of getting the variables in a method it would suffice to traverse only the method AST.

This is really rough WIP atm and contains an unknown number of bugs, so before reviewing do you agree with the general approach?

TODO:

- [ ] Types from static methods/properties.
- [ ] ...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: php
 
 php:
-  - 5.6
+    #  - 5.6
   - 7.0
   - hhvm
 
--- a/src/Reflection/ReflectionVariable.php
+++ b/src/Reflection/ReflectionVariable.php
@@ -9,6 +9,7 @@ use PhpParser\Node\Param;
 use PhpParser\Node\Expr\Variable;
 use BetterReflection\Reflection\ReflectionType;
 use BetterReflection\Reflection\ReflectionVariable;
+use BetterReflection\Reflection\ReflectionFunctionAbstract;
 
 class ReflectionVariable
 {
@@ -32,14 +33,27 @@ class ReflectionVariable
      */
     private $type;
 
-    public static function createFromParamAndType(Param $param, ReflectionType $type): ReflectionVariable
+    /**
+     * @var ReflectionFunctionAbstract
+     */
+    private $scopeReflection;
+
+    public static function createFromParamAndType(
+        Param $param,
+        ReflectionType $type,
+        ReflectionFunctionAbstract $scopeReflection = null
+    ): ReflectionVariable
     {
         return self::createFromNodeAndType($param, $type);
     }
 
-    public static function createFromVariableAndType(Variable $variable, ReflectionType $type): ReflectionVariable
+    public static function createFromVariableAndType(
+        Variable $variable,
+        ReflectionType $type,
+        ReflectionFunctionAbstract $scopeReflection = null
+    ): ReflectionVariable
     {
-        return self::createFromNodeAndType($variable, $type);
+        return self::createFromNodeAndType($variable, $type, $scopeReflection);
     }
 
     public function getName(): string
@@ -71,6 +85,12 @@ class ReflectionVariable
         return $this->endPos;
     }
 
+    public function getScopeReflection(): ReflectionFunctionAbstract
+    {
+        return $this->scopeReflection;
+    }
+    
+
     /**
      * Create a new reflection variable, 
      *
@@ -81,13 +101,18 @@ class ReflectionVariable
      *       the node, which means that the Lexer should be configured to provide
      *       them.
      */
-    private static function createFromNodeAndType(NodeAbstract $node, ReflectionType $type): ReflectionVariable
+    private static function createFromNodeAndType(
+        NodeAbstract $node,
+        ReflectionType $type,
+        ReflectionFunctionAbstract $scopeReflection = null
+    ): ReflectionVariable
     {
         $reflectionVariable = new self();
         $reflectionVariable->name = $node->name;
         $reflectionVariable->type = $type;
         $reflectionVariable->startPos = $node->getAttribute('startFilePos');
         $reflectionVariable->endPos = $node->getAttribute('endFilePos');
+        $reflectionVariable->scopeReflection = $scopeReflection;
 
         return $reflectionVariable;
     }
--- a/src/Reflection/ReflectionVariable.php
+++ b/src/Reflection/ReflectionVariable.php
@@ -4,6 +4,9 @@ namespace BetterReflection\Reflection;
 
 use phpDocumentor\Reflection\Type;
 use phpDocumentor\Reflection\Types;
+use PhpParser\NodeAbstract;
+use PhpParser\Node\Param;
+use PhpParser\Node\Expr\Variable;
 
 class ReflectionVariable
 {
@@ -15,26 +18,37 @@ class ReflectionVariable
     /**
      * @var int
      */
-    private $declaredAt;
+    private $startPos;
+
+    /**
+     * @var int
+     */
+    private $endPos;
 
     /**
      * @var $type
      */
     private $type;
 
-    /**
-     * @param string $name
-     * @param int $declaredAt
-     * @param string $type
-     * @return ReflectionType
-     */
-    public static function createFromName($name, ReflectionType $type = null, $declaredAt)
+    public static function createFromParamAndType(Param $param, ReflectionType $type)
     {
-        $reflectionType = new self();
-        $reflectionType->name = $name;
-        $reflectionType->declaredAt = $declaredAt;
-        $reflectionType->type = $type;
-        return $reflectionType;
+        return self::createFromNodeAndType($param, $type);
+    }
+
+    public static function createFromVariableAndType(Variable $variable, ReflectionType $type)
+    {
+        return self::createFromNodeAndType($variable, $type);
+    }
+
+    private static function createFromNodeAndType(NodeAbstract $node, ReflectionType $type)
+    {
+        $reflectionVariable = new self();
+        $reflectionVariable->name = $node->name;
+        $reflectionVariable->type = $type;
+        $reflectionVariable->startPos = $node->getAttribute('startPos');
+        $reflectionVariable->endPos = $node->getAttribute('endPos');
+
+        return $reflectionVariable;
     }
 
     public function getName() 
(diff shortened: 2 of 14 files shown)


src/Reflection/ReflectionVariable.php
```diff
@@ -0,0 +1,81 @@
...
+    private $declaredAt;
+
+    /**
+     * @var $type
+     */
+    private $type;
+
+    /**
+     * @param string $name
+     * @param int $declaredAt
```
Ocramius: Unsure about just having the line number.

src/Reflection/ReflectionVariable.php
```diff
@@ -0,0 +1,81 @@
...
+     * @param string $type
```
Ocramius: Doesn't say much - need some sort of more restrictive hint here

src/Reflection/ReflectionVariable.php
```diff
@@ -0,0 +1,81 @@
...
+        return $this->name;
+    }
+
+    public function getDeclaredAt() 
+    {
+        return $this->declaredAt;
+    }
+
+    /**
+     * Get a PhpDocumentor type object for this type
```
Ocramius: What if the type mutated during execution? Two variables with same name, but different type and declaration line, then?

src/Reflection/ReflectionVariable.php
```diff
@@ -0,0 +1,81 @@
...
+        return (!$this->type instanceof Types\Object_);
+    }
+
+    /**
+     * Convert this string type to a string
+     *
+     * @see https://github.com/php/php-src/blob/master/ext/reflection/php_reflection.c#L2993
+     * @return string
+     */
+    public function __toString()
```
Ocramius: Wouldn't add this

src/Util/Visitor/VariableCollectionVisitor.php
```diff
@@ -0,0 +1,273 @@
...
+                $types = $reflParam->getDocBlockTypes();
+                $type = count($types) ? ReflectionType::createFromType(reset($types), false) : null;
+            }
+
+            $this->methodParamTypes[$param->name] = $type;
+            $this->variables[] = ReflectionVariable::createFromName($param->name, $type, $param->getAttribute('startLine'));
+        }
+    }
+
+    private function processAssignation(Expr\Assign $node)
```
Ocramius: Just `processAssign`

Ocramius:
Approach looking sound to me. Loads of testing and refactoring to be done though

src/Reflection/ReflectionVariable.php
dantleech: Well, ideally there would be the range of (characters?) that the variable is available for, but the PHP parser just gives us line numbers ..

src/Reflection/ReflectionVariable.php
dantleech: We return all declarations, so you could get two with the same name, but covering different parts of the code.

src/Reflection/ReflectionVariable.php
dantleech: I found it useful for testing, but mainly I added it to be consistent with the other `Reflection*` classes.

src/Reflection/ReflectionVariable.php
dantleech: Its actually a `ReflectionType` ...

src/Reflection/ReflectionVariable.php
Ocramius: Unless it actually is required API, I wouldn't add it. `__toString()` is often abused, so let's not give users more rope to hang themselves :)

.travis.yml
```diff
@@ -1,7 +1,7 @@
 language: php
 
 php:
-  - 5.6
+    #  - 5.6
```
dantleech: Have switched to using PHP7 at least for development - guessing that you are going to switch to PHP7 in the near-future anyway?

src/Util/Visitor/VariableCollectionVisitor.php
```diff
@@ -0,0 +1,287 @@
...
+    }
+
+    private function createReflectionTypeFromDocType(Type $type): ReflectionType
+    {
+        return ReflectionType::createFromType($type, false);
+    }
+
+    private function createUnknownReflectionType()
+    {
+        return $this->createReflectionTypeFromString('mixed', false);
```
dantleech: What should we do when the type cannot be determined? Currently returning `mixed` - should we return the type as `null`?

Ocramius: Should be `null`, IMO. Adding `mixed` is a very specific thing, plus nothing prevents a `class mixed {}` to exist.

src/Util/Visitor/VariableCollectionVisitor.php
```diff
@@ -0,0 +1,287 @@
...
+        // if this is a function call, try and instantiate the runtime native
+        // \ReflectionFunction, if it is not internal then ignore it as we
+        // cannot guarantee that we are running in the same process as the code
+        // we are analyzing.
+        //
+        // TODO: This is no positive test case for this ... which PHP internal functions
+        //       actually have a return type??
+        if ($expr instanceof Expr\FuncCall) {
+            $func = (string) $expr->name;
+            $reflection = new \ReflectionFunction($func);
```
Ocramius: You can stop at the return type declared on the function btw - no need to introspect.

dantleech:
At the moment this PR scope is about collecting "available" variables (#222), currently including the parameters given in a method call. If taken further this means we should also return `$this` as a variable, and further still, globals, which we cannot do.

For my specifc use case, I need the "available" variables (albeit without the globals). 

Do you think this functionality be limited to variables *assigned* within a scope? Or should we return the func/method param variables also?

MarkRedeman:
@dantleech I think there are still a couple of variable definitions that you're missing in this PR.

- Variable definition through a `foreach` statement. (felixfbecker/php-language-server/#200)
- Pass by reference (felixfbecker/php-language-server#205)
- Assignment trough lists (felixfbecker/php-language-server#170)
- Array destructuring (PHP 7.1+, https://wiki.php.net/rfc/short_list_syntax)

dantleech:
> I think there are still a couple of variable definitions that you're missing in this PR.

indeed, it is far from being complete.

> Variable definition through a foreach statement. 

How? PHPStorm would rely on an annotation above the foreach loop here I think (?)

MarkRedeman:
> indeed, it is far from being complete.

I understand, I mostly wanted to make you aware of those use cases :).

> How? PHPStorm would rely on an annotation above the foreach loop here I think (?)

That's probably going to be a tricky one.. You might be able to check whether the type of the iterable. So that if you're iterating over an `Iterator` then you can check the return type of `current`.
If the iterable is an `IteratorAggregate` then this would require some recursive checks.
For arrays I guess the best use case would be to guess their types from the docblocks.

Btw the use cases I mentioned were mostly about finding the definitions of a variable, not necessarily including their types (though it would be very useful).

asgrim:
@dantleech note: the core namespace is now `\Roave\BetterReflection` so this PR will need updating. No need to provide BC shim class entries for this PR as it's a new feature anyway :)

TomasVotruba:
@dantleech Hey, thanks for this awesome PR. I need exactly this feature, just without reflection and based on `nikic\php-parser`, so I might take a lot from it.

Do you expect to finish this now on BR 2.0? I could help if you need

dantleech:
@TomasVotruba I don't anticipate working on this, I already started [another project](https://github.com/phpactor/worse-reflection) which does this. The problem is that it is a huge amount of work with lots of edge cases. My implementation works "OK" for the purpose (auto completion).

TomasVotruba:
@dantleech Oh, I found out your package [WorseReflection](https://github.com/phpactor/worse-reflection) where you've probably implemented this.

Where could I find there this feature? I looked to [ReflectionMethod](https://github.com/phpactor/worse-reflection/blob/master/lib/Bridge/TolerantParser/Reflection/ReflectionMethod.php) but could not find it.

dantleech:
The entry point is `Reflector::reflectOffset(): ReflectionOffset` (`$offset->frame()` returns the frame).

Most of the code is [here](https://github.com/phpactor/worse-reflection/blob/master/lib/Core/Inference/SymbolInformationResolver.php) and close by.

TomasVotruba:
@dantleech Not sure why, but I didn't see your comments on mobile, that explains my comments :).
Neither I got notified about your last comment. Thanks a lot for that, I'll look on it.

How are you satisfied with the `"microsoft/tolerant-php-parser"`?

Any useful code parts that you think might be useful for me?

dantleech:
In general I am happy with tolerant parser. It's designed for IDE usage, the API is richer than `nikic/php-parser`. For example `$node->getFirstAncestor(FunctionLike::class, ClassLike::class)`, `$node->getDescendantAtOffset(1234)`. It also will resolve the fully qualified names of things, so `$node->getFirstAncestor(ClassLike::class)->getFullyQualifiedName()` etc.

TomasVotruba:
@dantleech Thanks, I'll look on these features. 
I managed to implement most of them via custom `NodeVisitors` already.

Just off-topic: It looks like you have found another solution, so maybe this PR can be closed. I thouth it's WIP and expected it might be finished, but now I see there is complete another way you took. What do you think?

dantleech:
The approach is basically the same - the only reason I created WorseReflection was because it's the core domain of Phpactor and I had lots of requirements. I think the WR implementation could be used as a reference for implementing this feature in the future (I also think the PHP Language Server would be worth a look in this respect).

TomasVotruba:
In case anyone would be looking for `getFirstAncestor` in nikic/php-parser, here it is: https://github.com/RectorPHP/Rector/blob/d59ffe27e33c9a63df5b9e9cf2f0a432e4f4c401/packages/NodeTraverserQueue/src/BetterNodeFinder.php

PR #217, opened 2016-10-04
Update our code for API changes in PHP-Parser

asgrim:
Two things blocking the merge currently:
- [x] Need [`nikic/php-parser`](https://packagist.org/packages/nikic/php-parser) to be released (requires `3.0.0`)
- [x] Need [`jeremeamia/superclosure`](https://packagist.org/packages/jeremeamia/SuperClosure) to do a release (I expect it will be `2.3.0`) - `dev-master` has support for php-parser `3.x` now
- [x] Fix blocking issues/failing tests etc.

PR #226, opened 2016-11-30
Require minimum of PHP 7, enabling tests on PHP 7.1

### LINKED ISSUE #221: Upgrade minimum version to PHP 7.0
Inevitably going to happen. Only question: do we release the next milestone (`1.3.0` and it's features) before bumping the min, or just go ahead with it?

asgrim:
Fixes #221 

Rebase when #217 is merged

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -3,7 +3,7 @@
     "description": "Better Reflection - an improved code reflection API",
     "license": "MIT",
     "require": {
-        "php": "^5.6|^7.0",
+        "php": "^7.0|^7.1",
         "nikic/php-parser": "^3.0",
         "phpdocumentor/reflection-docblock": "^3.0",
         "phpdocumentor/type-resolver": "^0.2",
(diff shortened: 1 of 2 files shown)


asgrim:
Oh yay there's some fun going on here.

Two issues I've spotted:

`?` not supported, needs PHP Parser 3 most likely needs #217 to be merged

```
1) BetterReflectionTest\Reflection\ReflectionParameterTest::testGetDefaultValueAsString with data set #1 ('12.3', '12.300000000000001')
--- Expected
+++ Actual
@@ @@
-12.300000000000001
+12.3
```

composer.json
```diff
@@ -3,7 +3,7 @@
     "description": "Better Reflection - an improved code reflection API",
     "license": "MIT",
     "require": {
-        "php": "^7.0|^7.1",
+        "php": "^7.0",
```
asgrim: Hrmm, I know it's not technically needed, but I think it helps; especially when `php` itself isn't actually semver. In fact, perhaps better to be `7.0.*|7.1.*` heh?

Ocramius: @asgrim indeed, better choice. Updating now.

PR #227, opened 2016-11-30
Prefix namespaces with Roave

### LINKED ISSUE #223: Add vendor prefix to namespaces
Namespace IMO should be `Roave\BetterReflection` not just `BetterReflection`.

We can do this without breaking BC; just add a `autoload` file into `composer.json` that contains `class_alias`es for the old namespace.

From 2.0.0 we can drop that and the old namespace can disappear.

asgrim:
Fixes #223


namespace-bc-aliases.php
```diff
@@ -0,0 +1,75 @@
+<?php
+
+class_alias('Roave\BetterReflection\Identifier\Identifier', 'BetterReflection\Identifier\Identifier');
```
Ocramius: `::class` where applicable, please

test/unit/Reflection/ReflectionPropertyTest.php
```diff
@@ -238,7 +238,7 @@ public function testGetDefaultProperty()
 
     public function testCannotClone()
     {
-        $classInfo = $this->reflector->reflect('\BetterReflectionTest\Fixture\ExampleClass');
+        $classInfo = $this->reflector->reflect('\Roave\BetterReflectionTest\Fixture\ExampleClass');
```
Ocramius: We should use `::class` across the tests too. At least for the changes in this diff

asgrim:
@Ocramius FIXED ALL THE THINGS using `::class` \m/

PR #232, opened 2016-12-08
[RFC] Allow the Parser to be injected into the AST Locator

dantleech:
Currently it is not possible to customize the PHP parser instance in the `Ast\Locator`. This means it is not possible, f.e. to:

1. Disable throwing errors.
2. Add new attributes.

I propose adding the `Locator` as an optional argument to all source locators.

### DIFF
--- a/src/SourceLocator/Type/ComposerSourceLocator.php
+++ b/src/SourceLocator/Type/ComposerSourceLocator.php
@@ -21,9 +21,9 @@ class ComposerSourceLocator extends AbstractSourceLocator
      */
     private $classLoader;
 
-    public function __construct(ClassLoader $classLoader)
+    public function __construct(ClassLoader $classLoader, Locator $locator = null)
     {
-        parent::__construct();
+        parent::__construct($locator);
         $this->classLoader = $classLoader;
     }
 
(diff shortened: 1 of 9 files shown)


src/SourceLocator/Type/ComposerSourceLocator.php
```diff
@@ -21,9 +21,9 @@ class ComposerSourceLocator extends AbstractSourceLocator
      */
     private $classLoader;
 
-    public function __construct(ClassLoader $classLoader)
+    public function __construct(ClassLoader $classLoader, Locator $locator = null)
```
dantleech: This would be done for all source locators.

asgrim:
I don't see any harm in this really, not really useful for internal stuff, but if it makes usage easier/different then :+1:

Ocramius:
PR is good, but lacks testing. Doesn't need much coverage, but expecting calls on the injected locator would be sufficient.

test/unit/SourceLocator/Type/EvaledCodeSourceLocatorTest.php
```diff
@@ -9,18 +9,55 @@
...
+     * @var ReflectionClass
+     */
+    private $reflectionClass;
+
+    /**
+     * @var Locator
+     */
+    private $locator;
+
+    public function setUp()
```
dantleech: Have refactored this test slightly - I inject the `Locator` mock into each new instance rather than creating a new test which duplicates another test in order to see if an arbitrary method was called on the newly injected `Locator`, which seems fragile.

As we have more mocks I switched to a `setUp` style for creating the mock objects.

wdyt?

asgrim: Yep, LGTM. Inject all the things! :+1: :)

Ocramius:
Implemented after a lot of rework via https://github.com/Roave/BetterReflection/pull/338

PR #236, opened 2016-12-24
Fix exception being thrown in AutoloadSourceLocator

### LINKED ISSUE #234: AutoloadSourceLocator#locateFunctionByName should return `null`
```php
private function locateFunctionByName($functionName)
{
    if (!function_exists($functionName)) {
        throw new FunctionUndefined('Function ' . $functionName . ' was not already defined');
    }

    $reflection = new \ReflectionFunction($functionName);
    return $reflection->getFileName();
}
```

This should not throw an exception, it should `return null;` instead. A source locator should not throw an exception unless something is drastically wrong (e.g. misconfiguration of it); a SL should either return the located item, or return `null` (success or fail, not error state).

It's important because the `AutoloadSourceLocator` can't autoload a function unless it's already in memory, but if you have a later source locator that *could* find that same identifier (in an aggregate), then you're breaking the chain.

asgrim:
The handling in FunctionReflector was actually wrong - it was returning `null` when it should be throwing an `IdentifierNotFound` exception. In the case of using AutoloadSourceLocator when reflecting on a function, an exception `FunctionUndefined` was being thrown. This has been removed (therefore this is sadly a minor BC break) and we handle this correctly.

Fixes #234

### DIFF
--- a/src/SourceLocator/Exception/FunctionUndefined.php
+++ b/src/SourceLocator/Exception/FunctionUndefined.php
@@ -1,7 +1,0 @@
-<?php
-
-namespace Roave\BetterReflection\SourceLocator\Exception;
-
-class FunctionUndefined extends \RuntimeException
-{
-}
--- a/src/SourceLocator/Exception/FunctionUndefined.php
+++ b/src/SourceLocator/Exception/FunctionUndefined.php
@@ -1,0 +1,16 @@
+<?php
+
+namespace Roave\BetterReflection\SourceLocator\Exception;
+
+/**
+ * This is removed in PR #236 - this should never have existed, but leaving here for BC. This exception is never thrown
+ * so you should remove it from your code.
+ *
+ * You were probably looking to catch \Roave\BetterReflection\Reflector\Exception\IdentifierNotFound instead.
+ *
+ * @deprecated You're probably looking for `IdentifierNotFound`
+ * @see \Roave\BetterReflection\Reflector\Exception\IdentifierNotFound
+ */
+class FunctionUndefined extends \RuntimeException
+{
+}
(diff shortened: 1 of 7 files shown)


src/SourceLocator/Exception/FunctionUndefined.php
```diff
@@ -1,7 +0,0 @@
-<?php
-
-namespace Roave\BetterReflection\SourceLocator\Exception;
-
-class FunctionUndefined extends \RuntimeException
```
Ocramius: Deprecate, don't drop

asgrim:
@Ocramius fixed; ready to review whenever build decides to complete :)

src/SourceLocator/Type/AutoloadSourceLocator.php
```diff
@@ -95,13 +94,12 @@ class_exists($className);
...
      * @param string $functionName
-     * @return string
-     * @throws FunctionUndefined
+     * @return string|null
      */
     private function locateFunctionByName($functionName)
     {
         if (!function_exists($functionName)) {
-            throw new FunctionUndefined('Function ' . $functionName . ' was not already defined');
+            return null;
```
Ocramius: Also after this block: https://3v4l.org/vJkTR

```php
(new ReflectionFunction('preg_match'))->getFileName()
```

This produces `false`, not `null`. Can you factor it in?

asgrim: @Ocramius this isn't relevant yet, and also won't work (we can't yet reflect internal functions - see #38) so it doesn't affect us yet, and is also irrelevant for this PR

Ocramius: Moved to https://github.com/Roave/BetterReflection/issues/237

Ocramius:
👍

PR #238, opened 2016-12-24
LocatedSource now allows empty strings as source code

### LINKED ISSUE #235: LocatedSource#__construct() should accept empty string as $source
```php
public function __construct($source, $filename)
    {
        if (!is_string($source) || empty($source)) {
            throw new \InvalidArgumentException(
                'Source code must be a non-empty string'
            );
        }
```

There's no real reason we have to throw an exception here; allow empty strings, yo!

asgrim:
Fixes #235


Ocramius:
👍

PR #239, opened 2016-12-26
[POC/RFC] Performance issues

dantleech:
There are some performance issues which become very pronounced when:

- Reflecting large classes and their methods / properties (e.g. `__toString`), for example `PHPUnit_Framework_TestCase`.
- Calling `ReflectionClass#isSubClass` [many times](https://github.com/dantleech/phpactor/blob/master/lib/Complete/Provider/FetchProvider.php#L253).

The first performance issue can be greatly improved by caching in
`FindTypeFromAst` class, the second by caching calls to
`ReflectionClass#getInheritanceClassHierarchy`.

In this PR I included fixes to illustrate the issue, but ultimately it might be
better to take a different strategy (e.g. creating a service to resolve types
from AST).

In this PR I added a PHPBench benchmark which uses a copy of the `PHPUnit_Framework_TestCase` as a subject.

Before:

```
\Roave\BetterReflectionBenchmark\PhpunitFileBench

    reflect_phpunit_class         I0 P0         [μ Mo]/r: 0.547 0.547 (s)       [μSD μRSD]/r: 0.000s 0.00%
    reflection_to_string          I0 P0         [μ Mo]/r: 92.007 92.007 (s)     [μSD μRSD]/r: 0.000s 0.00%
```

After:

```
\Roave\BetterReflectionBenchmark\PhpunitFileBench

    reflect_phpunit_class         I0 P0         [μ Mo]/r: 0.479 0.479 (s)       [μSD μRSD]/r: 0.000s 0.00%
    reflection_to_string          I0 P0         [μ Mo]/r: 0.967 0.967 (s)       [μSD μRSD]/r: 0.000s 0.00%A
```

### DIFF
--- a/src/Reflection/ReflectionClass.php
+++ b/src/Reflection/ReflectionClass.php
@@ -61,6 +61,8 @@ class ReflectionClass implements Reflection, \Reflector
      */
     private $cachedMethods;
 
+    private $hierarchy;
+
     private function __construct()
     {
     }
@@ -1014,10 +1016,22 @@ class ReflectionClass implements Reflection, \Reflector
             : [];
     }
 
+
+    private function getInheritanceClassHierarchy()
+    {
+        if ($this->hierarchy) {
+            return $this->hierarchy;
+        }
+
+        $this->hierarchy = $this->doGetInheritanceClassHierarchy();
+
+        return $this->hierarchy;
+    }
+
     /**
      * @return ReflectionClass[] ordered from inheritance root to leaf (this class)
      */
-    private function getInheritanceClassHierarchy()
+    private function doGetInheritanceClassHierarchy()
     {
         $parentClass = $this->getParentClass();
 
--- a/src/TypesFinder/FindTypeFromAst.php
+++ b/src/TypesFinder/FindTypeFromAst.php
@@ -19,10 +19,16 @@ class FindTypeFromAst
      */
     public function __invoke($astType, LocatedSource $locatedSource, $namespace = '')
     {
-        $context = (new ContextFactory())->createForNamespace(
-            $namespace,
-            $locatedSource->getSource()
-        );
+        static $contexts = [];
+
+        if (false === isset($contexts[$namespace])) {
+            $contexts[$namespace] = (new ContextFactory())->createForNamespace(
+                $namespace,
+                $locatedSource->getSource()
+            );
+        }
+
+        $context = $contexts[$namespace];
 
         // @todo Nullable types are effectively ignored - to be fixed
         /* @see https://github.com/Roave/BetterReflection/issues/202 */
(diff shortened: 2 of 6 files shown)


src/Reflection/ReflectionClass.php
```diff
@@ -1014,10 +1016,22 @@ function (Node\Name $interfaceName) {
             : [];
     }
 
+
+    private function getInheritanceClassHierarchy()
+    {
+        if ($this->hierarchy) {
+            return $this->hierarchy;
```
dantleech: I guess there is no harm in this?

src/TypesFinder/FindTypeFromAst.php
```diff
@@ -19,10 +19,16 @@ class FindTypeFromAst
      */
     public function __invoke($astType, LocatedSource $locatedSource, $namespace = '')
     {
-        $context = (new ContextFactory())->createForNamespace(
-            $namespace,
-            $locatedSource->getSource()
-        );
+        static $contexts = [];
```
dantleech: This class is instantiated everytime it is required, and the AST is parsed each time, even when the same namespace is used.

dantleech: @asgrim @Ocramius any ideas for improving this? could this be made into a statefull service?

src/Reflection/ReflectionClass.php
Ocramius: Generally not, but we have to freeze the assumption that source locators are to be considered immutable, and a memoizing wrapper probably needs to be added there...

src/TypesFinder/FindTypeFromAst.php
Ocramius: Would rather inject a context, then let the user decide whether to use a stateless or stateful service

src/Reflection/ReflectionClass.php
dantleech: something like CachedReflectionClass ?

Ocramius: @dantleech yes, that's what I suggested to @asgrim to avoid having a mess of complexity. Not sure how well it plays with the entire reflection traversal...

src/TypesFinder/FindTypeFromAst.php
dantleech: Contexts need to be cached both by namespace and the located source (f.e. with a hash).

Just writing this down for the sake of moving on from this today, I suggest creating a new abstract class, f.e. 

```php
class ContextReflection
{
    public function getNamespaceName()
    {
        // ...
    }

    public function getLocatedSource(): LocatedSource
    {
        // ...
    }
}
```

Both `ReflectionClass` and `AbstractReflectionFunction` could then extend this class and the above methods can be removed from them.

A `CachedContextFactory` could then be injected from the Reflector instance.

Ocramius:
@dantleech can you check this on top of https://github.com/Roave/BetterReflection/pull/338 and see if it is still relevant?

dantleech:
@Ocramius  Created #345

Ocramius:
See #346

Closing as `duplicate` here

Issue #240, opened 2017-01-03
ReflectionParameter doesn't resolve parameter type `self`


malukenho:
It looks like we need some parity between this two objects methods.

Link to reproduce the bug: https://gist.github.com/malukenho/97a19b18b283b633b252aaecfcd4ac6d

Ocramius:
Handled in #241

PR #241, opened 2017-01-09
Fix `self` hinted ReflectionParameter class reference

malukenho:
Linking #240

### DIFF
--- a/src/Reflection/ReflectionParameter.php
+++ b/src/Reflection/ReflectionParameter.php
@@ -504,12 +504,16 @@ class ReflectionParameter implements \Reflector
     public function getClass()
     {
         $hint = $this->getTypeHint();
-        if (!($hint instanceof Types\Object_)) {
+        if (!($hint instanceof Types\Object_  || $hint instanceof Types\Self_)) {
             return null;
         }
 
         if (!$this->reflector instanceof ClassReflector) {
             throw new \RuntimeException('Unable to reflect class type because we were not given a ClassReflector');
+        }
+
+        if ($hint instanceof Types\Self_) {
+            return $this->getDeclaringClass();
         }
 
         return $this->reflector->reflect($hint->getFqsen()->__toString());
--- a/src/Reflection/ReflectionParameter.php
+++ b/src/Reflection/ReflectionParameter.php
@@ -512,6 +512,10 @@ class ReflectionParameter implements \Reflector
             return $this->getDeclaringClass();
         }
 
+        if ('parent' === $hint->getFqsen()->getName()) {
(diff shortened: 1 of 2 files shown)


src/Reflection/ReflectionParameter.php
```diff
@@ -504,14 +504,18 @@ public function getDefaultValueConstantName()
...
-        if (!($hint instanceof Types\Object_)) {
+        if (!($hint instanceof Types\Object_  || $hint instanceof Types\Self_)) {
             return null;
         }
 
         if (!$this->reflector instanceof ClassReflector) {
             throw new \RuntimeException('Unable to reflect class type because we were not given a ClassReflector');
         }
 
+        if ($hint instanceof Types\Self_) {
```
asgrim: This can go before the `if (!$this->reflector instanceof ClassReflector) {` check

test/unit/Reflection/ReflectionParameterTest.php
```diff
@@ -549,4 +549,20 @@ public function testCannotClone()
         $this->expectException(Uncloneable::class);
         $unused = clone $paramInfo;
     }
+
+    public function testGetClassFromSelfTypeHintedProperty()
+    {
+        $content = '<?php class Foo { public function myMethod(self $param) {} }';
+
+        $reflector = new ClassReflector(new AggregateSourceLocator([
```
asgrim: don't need `PhpInternalSourceLocator` do we ?

src/Reflection/ReflectionParameter.php
```diff
@@ -504,10 +504,14 @@ public function getDefaultValueConstantName()
     public function getClass()
     {
         $hint = $this->getTypeHint();
-        if (!($hint instanceof Types\Object_)) {
+        if (!($hint instanceof Types\Object_  || $hint instanceof Types\Self_)) {
```
Ocramius: There should also be a `parent` hint

src/Reflection/ReflectionParameter.php
```diff
@@ -504,10 +504,18 @@ public function getDefaultValueConstantName()
...
             return null;
         }
 
+        if ($hint instanceof Types\Self_) {
+            return $this->getDeclaringClass();
+        }
+
+        if ('parent' === $hint->getFqsen()->getName()) {
```
Ocramius: Moved to https://github.com/phpDocumentor/TypeResolver/issues/27

test/unit/Reflection/ReflectionParameterTest.php
Ocramius: Was removed

PR #136, opened 2015-10-21
Internal refactoring of source locators

asgrim:
Fixes #123 and #43

* Organised the `SourceLocator` folder nicer
* `Reflector\Generic` is now gone
   * Refactored into the `SourceLocator\Ast\Locator` and related classes
* Refactored all `SourceLocators` to use the `SourceLocator\Ast\Locator` that "definitely" returns source

Issue #131, opened 2015-10-11
Make Better Reflection faster


asgrim:
It's pretty slow. Benchmarking a simple `hasMethod` call vs `method_exists` and core reflection, just 1,000 iterations:

```
using method_exists... 0.0006260871887207 seconds.
using core reflection... 0.0013439655303955 seconds.
using better reflection... 7.447105884552 seconds.
```

Pretty pitiful performance if I do say so myself :chart_with_downwards_trend:

Ocramius:
Not sure if it's relevant, tbh.

Reflection is usually to be avoided in runtime context, so I would rather benchmark it only for contexts where performance is actually relevant.

The only context where better reflection may be more interesting to use at runtime is AOP, and there it usually is cached (reflection happens only once, to see whether the file containing a class was changed)

asgrim:
I think we can still do better than 7 seconds, but in my opinion this isn't a high priority thing. Also, I haven't actually had any thoughts on how to improve performance aside from caching stuff more. Additionally, I have no clue about the performance of PhpParser (it may be fast or that may be a bottleneck - if that is a bottleneck, we could compare the performance of the PHP 7 builtin AST parser, although this may mean rewriting a significant amount of internals already.

Ocramius:
Profiling the benchmarks will give you a good overview

asgrim:
Interesting; what sort of use case do you have that means you need to reflect so many items? In honestly, I've only used it in quite specific scenarios so far, so interested to hear what folks are doing with BR.

kdambekalns:
Well, I use annotations to control exporting of "controllers" and domain models via OData. There is a bunch of classes in the system that need to be reflected for their properties and potentially exported methods. I checked, there are 81 model classes and around 80 other classes. Not really "a ton" of classes…

I even filter for a special annotation (using `strpos` on the source) before actually reflecting, but still it takes a lot of time. Unfortunately using the builtin reflection is not possible, due to some crappy code being in that system (where merely _loading_ a class has side-effects already, like actually _running_ some code)…

Ocramius:
The only thing I can think of here is to add a caching layer in the source locators (`CachedSourceLocator`?), but then, not the scope of the lib, IMO.

kukulich:
@Ocramius This can be closed I think, hmm?

Ocramius:
Yup!

Ocramius:
Ref: https://github.com/Roave/BetterReflection/pull/338
[...]

Issue #123, opened 2015-08-31
SourceLocators should return *definitely* located source, not *potentially* located source


asgrim:
Currently the `SourceLocators` return "potentially" located source, which is causing issues such as described in #118 (inability to use multiple `StringSourceLocator`s for example). The fact it is unclear if a located source actually contains the requested identifier is pretty weak and flaky behaviour.

A `SourceLocator` should return "definitely" located source, maybe by peeking at the AST to ensure the requested identifier can be located. Need to be careful to re-use/abstract the logic contained in `Reflector\Generic` so that we don't dupe code and become unperformant.

asgrim:
Also related: #43

PR #245, opened 2017-01-14
[WIP] Cached context

dantleech:
Following from #239 

Currently the `Context` (the namespace and imported aliases for some code) is determined multiple times, for each class, method and parameter. This can cause very serious performance issues.

This PR introduces a native `ContextFactory` interface and provides a `phpdocumentor` implementation and a `CachedContextFactory`. It then makes the factory available (through the reflector) to its creations.

The performance gain is very significant (see linked PR).

Note this is still a WIP and there is an architectural question to be resolved.


src/Reflection/ReflectionClass.php
```diff
@@ -1238,4 +1244,17 @@ public function removeProperty($propertyName)
 
         return false;
     }
+
+    private function findTypeFromAst($namespace, $locatedSource, $type)
```
dantleech: Currently this method is repeated four times, (`ReflectionClass`, `AbstractReflectionFunction`, `ReflectionParameter` and `ReflectionMethod` ? ). We can either create a new base class, or turn the `ContextFactory` into a `TypeFactory`...

src/Reflector/ClassReflector.php
```diff
@@ -69,4 +79,14 @@ public function getAllClasses()
             new IdentifierType(IdentifierType::IDENTIFIER_CLASS)
         );
     }
+
+    /**
+     * Return the class context factory.
+     *
+     * @return ContextFactory
+     */
+    public function getContextFactory()
```
dantleech: The reflector holds the context factory service, and makes it available to it's offspring.

test/unit/Context/PhpDocumentorContextFactoryTest.php
```diff
@@ -0,0 +1,35 @@
...
+namespace Foobar;
+
+use Barfoo as Foo;
+use Foobar as Boo;
+EOT
+        ;
+
+        $innerContextFactory = new ContextFactory();
+
+        $contextFactory = new PhpDocumentorContextFactory($innerContextFactory);
```
dantleech: Remove inner context factory. This object constructs the "php documentor" context factory itself, it has no arguments.

src/Context/ContextFactory.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+namespace Roave\BetterReflection\Context;
+
+interface ContextFactory
```
dantleech: Initially I created a `Context` interface too, but didn't want to push it too far.

dantleech:
All of the `TypesFinder\Find*` classes should be refactored to use this approach.

src/Context/CachedContextFactory.php
```diff
@@ -0,0 +1,31 @@
...
+    {
+        $this->innerFactory = $innerFactory;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function createForNamespace($namespace, $fileContents)
+    {
+        $hash = $namespace . md5($fileContents);
```
Ocramius: At least SHA1, please

src/Context/ContextFactory.php
```diff
@@ -0,0 +1,18 @@
...
+{
+    /**
+     * Build a Context for a namespace in the provided file contents.
+     *
+     * @param string $namespace It does not matter if a `\` precedes the namespace name, this method first normalizes.
```
Ocramius: This is something that should be enforced by a value object, not by a docblock comment and the implementation

src/Reflection/ReflectionClass.php
Ocramius: Extract rather than base class, IMO

src/Reflection/ReflectionFunctionAbstract.php
```diff
@@ -619,4 +619,17 @@ public function getReturnStatementsAst()
 
         return $visitor->getReturnNodes();
     }
+
+    private function findTypeFromAst($namespace, $locatedSource, $type)
+    {
+        $objectType = (new FindTypeFromAst())->__invoke(
```
Ocramius: Variable is redundant

src/Reflection/ReflectionParameter.php
```diff
@@ -530,4 +530,17 @@ public function __clone()
     {
         throw Exception\Uncloneable::fromClass(__CLASS__);
     }
+
+    private function findTypeFromAst($namespace, $locatedSource, $type)
+    {
+        $objectType = (new FindTypeFromAst())->__invoke(
```
Ocramius: Variable is redundant

src/Context/CachedContextFactory.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
+namespace Roave\BetterReflection\Context;
```
Ocramius: What does "context" mean for you? Can a different naming express it better?

src/Reflection/ReflectionFunctionAbstract.php
```diff
@@ -619,4 +619,17 @@ public function getReturnStatementsAst()
...
+            $this->reflector->getContextFactory()->createForNamespace(
```
Ocramius: This is a Demeter violation: can the context be injected instead?

src/Reflector/ClassReflector.php
Ocramius: Disagree with this approach: context is to be injected IMO

src/Context/CachedContextFactory.php
```diff
@@ -0,0 +1,31 @@
...
+
+use Roave\BetterReflection\Context\ContextFactory;
+
+class CachedContextFactory implements ContextFactory
```
Ocramius: Probably "InMemory" to be prepended to the name

src/Reflector/ClassReflector.php
```diff
@@ -18,11 +21,18 @@ class ClassReflector implements Reflector
...
+     * @var ContextFactory
+     */
+    private $contextFactory;
+
+    /**
      * @param SourceLocator $sourceLocator
+     * @param ContextFactory $contextFactory
      */
-    public function __construct(SourceLocator $sourceLocator)
+    public function __construct(SourceLocator $sourceLocator, ContextFactory $contextFactory = null)
```
Ocramius: @asgrim can a BC break be introduced here, to enforce the new parameter?

src/Reflector/ClassReflector.php
```diff
@@ -18,11 +21,18 @@ class ClassReflector implements Reflector
...
     {
         $this->sourceLocator = $sourceLocator;
+        $this->contextFactory = $contextFactory ?: new CachedContextFactory(new PhpDocumentorContextFactory());
```
Ocramius: Need a non-cached as default, I'd say

src/TypesFinder/FindTypeFromAst.php
```diff
@@ -17,13 +18,8 @@ class FindTypeFromAst
      * @param string $namespace
      * @return \phpDocumentor\Reflection\Type|null
      */
-    public function __invoke($astType, LocatedSource $locatedSource, $namespace = '')
+    public function __invoke(Context $context, $astType)
```
Ocramius: BC break to be documented

src/Context/CachedContextFactory.php
```diff
@@ -0,0 +1,31 @@
+<?php
```
asgrim: All new files to be `declare(strict_types=1);` please

src/Context/ContextFactory.php
```diff
@@ -0,0 +1,18 @@
...
+     * @param string $fileContents the file's contents to retrieve the aliases from with the given namespace.
+     *
+     * @see Context for more information on Contexts.
+     *
+     * @return Context
+     */
+    public function createForNamespace($namespace, $fileContents);
```
asgrim: Scalar type declarations & return type here

src/Context/CachedContextFactory.php
```diff
@@ -0,0 +1,31 @@
...
+
+namespace Roave\BetterReflection\Context;
+
+use Roave\BetterReflection\Context\ContextFactory;
+
+class CachedContextFactory implements ContextFactory
+{
+    private $contexts = [];
```
asgrim: can we require [PSR-16 Simple Cache](https://github.com/php-fig/simple-cache) interface for caching here, thus providing better options for caching (and also unburdening ourselves of caching mechanisms)? The only thing we'd have to take care of then is (un)serialisation.

src/Context/CachedContextFactory.php
asgrim: `final` also?

src/Context/PhpDocumentorContextFactory.php
```diff
@@ -0,0 +1,26 @@
+<?php
+
+namespace Roave\BetterReflection\Context;
+
+use phpDocumentor\Reflection\Types\ContextFactory as InnerFactory;
+
+class PhpDocumentorContextFactory implements ContextFactory
```
asgrim: `final` as this is an implementation

src/Reflector/ClassReflector.php
asgrim: Yes, just need a major version bump, but I'm not averse to doing that.

src/TypesFinder/FindTypeFromAst.php
asgrim: I know `FindTypeFromAst` is used outside the context of just reflection, so strong :+1: on this; documentation on how to change existing code for the BC must be provided.

Ocramius:
Just as a tip: let's hold off documentation/BC break documentation until we have the final shape of the feature.

src/Reflection/ReflectionClass.php
dantleech: There are multiple places where the context is envoked for different `TypeFinder\Find*` classes (`FindTypeFromAst`, `FindReturnType`, `FindParameterType`). An a `TypeFinder` class might look like:

```php
class TypeFinder
{
    public function findTypeForAst($ast);
    public function findTypeForReflectionParameter();
    public function findTypeForFunction();
    public function findType...
}
```

Would that be acceptable?

src/Context/CachedContextFactory.php
dantleech: Yeah, it is confusing and I did experiment with better names before giving up. A context is a namespace and a list of imported namespaces and aliases. Perhaps `NamespaceContext` or `NamespaceEnvironment` or <something better here>

src/Context/CachedContextFactory.php
dantleech: For a hash? I assumed MD5 was faster and cryptography isn't the point here?

src/Context/ContextFactory.php
dantleech: This was just a carbon copy of the original, but happy to add value objects.

src/Reflection/ReflectionClass.php
Ocramius: Possibly. It seems like some sort of lookup map, if I understand it correctly.

src/Context/CachedContextFactory.php
Ocramius: Is it namespace based? Or is it just instance-based?

Ocramius: What is the defining criteria for a context? Is it the currently loaded source locators?

src/Context/CachedContextFactory.php
Ocramius: Collision prevention. I'd be fine with having `$fileContents` as key (even better)

src/Context/CachedContextFactory.php
Ocramius: For instance, I don't understand why namespace-based is important.

src/Context/ContextFactory.php
Ocramius: Anything that is a bit more than "just a string" shouldn't be a `string` ;-)

src/Context/CachedContextFactory.php
dantleech: guess that makes sense. In that case would we instantiate the `Reflector` with a single cache implementation? Does it make sense to implement it in this PR or a follow-up?

src/Context/CachedContextFactory.php
dantleech: cool, ok, will experiment.

src/Context/CachedContextFactory.php
dantleech: Its the "namespace environment" of the file, i.e. which names are included and which namespace the file resides in.

src/Reflection/ReflectionFunctionAbstract.php
dantleech: Probably - the thing that makes me hesitate is that the `SourceLocator` classes then have to be aware of both the `Reflector` and the `ContextFactory` (which belongs to the `Reflector`).

The `SourceLocator` classes are currently responsible for instantiating the reflection classes.

I wonder if it is possible to make the `Reflector` classes responsible for creating the `Reflection[Class|Function]` instances, and the `SourceLocator` responsible only for getting `LocatedSource` /cc @asgrim

asgrim: We previously had it so `SourceLocator` returned a `LocatedSource`, but the suggestion was that the `LocatedSource` would definitely contain the expected source (and this is not always the case); so we mulled over things like a `PotentiallyLocatedSource` and so on, but in the end we decided that `SourceLocator` simply returning a reflection, if located, was simpler and cleaner.

An interesting thought: what if one method of loading source (e.g. a `FooSourceLocator`) needs a different way of determining context than another? May not be useful, but injecting the `ContextFactory` at the `SourceLocator` might make more sense?

src/Context/CachedContextFactory.php
Ocramius: If the entire source is used, would it be OK to drop the `$namespace` parameter?

src/Reflection/ReflectionFunctionAbstract.php
Ocramius: Indeed, `new CachedLocator($context, new AggregateSourceLocator([...]))` maybe. That's what would make sense to me, at least

src/Context/CachedContextFactory.php
Ocramius: I would stop at the in-memory caching as-is here. Further caching is not needed in my opinion, and can live in consumer projects instead. I also wouldn't want to deal with serialization here - let someone else have a painful experience.

src/Reflection/ReflectionFunctionAbstract.php
dantleech: >> Indeed, new CachedLocator($context, new AggregateSourceLocator([...]))

`$context` => `$contextFactory` ? 

--

The `ContextFactory` is invoked from the `ReflectionClass` and friends (for each method, parameters, property, etc), so I am not sure how wrapping the `SourceLocator` helps here.

If the locator cannot find the source (in the case of the `Aggregate` source locator) could we not throw a `SourceNotFound` exception?

The reasoning would be that the reflection instances would only be instantiated in one place (the reflectors) - reducing duplication and the responsibility of the locators.

If that makes sense, it is probably quite a large change so I would be happy to inject it (the `ContextFactory`) into the 5 locators that require it or now.

src/Context/CachedContextFactory.php
dantleech: Good point, probably not.

src/Context/CachedContextFactory.php
dantleech: Seems to be the fastest option: https://gist.github.com/dantleech/b442a77dc75c325c83d89b1c2c7db3cb

I assume there will not be any illegal char problems.

src/Reflection/ReflectionClass.php
asgrim: Wouldn't it make more sense to create the context in `FindTypeForAst` helper itself, as it currently does (also means no BC break in that aspect)?

asgrim: Or, inject the context into the `FindType*` classes constructor, and have those injected individually; I'd prefer to not have them all bunched up into one class; they although they serve the same logical purpose, their semantics and usage is still subtly different I think. I'm open to ideas though.

src/Reflection/ReflectionFunctionAbstract.php
asgrim: A `SourceLocator` should not throw any exceptions (unless something is critically broken). They must return either `Reflection*` or `null`. If it throws an exception, it stops further source location from happening. Although as a rule of thumb I throw exceptions all over the place, allowing them to be thrown breaks many combinations of chaining of aggregate/non-aggregate source locators that we have at the moment. Not sure of a nicer solution however, if you've got ideas, I'm all ears. We've already encountered all those problems in early development, hence why they return `Reflection|null` and return the `Reflection` directly; it was the cleaner solution in the end :s

dantleech: Not sure - I think its pretty acceptable to catch exceptions in in this case (e.g. the [CMF chain router](https://github.com/symfony-cmf/routing/blob/master/src/ChainRouter.php#L176)).

Besides, if it can return NULL (f.e. instead of a `LocatedSource`), then I still don't see why the locator should instantiate the reflection class? Sorry, I realise that I am probably missing something here. Will have to investigate further.

src/Reflection/ReflectionClass.php
dantleech: > Wouldn't it make more sense to create the context in FindTypeForAst helper itself

Well, seems unnecessary, as we would just be injecting the factory which would always be invoked in the same way.

> Or, inject the context into the FindType* classes constructor

This is what happens in this PR.

>  I'd prefer to not have them all bunched up into one class

I don't think its particularly elegant either, will search for another solution

src/Reflection/ReflectionFunctionAbstract.php
dantleech: I think the big red light here, for me, is that in order to inject the ContextFactory (or similar) into the `ReflectionClass`, we need to pass it to all of the source locators, and all new `AstLocator` instances and then finally to the `NodeToReflection` class, when it is only required for the creation of the ReflectionClass.

Is there any reason, in principle, that it cannot be refactored to instantiate the `ReflectionClass` in the `Reflector`?

asgrim: I've dug out the original discussions; PR #136 refactored `SourceLocator`s so that they return `Reflection|null` instead of `LocatedSource|null` (i.e. you're suggesting reversing that change). We had issues with the original implementation, #123 and #43 were solved by this change. The reason we changed it is the alternative solution is to load the AST of the source in the `SourceLocator` to check to see if the identifier is in there anyway, before returning the `LocatedSource` (with the assertion that the source has been found)... it makes no sense to parse the AST twice, so we moved the entire functionality into the `SourceLocator` directly. We toyed with the idea of having `PotentiallyLocatedSource` and so on, but those ideas still resulted in flakey, unpredictable behaviour. I can't honestly remember the exact reasons and issues we encountered, but we kept encountering problems, and all those problems disappeared when we moved the "AST to Reflection" conversion into the `SourceLocator`s themselves.

Ignoring the proposals made here for a moment, let's focus on the existing implementation of the `TypeFinders`. The context (i.e. "what namespace are we in, and what are the imports") is an absolute requirement to reliably determine types. At the moment, this functionality is tightly coupled to the instantiation and use of `phpDocumentor\Reflection\Types\ContextFactory`. This is made possible easily in the `Find(Parameter|Property|Return)Type` trio because we store the `LocatedSource` within the `Reflection` class itself. It's not as easy in `FindTypeFromAst` as we're not attaching to anything; but it's a reasonable ask to just inject a `LocatedSource` instance, which results in consistent behaviour across the 4 finders. In reality, these four classes are nothing more than a thin encapsulation that results in using the functionality from `phpDocumentor\Reflection\TypeResolver`.

Okay; lets look back at the initial cause for this PR: **performance**. We already mused a little about performance in #131. What's interesting is that by opening this particular PR, we can see the original implementation for the type finders was not ideal. They're tightly coupled, and don't allow any change in behaviour. The upside of the type finders is that their usage is simple, and that's led to them (or rather, `FindTypeFromAst` being used independently from the reflections itself; purely because it's a neat little encapsulation. In turn, many parts of Better Reflection itself are very tightly coupled to these classes. It is indeed a red light; in order to make those classes "flexible" (so that we can inject a `ContextFactory`), as you say, it involves passing it down the whole stack. I can certainly understand the reasoning for wanting to therefore have the `Reflector` contain the `ContextFactory`, but as Marco already pointed out, this isn't an ideal solution either.

I wonder if time would be better spent taking a step back and re-applying this in a different way. Rather than changing anything external to the type finders, we could I think, in theory, replace the functionality directly in the type finders (which are already tight couplings, so it's not exactly going to make it worse). You could still create a "native" `ContextFactory`, but it would 1) still remain coupled to `phpDocumentor` components and 2) add the caching behaviour in by default for *all* reflections. In essence, this would result in 1) no improvement in coupling (not good) 2) no BC breaks or large refactoring (preferable) 3) potential improvement of performance (good).

Not to say we don't want to go ahead and refactor properly, but if we do then it must be done properly, in which case the `ContextFactory` (or `Context` perhaps?) would indeed need to be injected all the way down the stack; however, this would be a big BC break (like I said before, I'm not averse to bumping a major version to do so), but we just need to be careful to ensure a reasonable upgrade path (and provide documentation). Maybe there is a better alternative solution we've not thought of as well! In any eventuality, I don't think reversing the changes made in PR #136 is a good idea (we'll end up in more pain; unless we can find a better solution there - but it's not a straightforward change I expect). I guess the TL;DR is "all or nothing" with regards to fixing the endemic issues of the type finders here ;)

dantleech: > 3) potential improvement of performance (good).

So, do I understand corrcetly, we instantiate a `[Cached]ContextFactory` for each `Reflection[Class|Function]`. This would be good I think - in most cases it would not benefit to have a "shared" cache between reflection classes as people generally define a single class per file, for functions the performance benefit would be less, but we can still benefit from not parsing the AST for each function argument (and later the same for variable type inference).

We could default to `[InMemory]CachedContextFactory` and potentially still allow injection of the factory which would be compatible with future refactoring.

Did I understand it right?

dantleech: > in which case the ContextFactory (or Context perhaps?) 

I think we will always need a `Context` and `ContextFactory`, we could inject the context on creation of the `Reflection[Thing]` but 

--

btw, this change would also mean injecting the context factory to all the methods, parameters etc. of the reflection, but I guess that was always the case, and it is better that the Reflector would no longer contain the `ContextFactory` - which leads me to wonder if, it is ever refactored in the way I suggested, if the `Reflector` should not also be responsible for instantiating the methods, parameters, etc, but I digress :)

asgrim: > as people generally define a single class per file

This is a reasonable assumption, but the library *must* cater for those with multiple classes.

What I was suggesting is instantiate the `[Cached]ContextFactory` directly in the type finder classes, as it already does (with phpdocumentor's own one). Yes, it'd be a tight coupling, but it already is anyway, so we're not making it worse - i.e. you'd *just* be introducing the performance increase by caching the contexts.

> but we can still benefit from not parsing the AST

I think caching the AST parsing should be dealt with separately; it's distinct from the type resolving.

dantleech: > but the library must cater for those with multiple classes.

It would do - just that each reflection class would have it's own context factory and not benefit from a shared cache.

> What I was suggesting is instantiate the [Cached]ContextFactory directly in the type finder classes

This would mean that each TypeFinder would have it's own CachedContextFactory, this would not convey any performance benefit (?)

dantleech:
Updated, instantiating the `ContextFactory` in the [NodeToReflection](https://github.com/dantleech/BetterReflection/blob/8a5c4151ea11ac4b04a46806ff69e409425a939d/src/SourceLocator/Ast/Strategy/NodeToReflection.php#L25) class. This means that `Reflection[Class|Function]` instances receive a `Context` (instead of a `ContextFactory`).

The `Context` replaces the `$namespace` argument, and I added a new method `getContext` to the `ReflectionClass`.

I am thinking to add a `Context` interface and create a lazy context implementation (as sometimes it may not be required at all).

asgrim:
So I've been thinking about this; I wonder if it might make more sense to have a "Reflection Manager" (concept similar to `EntityManager` in Doctrine) that we cache reflections as a whole... it's something I considered in the past and I wanted to avoid (I'd prefer to keep things stateless), but the more I think about it, the more it can probably make things much faster; this way the cache can load up reflections quickly rather than re-reflecting stuff. It's pretty much a radical change from this proposal I understand. I wonder what might be better. Feedback welcome @dantleech @Ocramius

Ocramius:
@asgrim fwiw, PHP-SRC does something like that when first creating a `ReflectionClass` (which is also why the myth of "reflection being slow" was born - the first reflection of a class is actually slow).

I am currently too unfocused to follow the issue, sorry :-(

dantleech:
I think caching the reflection classes side-steps the issue - although it may be beneficial in any case.

The main bottle-neck is re-parsing the AST every time a type is resolved.

I experimented with a new approach:

    https://github.com/dantleech/worse-reflection/blob/master/lib/SourceContext.php

The SourceContext passed to all reflections. The AST is passed once per-file (no caching).

It does *not* use the docblock parser, which iirc is not performant (don't quote me) in which case would consider rewriting.

On 16 February 2017 20:11:07 GMT+00:00, James Titcumb <notifications@github.com> wrote:
>So I've been thinking about this; I wonder if it might make more sense
>to have a "Reflection Manager" (concept similar to `EntityManager` in
>Doctrine) that we cache reflections as a whole... it's something I
>considered in the past and I wanted to avoid (I'd prefer to keep things
>stateless), but the more I think about it, the more it can probably
>make things much faster; this way the cache can load up reflections
>quickly rather than re-reflecting stuff. It's pretty much a radical
>change from this proposal I understand. I wonder what might be better.
>Feedback welcome @dantleech @Ocramius 
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/Roave/BetterReflection/pull/245#issuecomment-280444923

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

Ocramius:
Closing here: moved to https://github.com/Roave/BetterReflection/pull/338 for now. Note that #338 does not fix caching of PhpDocumentor types, but only AST/reflection.

PR #246, opened 2017-02-05
Fixed class constant value fetch when constant is declared in parent

janlanger:
Class constant value referenced via `self` was not fetched properly when the constant is declared in parent class.

### DIFF
--- a/src/NodeCompiler/CompileNodeToValue.php
+++ b/src/NodeCompiler/CompileNodeToValue.php
@@ -124,7 +124,7 @@ class CompileNodeToValue
 
         $classInfo = null;
         if ('self' === $className || 'static' === $className) {
-            $classInfo = $context->getSelf();
+            $classInfo = $this->getConstantDeclaringClass($node->name, $context);
         }
 
         if (null === $classInfo) {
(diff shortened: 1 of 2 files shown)


src/NodeCompiler/CompileNodeToValue.php
```diff
@@ -125,6 +125,9 @@ private function compileClassConstFetch(Node\Expr\ClassConstFetch $node, Compile
         $classInfo = null;
         if ('self' === $className || 'static' === $className) {
             $classInfo = $context->getSelf();
+            while (!$classInfo->hasConstant($node->name) && $classInfo->getParentClass() !== null) {
```
Ocramius: Needs to be moved to a private method

test/unit/NodeCompiler/CompileNodeToValueTest.php
```diff
@@ -310,4 +310,23 @@ class Bat {
         $classInfo = $reflector->reflect('Bar\Bat');
         $this->assertSame('My\Awesome\Foo', $classInfo->getConstant('QUX'));
     }
+
+    public function testClassConstantResolutionFromParent()
```
Ocramius: Can you please add a test with a third inheritance level?

janlanger:
@Ocramius updated

Ocramius:
@janlanger 👍

PR #247, opened 2017-02-10
Failing test - constant from different class as default value

janlanger:
If property default value is class constant defined in different class, call `propertyRefl->getDefaultValue()` fails with `"Foo" could not be found in the located source`. 

It only happens when "Foo:BAR" is not referenced as FQN.

I didn't figure out how to fix it, so at least failing test case...


asgrim:
Thanks @janlanger for the test, this is useful, will investigate :)

asgrim:
I've opened new PR to fix the reported issue in #248 with your test, awesome - thanks! :D

PR #248, opened 2017-02-16
Constant default value failing test

asgrim:
Building on PR #247 - fixes test from @janlanger where `ClassConstFetch` node could not be resolved as we were not looking up the full qualified class name. Add some additional code into `CompileNodeToValue` to resolve this using the `FindTypeFromAst`.


Ocramius:
LGTM 👍

PR #253, opened 2017-03-30
Usage: fix DirectoriesSourceLocator usage

TomasVotruba:
My first usage of this package in years. Let me know if there is another way

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -188,25 +188,14 @@ $reflector = new ClassReflector($singleDirectorySourceLocator);
 $classes = $reflector->getAllClasses();
 ```
 
-### Fetch reflections of all the classes in a directory (including sub directories)
+### Fetch reflections of all the classes in one or more directories (including sub directories)
 
 ```php
 <?php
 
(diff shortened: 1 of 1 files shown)


TomasVotruba:
Thank you!

Btw, does this work only on class reflections? 

I need also functions and constants, but there is no such method in `FunctionReflector`.

Ocramius:
@TomasVotruba source locators work with any kind of AST node, but we didn't implement constant lookup (only functions and classes, so far)

TomasVotruba:
@Ocramius I see, thanks.

[I didn't find in docs](https://github.com/Roave/BetterReflection/blob/master/docs/usage.md#reflecting-functions), how to get functions from specific directory.


Sth like:

```php
$reflector->getAllFunctions();
```

Is there a way?

asgrim:
@TomasVotruba yes - it was recently added in #252 but it's not been tagged yet :)

TomasVotruba:
@asgrim Perfect, thank you! I can roll on dev for a while :)

Issue #255, opened 2017-03-31
create ReflectionParameter from closure


marcosh:
At the moment trying to call `ReflectionParameter::createFromSpec` with a `\Closure` throws an exception.

Is there any particular reason why is it so?

asgrim:
Yep, as you discovered (and fixed in PR #256) we just never got around to implementing ;)

Ocramius:
Closing as per #256

PR #256, opened 2017-03-31
create reflection parameter from closure

marcosh:
solves #255

### DIFF
--- a/src/Reflection/ReflectionParameter.php
+++ b/src/Reflection/ReflectionParameter.php
@@ -105,7 +105,7 @@ class ReflectionParameter implements \Reflector
      * @param string $parameterName
      * @return ReflectionParameter
      */
-    public static function createFormClosure(\Closure $closure, $parameterName)
+    public static function createFromClosure(\Closure $closure, string $parameterName): ReflectionParameter
     {
         return ReflectionFunction::createFromClosure($closure)
             ->getParameter($parameterName);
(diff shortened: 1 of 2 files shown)


src/Reflection/ReflectionParameter.php
```diff
@@ -99,6 +99,19 @@ public static function createFromClassInstanceAndMethod($instance, $methodName,
     }
 
     /**
+     * Create a reflection of a parameter using a closure
+     *
+     * @param \Closure $closure
+     * @param string $parameterName
+     * @return ReflectionParameter
+     */
+    public static function createFormClosure(\Closure $closure, $parameterName)
```
asgrim: * typo in method name `createFromClosure`
 * also please add type declarations (`string $parameterName` and ` : ReflectionParameter`) (I'm working on adding type declarations everywhere else as we're min PHP 7 for next release now)

test/unit/Reflection/ReflectionParameterTest.php
```diff
@@ -49,6 +49,14 @@ public function testCreateFromClassInstanceAndMethod()
         $this->assertSame('index', $parameterInfo->getName());
     }
 
+    public function testCreateFromCallable()
+    {
+        $parameterInfo = ReflectionParameter::createFormClosure(function ($a) {}, 'a');
+
+        $this->assertInstanceOf(ReflectionParameter::class, $parameterInfo);
```
asgrim: please use `self::assertInstanceOf`

test/unit/Reflection/ReflectionParameterTest.php
```diff
@@ -49,6 +49,14 @@ public function testCreateFromClassInstanceAndMethod()
...
+        $this->assertSame('a', $parameterInfo->getName());
```
asgrim: please use `self::assertSame`

marcosh:
fixed the minor changes

PR #267, opened 2017-04-16
Testing that class constant visibility works

### LINKED ISSUE #204: Review PHP 7.1 feature: Class constant visibility modifiers
https://wiki.php.net/rfc/class_const_visibility

Ensure we support new PHP 7.1 feature for class constant visibility modifiers.

asgrim:
Fixes #204

Works already (as class constant visibility was added in PhpParser 3.0.0+ a while ago), no new reflection API (that I could find) to determine visibility of a class constant, so just added a test to ensure we get all class constants we expect, regardless of visibility.

### DIFF
--- a/test/unit/Reflection/ReflectionClassTest.php
+++ b/test/unit/Reflection/ReflectionClassTest.php
@@ -1142,4 +1142,39 @@ class ReflectionClassTest extends \PHPUnit_Framework_TestCase
         self::assertTrue($reflection->hasProperty('staticBar'));
         self::assertTrue($reflection->getProperty('staticBar')->isStatic());
     }
+
+    public function testGetConstantsReturnsAllConstantsRegardlessOfVisibility()
+    {
+        $php = '<?php
+            class Foo {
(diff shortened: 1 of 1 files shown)


test/unit/Reflection/ReflectionClassTest.php
```diff
@@ -1142,4 +1142,39 @@ class Foo {
         self::assertTrue($reflection->hasProperty('staticBar'));
         self::assertTrue($reflection->getProperty('staticBar')->isStatic());
     }
+
+    public function testGetConstantsReturnsAllConstantsRegardlessOfVisibility()
```
Ocramius: Is this also the PHP internal behaviour?

asgrim: Yes: http://php.net/manual/en/reflectionclass.getconstants.php

> Gets all defined constants from a class, regardless of their visibility.

Ocramius:
:ship:

PR #268, opened 2017-04-17
Ensure locateFunctionByName returns ?string always

### LINKED ISSUE #237: `AutoloadSourceLocator` reports `false` on internal functions: should report `null` as per interface definition
As per https://github.com/Roave/BetterReflection/pull/236/files#r93817241, [AutoloadSourceLocator#attemptAutoloadForIdentifier()](https://github.com/Roave/BetterReflection/blob/d6784562fa91cf45ec4833de8c490c6b681a69ed/src/SourceLocator/Type/AutoloadSourceLocator.php#L47) has a type signature of `string|null`, whereas internal reflection usage causes it to produce `string|null|bool`.

Either a fix in the type signature or in the implementations that should return `string|null` is needed.

asgrim:
Fixes #237 

Now that we have strict type declarations, I don't think an additional test is necessary - this is essentially a bugfix now IMO...

### DIFF
--- a/src/SourceLocator/Type/AutoloadSourceLocator.php
+++ b/src/SourceLocator/Type/AutoloadSourceLocator.php
@@ -45,6 +45,7 @@ class AutoloadSourceLocator extends AbstractSourceLocator
      *
      * @param Identifier $identifier
      * @return string|null
+     * @throws \ReflectionException
      */
     private function attemptAutoloadForIdentifier(Identifier $identifier) : ?string
     {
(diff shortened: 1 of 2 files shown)


src/SourceLocator/Type/AutoloadSourceLocator.php
```diff
@@ -113,7 +116,13 @@ private function locateFunctionByName(string $functionName) : ?string
         }
 
         $reflection = new \ReflectionFunction($functionName);
-        return $reflection->getFileName();
+        $reflectionFileName = $reflection->getFileName();
+
+        if (! is_string($reflectionFileName)) {
```
Ocramius: Test? :-\

asgrim: Added.

Ocramius:
:ship:

Issue #231, opened 2016-12-07
Inherited properties not returned by getProperties


dantleech:
Currently only properties from the "top" class are returned by `ReflectionClass#getProperties`, in contrast with `getMethods` which returns inherited methods.

asgrim:
Note: when fixing this, we need to exclude `private` properties in parent classes to replicate PHP behaviour:

```php
class Foo
{
  private $a;
}

class Bar extends Foo
{
  public $b;
}

$r = new \ReflectionClass(Bar::class);
var_dump($r->getProperties());
```

In this example, `$a` is not output, but if it's changed to `protected` or `public`, it is visible (as one would expect by the inheritance rules)

asgrim:
Handled in #271

PR #271, opened 2017-05-06
getProperties returns inherited properties

marcosh:
solves #231

### DIFF
--- a/src/Reflection/ReflectionClass.php
+++ b/src/Reflection/ReflectionClass.php
@@ -475,6 +475,25 @@ class ReflectionClass implements Reflection, \Reflector
     }
 
     /**
+     * Get only the properties for this specific class (i.e. do not search
+     * up parent classes etc.)
+     *
+     * @return ReflectionProperty[]
+     */
+    public function getImmediateProperties() : array
+    {
+        $properties = [];
+        foreach ($this->node->stmts as $stmt) {
+            if ($stmt instanceof PropertyNode) {
+                $prop = ReflectionProperty::createFromNode($this->reflector, $stmt, $this);
+                $properties[$prop->getName()] = $prop;
+            }
+        }
+
+        return $properties;
+    }
+
+    /**
      * Get the properties for this class.
      *
      * @return ReflectionProperty[]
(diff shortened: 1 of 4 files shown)


src/Reflection/ReflectionClass.php
```diff
@@ -493,11 +490,51 @@ public function getProperties() : array
...
+                    function (ReflectionClass $trait) {
+                        return $trait->getProperties();
+                    },
+                    $this->getTraits()
+                )
+            ),
+            $this->getImmediateProperties()
+        );
+
+        $this->cachedProperties = $inheritedProperties;
```
Ocramius: Can return from this expression directly

src/Reflection/ReflectionClass.php
```diff
@@ -475,16 +475,13 @@ public function getConstructor() : ReflectionMethod
...
-     * Get the properties for this class.
+     * Get only the properties for this specific class (i.e. do not search
+     * up parent classes etc.)
      *
      * @return ReflectionProperty[]
      */
-    public function getProperties() : array
+    public function getImmediateProperties() : array
     {
-        if (null !== $this->cachedProperties) {
```
asgrim: Doesn't it make more sense to cache immediate properties, so they're cached at each level of the inheritance tree, rather than caching the whole tree at each level?

test/unit/Reflection/ReflectionClassTest.php
```diff
@@ -206,6 +206,44 @@ public function testGetProperties() : void
         self::assertCount(4, $properties);
     }
 
+    public function testGetPropertiesReturnsInheritedProperties() : void
+    {
+        $reflector = new ClassReflector(new SingleFileSourceLocator(__DIR__ . '/../Fixture/InheritedClassProperties.php'));
+        $classInfo = $reflector->reflect('Qux');
```
asgrim: Can use `::class` here (and below in assertions)

TomasVotruba: Just wondering, do you know about sniff/fixer for this?

Ocramius: Phpstorm ea inspections detects these

TomasVotruba: I'd love to have in CI.

Could you share the name? I'm unable to guess it: https://github.com/kalessil/phpinspectionsea/tree/master/src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors

marcosh:
fixed the changes you suggested

asgrim:
Nice! Thanks, and great fix! :)

Issue #103, opened 2015-08-17
ReflectionClass getMethods() and getProperties() do not filter


asgrim:
http://php.net/manual/en/reflectionclass.getmethods.php
http://php.net/manual/en/reflectionclass.getproperties.php

The `getMethods()` and `getProperties()` both do not actually filter at all - they need to accept an integer parameter.

``` php
// Fetch all private methods
$classInfo->getMethods(ReflectionMethod::IS_PRIVATE);
```

Ocramius:
Handled in #275

PR #275, opened 2017-05-17
Filter get methods and properties

marcosh:
solves #103

### DIFF
--- a/src/Reflection/ReflectionClass.php
+++ b/src/Reflection/ReflectionClass.php
@@ -338,11 +338,23 @@ class ReflectionClass implements Reflection, \Reflector
     /**
      * Fetch an array of all methods for this class.
      *
+     * @param ?int
      * @return ReflectionMethod[]
      */
-    public function getMethods() : array
+    public function getMethods(?int $filter = null) : array
     {
-        return array_values($this->getMethodsIndexedByName());
+        if (null === $filter) {
+            return array_values($this->getMethodsIndexedByName());
+        }
+
+        return array_values(
+            array_filter(
+                $this->getMethodsIndexedByName(),
+                function (ReflectionMethod $method) use ($filter) {
+                    return $filter & $method->getModifiers();
+                }
+            )
+        );
     }
 
     /**
--- a/src/Reflection/ReflectionClass.php
+++ b/src/Reflection/ReflectionClass.php
@@ -363,7 +363,7 @@ class ReflectionClass implements Reflection, \Reflector
      *
      * @return ReflectionMethod[]
      */
-    public function getImmediateMethods() : array
+    public function getImmediateMethods(?int $filter = null) : array
     {
         /* @var $methods \ReflectionMethod[] */
         $methods = array_map(
@@ -376,7 +376,9 @@ class ReflectionClass implements Reflection, \Reflector
         $methodsByName = [];
 
         foreach ($methods as $method) {
-            $methodsByName[$method->getName()] = $method;
+            if (null === $filter || $filter & $method->getModifiers()) {
+                $methodsByName[$method->getName()] = $method;
+            }
         }
 
         return $methodsByName;
--- a/src/Reflection/ReflectionClass.php
+++ b/src/Reflection/ReflectionClass.php
@@ -338,7 +338,7 @@ class ReflectionClass implements Reflection, \Reflector
     /**
      * Fetch an array of all methods for this class.
      *
-     * @param ?int
+     * @param ?int $filter
      * @return ReflectionMethod[]
      */
     public function getMethods(?int $filter = null) : array
@@ -361,6 +361,7 @@ class ReflectionClass implements Reflection, \Reflector
      * Get only the methods that this class implements (i.e. do not search
      * up parent classes etc.)
      *
+     * @param ?int $filter
      * @return ReflectionMethod[]
      */
     public function getImmediateMethods(?int $filter = null) : array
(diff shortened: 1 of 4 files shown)


src/Reflection/ReflectionClass.php
```diff
@@ -338,20 +338,33 @@ private function getMethodsIndexedByName() : array
...
      * Fetch an array of all methods for this class.
      *
+     * @param ?int $filter
      * @return ReflectionMethod[]
      */
-    public function getMethods() : array
+    public function getMethods(?int $filter = null) : array
     {
-        return array_values($this->getMethodsIndexedByName());
+        if (null === $filter) {
```
roukmoute: Use `is_null()` function ?

Ocramius: @roukmoute no, please don't do that, unless you aren't using `array_map()` :-P

src/Reflection/ReflectionClass.php
```diff
@@ -338,20 +338,33 @@ private function getMethodsIndexedByName() : array
     /**
      * Fetch an array of all methods for this class.
      *
+     * @param ?int $filter
```
Ocramius: The type is `int|null`

Ocramius: More documentation is needed on `$filter`, with examples, since this is a bitmask.

src/Reflection/ReflectionClass.php
```diff
@@ -338,20 +338,33 @@ private function getMethodsIndexedByName() : array
...
+                }
+            )
+        );
     }
 
     /**
      * Get only the methods that this class implements (i.e. do not search
      * up parent classes etc.)
      *
+     * @param ?int $filter
```
Ocramius: Same remarks here

src/Reflection/ReflectionClass.php
```diff
@@ -478,40 +493,47 @@ public function getConstructor() : ReflectionMethod
      * Get only the properties for this specific class (i.e. do not search
      * up parent classes etc.)
      *
+     * @param ?int $filter
```
Ocramius: same

src/Reflection/ReflectionClass.php
```diff
@@ -478,40 +493,47 @@ public function getConstructor() : ReflectionMethod
...
             }
+
+            $this->cachedProperties = $properties;
         }
 
-        return $this->cachedProperties = $properties;
+        return array_filter(
+            $this->cachedProperties,
+            function (ReflectionProperty $property) use ($filter) {
+                return null === $filter || $filter & $property->getModifiers();
```
Ocramius: Weird that in this scenario you went for the "always true" condition in the filter callback

src/Reflection/ReflectionClass.php
```diff
@@ -478,40 +493,47 @@ public function getConstructor() : ReflectionMethod
...
+            }
+        );
     }
 
     /**
      * Get the properties for this class.
      *
+     * @param ?int $filter
```
Ocramius: same

src/Reflection/ReflectionObject.php
```diff
@@ -95,9 +95,10 @@ public static function createFromInstance($object)
     /**
      * Reflect on runtime properties for the current instance
      *
+     * @param ?int $filter
```
Ocramius: same - maybe we should use `@see` and point to a single location

src/Reflection/ReflectionClass.php
roukmoute: @Ocramius But I've never understood why you prefer to test `null` while a function exists.
Can you clarify me on this point?

Ocramius: @roukmoute `is_null()` is a function: while it allows for function composition, calling functions in PHP is the slowest thing ever, and the first/simplest thing you can do to remove `O(1)` overhead on any codebase.

To clarify:

```php
// good
array_keys(array_filter([null, 1, 2, null], 'is_null'));

// bad
if (is_null($something)) {
    // ...
}

// good
if (null === $something) {
    // ...
}
```

roukmoute: Thanks @Ocramius for your great explanation :)

But I've few remarks:
- `is_null()` does the job in C, and the comparison is not the same things. In fact is it really important to optimize this kind of things? (This is a real question :smile: )
- You talking about big O notation, is it really necessary to think to this low level ? But I do a [benchmark](https://gist.github.com/roukmoute/a65eb6749a8f506ac048b008f07dc27a), and finally with latest version of PHP, `is_null` is more fastest.

Here is a small [screencast](https://asciinema.org/a/08670iigxidrc3okzftxrijsq) about that.

Ocramius: @roukmoute you forgot to namespace it :-)

src/Reflection/ReflectionObject.php
marcosh: would it be ok to point directly to [http://php.net/manual/en/reflectionclass.getmethods.php](http://php.net/manual/en/reflectionclass.getmethods.php) where the parameter meaning is explained?

src/Reflection/ReflectionClass.php
lisachenko: @Ocramius Can not agree with you that calling special functions are slow. Because you missing the presence of Optimizer in PHP. So if you will use `\is_null($something);` then it won't be function call at all, because PHP optimizer will inline it as simple type check.

To prove it, check https://3v4l.org/Ivbid/vld#output to see `TYPE_CHECK` opcode for that. And it's very fast 😄  But without leading slash it will be slow function call: https://3v4l.org/asV6T/vld#output

Ocramius:
Well aware of that optimisation step, still a lot of noise though.

marcosh:
@Ocramius I updated the PR with the fixes you suggested

src/Reflection/ReflectionClass.php
Ocramius: Yes, well aware of the optimisation - still a huge load of noise for nothing IMO

Ocramius:
@marcosh awesome work! :+1:

PR #278, opened 2017-05-29
object immediate properties includes runtime properties

marcosh:


test/unit/Reflection/ReflectionObjectTest.php
```diff
@@ -106,6 +106,19 @@ public function testGetRuntimePropertiesWithFilter() : void
...
 
+    public function testGetRuntimeImmediatePropertiesWithFilter() : void
+    {
+        $foo = new \StdClass();
+        $foo->bar = 'huzzah';
+
+        $classInfo = ReflectionObject::createFromInstance($foo);
+
+        self::assertEmpty($classInfo->getImmediateProperties(\ReflectionProperty::IS_STATIC));
+        self::assertCount(1, $classInfo->getImmediateProperties(\ReflectionProperty::IS_PUBLIC));
```
Ocramius: Uhm, I am not sure if this should be in the same reflection API, since runtime access will fail when done via normal reflection.

Should probably be a separate API endpoint.

marcosh: sorry @Ocramius but your comment is not clear to me...

In normal reflection, the method `getImmediateProperties` does not exist, and using `getProperties` the test would pass.

What do you mean by `runtime access will fail`?

Ocramius: Oh, this is my mistake! I thought that these examples would fail, but they don't:

https://3v4l.org/b7PUE

```php
<?php

$foo = (object) ['bar' => 'baz'];

$property = new \ReflectionProperty($foo, 'bar');

var_dump($property->getValue($foo));
```

```php
```

test/unit/Reflection/ReflectionObjectTest.php
```diff
@@ -106,6 +106,19 @@ public function testGetRuntimePropertiesWithFilter() : void
         self::assertEmpty($classInfo->getProperties(\ReflectionProperty::IS_PRIVATE));
     }
 
+    public function testGetRuntimeImmediatePropertiesWithFilter() : void
+    {
+        $foo = new \StdClass();
```
asgrim: Sorry, I know it's very nitpicky, but please use `\stdClass`

marcosh: done!

Ocramius:
LGTM :+1:

PR #280, opened 2017-06-01
[NodeCompiler] add support for __DIR__

TomasVotruba:
Parser is failing at the moment on misisng NodeCompiler for `__DIR__`:

```php
class ConstantInClass
{
    /**
     * @var string
     */
   protected const COMPOSED_WITH_DIR = __DIR__ . '/here';
}
```

This PR fixes it. Test included.

### DIFF
--- a/src/NodeCompiler/CompileNodeToValue.php
+++ b/src/NodeCompiler/CompileNodeToValue.php
@@ -270,8 +270,7 @@ class CompileNodeToValue
      */
     private function compileDirConstant(Node\Scalar\MagicConst\Dir $node, CompilerContext $context): string
     {
-        $fileName = $context->getSelf()->getLocatedSource()->getFileName();
-        return dirname(realpath($fileName));
+        return dirname(realpath($context->getFileName()));
     }
 
     private function getConstantDeclaringClass(string $constantName, ReflectionClass $class) : ?ReflectionClass
--- a/src/NodeCompiler/CompileNodeToValue.php
+++ b/src/NodeCompiler/CompileNodeToValue.php
@@ -47,6 +47,10 @@ class CompileNodeToValue
             return $this->compileBinaryOperator($node, $context);
         }
 
+        if ($node instanceof Node\Scalar\MagicConst\Dir) {
+            return '__DIR__';
+        }
+
         throw new Exception\UnableToCompileNode('Unable to compile expression: ' . get_class($node));
     }
 
--- a/src/NodeCompiler/CompileNodeToValue.php
+++ b/src/NodeCompiler/CompileNodeToValue.php
@@ -48,7 +48,7 @@ class CompileNodeToValue
         }
 
         if ($node instanceof Node\Scalar\MagicConst\Dir) {
-            return '__DIR__';
+            return $this->compileDirConstant($node, $context);
         }
 
         throw new Exception\UnableToCompileNode('Unable to compile expression: ' . get_class($node));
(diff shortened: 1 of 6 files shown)


src/NodeCompiler/CompileNodeToValue.php
```diff
@@ -47,6 +47,10 @@ public function __invoke(Node $node, CompilerContext $context)
             return $this->compileBinaryOperator($node, $context);
         }
 
+        if ($node instanceof Node\Scalar\MagicConst\Dir) {
+            return '__DIR__';
```
asgrim: Hmm, not sure this is correct; surely it is the directory in which the context's class sits in, not just the string literal `__DIR__`?

TomasVotruba: I realize this doesn't provide a valid case. It just works on my machine.

How would a test for that look like to comply with current test set?

asgrim: I'd provide a test case where the real value of `__DIR__` compiled by the `NodeCompiler` matches the test suite class's path

TomasVotruba: I see. Test case fixed and now it fails.

But I'm not sure how to get `__DIR__` value from `$context` or `$node`.

asgrim: If we are in the context of a class, `$context->getSelf()` returns the class, from which you can get the filename. Currently wouldn't work for a function though. It wouldn't be too difficult to also inject a filename though into `CompilerContext` constructor.

TomasVotruba: > If we are in the context of a class.

So write a test for class context, as I stated in code above?

asgrim: Yep! :)

TomasVotruba: Allright :)
, let me a get :coffee:

TomasVotruba: My attemt submitted. Check it!

src/NodeCompiler/CompileNodeToValue.php
```diff
@@ -257,6 +261,19 @@ private function compileBinaryOperator(Node\Expr\BinaryOp $node, CompilerContext
...
+    /**
+     * Compile a __DIR__ node
+     *
+     * @param Node\Scalar\MagicConst\Dir $node
+     * @param CompilerContext $context
+     * @return string
+     */
+    private function compileDirConstant(Node\Scalar\MagicConst\Dir $node, CompilerContext $context): string
+    {
+        $fileName = $context->getSelf()->getLocatedSource()->getFileName();
```
asgrim: Try to follow the law of Demeter here; `ReflectionClass` has a method `getFileName()` you can call to avoid using the located class API.

In fact I'd suggest making a new method `getFileName` on the `CompilerContexf` (that calls `ReflectionClass#getFileName`, not the `LocatedSource` API)

TomasVotruba: > has a method getFileName() you can call to avoid using the located class API.

Sure, just didn't know about that.

> In fact I'd suggest making a new method...

Allright, makes sense.

TomasVotruba:
What about now?

TomasVotruba:
Awesome. Thanks for guidance :+1:

Issue #281, opened 2017-06-07
[PHP7.1] ::getReflectionConstant() and ::getReflectionConstants() support


POPSuL:
Now we have public/private/protected constants in classes, but we don't have opportunity for getting visibility level.

Related:
https://bugs.php.net/bug.php?id=74261
#204
https://github.com/ApiGen/ApiGen/issues/877

asgrim:
Thanks for the report; good suggestion for compatibility

POPSuL:
I will do pr to solve this issue within an hour

Ocramius:
Handled in #282

PR #282, opened 2017-06-07
[PHP 7.1] Add support for ReflectionClassConstant (#281)

POPSuL:
https://github.com/Roave/BetterReflection/issues/281 

Note: I did not understand how testReflectionObjectOverridesAllMethodsInReflectionClass works, therefore I missed it.

### DIFF
--- a/src/Reflection/Adapter/ReflectionClassConstant.php
+++ b/src/Reflection/Adapter/ReflectionClassConstant.php
@@ -7,13 +7,16 @@ use Roave\BetterReflection\Reflection\ReflectionClassConstant as BetterReflectio
 
 class ReflectionClassConstant extends CoreReflectionClassConstant
 {
+
+    /**
+     * @var BetterReflectionClassConstant
+     */
     private $betterClassConstant;
 
     public function __construct(BetterReflectionClassConstant $betterClassConstant)
     {
         $this->betterClassConstant = $betterClassConstant;
     }
-
 
     /**
      * Get the name of the reflection (e.g. if this is a ReflectionClass this
--- a/src/Reflection/Adapter/ReflectionClassConstant.php
+++ b/src/Reflection/Adapter/ReflectionClassConstant.php
@@ -1,0 +1,110 @@
+<?php
+
+namespace Roave\BetterReflection\Reflection\Adapter;
+
+use \ReflectionClassConstant as CoreReflectionClassConstant;
+use Roave\BetterReflection\Reflection\ReflectionClassConstant as BetterReflectionClassConstant;
+
+class ReflectionClassConstant extends CoreReflectionClassConstant
+{
+    private $betterClassConstant;
+
+    public function __construct(BetterReflectionClassConstant $betterClassConstant)
+    {
+        $this->betterClassConstant = $betterClassConstant;
+    }
+
+
+    /**
+     * Get the name of the reflection (e.g. if this is a ReflectionClass this
+     * will be the class name).
+     *
+     * @return string
+     */
+    public function getName() : string
+    {
+        return $this->betterClassConstant->getName();
+    }
+
+    /**
+     * Returns constant value
+     *
+     * @return mixed
+     */
+    public function getValue()
+    {
+        return $this->betterClassConstant->getValue();
+    }
+
+    /**
+     * Constant is public
+     *
+     * @return bool
+     */
+    public function isPublic() : bool
+    {
+        return $this->betterClassConstant->isPublic();
+    }
+
+    /**
+     * Cosnstant is private
+     *
+     * @return bool
+     */
+    public function isPrivate() : bool
+    {
+        return $this->betterClassConstant->isPrivate();
+    }
+
+    /**
+     * Constant is protected
+     *
+     * @return bool
+     */
+    public function isProtected() : bool
+    {
+        return $this->betterClassConstant->isProtected();
+    }
+
+    /**
+     * Returns a bitfield of the access modifiers for this constant
+     *
+     * @return int
+     */
+    public function getModifiers() : int
+    {
+        return $this->betterClassConstant->getModifiers();
+    }
+
+    /**
+     * Get the declaring class
+     *
+     * @return ReflectionClass
+     */
+    public function getDeclaringClass() : ReflectionClass
+    {
+        return new ReflectionClass($this->betterClassConstant->getDeclaringClass());
+    }
+
+    /**
+     * Returns the doc comment for this constant
+     *
+     * @return string
+     */
+    public function getDocComment() : string
+    {
+        return $this->betterClassConstant->getDocComment();
+    }
+
+    /**
+     * To string
+     *
+     * @link http://php.net/manual/en/reflector.tostring.php
+     * @return string
+     * @since 5.0
+     */
+    public function __toString()
+    {
+        return $this->betterClassConstant->__toString();
+    }
+}
--- a/src/Reflection/ReflectionClass.php
+++ b/src/Reflection/ReflectionClass.php
@@ -2,6 +2,18 @@
 
 namespace Roave\BetterReflection\Reflection;
 
+use phpDocumentor\Reflection\Types\Object_;
+use PhpParser\Node;
+use PhpParser\Node\Stmt\Class_ as ClassNode;
+use PhpParser\Node\Stmt\ClassConst;
+use PhpParser\Node\Stmt\ClassConst as ConstNode;
+use PhpParser\Node\Stmt\ClassLike as ClassLikeNode;
+use PhpParser\Node\Stmt\ClassMethod;
+use PhpParser\Node\Stmt\Interface_ as InterfaceNode;
+use PhpParser\Node\Stmt\Namespace_ as NamespaceNode;
+use PhpParser\Node\Stmt\Property as PropertyNode;
+use PhpParser\Node\Stmt\Trait_ as TraitNode;
+use PhpParser\Node\Stmt\TraitUse;
 use Roave\BetterReflection\NodeCompiler\CompileNodeToValue;
 use Roave\BetterReflection\NodeCompiler\CompilerContext;
 use Roave\BetterReflection\Reflection\Exception\NotAClassReflection;
@@ -12,17 +24,6 @@ use Roave\BetterReflection\Reflector\Reflector;
 use Roave\BetterReflection\SourceLocator\Located\LocatedSource;
 use Roave\BetterReflection\TypesFinder\FindTypeFromAst;
 use Roave\BetterReflection\Util\GetFirstDocComment;
-use phpDocumentor\Reflection\Types\Object_;
-use PhpParser\Node;
-use PhpParser\Node\Stmt\ClassMethod;
-use PhpParser\Node\Stmt\Namespace_ as NamespaceNode;
-use PhpParser\Node\Stmt\ClassLike as ClassLikeNode;
-use PhpParser\Node\Stmt\Class_ as ClassNode;
-use PhpParser\Node\Stmt\Trait_ as TraitNode;
-use PhpParser\Node\Stmt\Interface_ as InterfaceNode;
-use PhpParser\Node\Stmt\ClassConst as ConstNode;
-use PhpParser\Node\Stmt\Property as PropertyNode;
-use PhpParser\Node\Stmt\TraitUse;
 
 class ReflectionClass implements Reflection, \Reflector
 {
@@ -503,11 +504,11 @@ class ReflectionClass implements Reflection, \Reflector
      * @param string $name
      * @return ReflectionClassConstant|null
      */
-    public function getReflectionConstant(string $name): ?ReflectionClassConstant
+    public function getReflectionConstant(string $name) : ?ReflectionClassConstant
     {
         $constants = $this->getReflectionConstants();
 
-        if (!isset($constants[$name])) {
+        if (!array_key_exists($name, $constants)) {
             return null;
         }
 
@@ -515,34 +516,45 @@ class ReflectionClass implements Reflection, \Reflector
     }
 
     /**
+     * Returns array of constants for this class
+     *
+     * @param array $accumulator
+     * @param Node\Stmt $stmt
+     * @return array
+     */
+    private function processReflectionConstants(array $accumulator, Node\Stmt $stmt) : array
+    {
+        if ($stmt instanceof ClassConst) {
+            $const = ReflectionClassConstant::createFromConstFlagsNameAndValue(
+                $stmt->flags,
+                $stmt->consts[0]->name,
+                (new CompileNodeToValue())->__invoke(
+                    $stmt->consts[0]->value,
+                    new CompilerContext($this->reflector, $this)
+                )
+            );
+            $accumulator[$const->getName()] = $const;
+        }
+
+        return $accumulator;
+    }
+
+    /**
      * Get an array reflection object of the defined constants in this class.
      *
      * @return ReflectionClassConstant[]
      */
-    public function getReflectionConstants(): array
+    public function getReflectionConstants() : array
     {
-        if (null !== $this->cachedReflectionConstants) {
-            return $this->cachedReflectionConstants;
+        if (null === $this->cachedReflectionConstants) {
+            $this->cachedReflectionConstants = array_reduce(
+                $this->node->stmts,
+                [$this, 'processReflectionConstants'],
+                []
+            );
         }
 
-        $constants = [];
-
-        foreach ($this->node->stmts as $stmt) {
-            if ($stmt instanceof ConstNode) {
-                $const = ReflectionClassConstant::createFromConstFlagsNameAndValue(
-                    $stmt->flags,
-                    $stmt->consts[0]->name,
-                    (new CompileNodeToValue())->__invoke(
-                        $stmt->consts[0]->value,
-                        new CompilerContext($this->reflector, $this)
-                    )
-                );
-                $constants[$const->getName()] = $const;
-            }
-        }
-
-        $this->cachedReflectionConstants = $constants;
-        return $constants;
+        return $this->cachedReflectionConstants;
     }
 
     /**
--- a/src/Reflection/ReflectionClass.php
+++ b/src/Reflection/ReflectionClass.php
@@ -504,13 +504,7 @@ class ReflectionClass implements Reflection, \Reflector
      */
     public function getReflectionConstant(string $name) : ?ReflectionClassConstant
     {
-        $constants = $this->getReflectionConstants();
-
-        if (!array_key_exists($name, $constants)) {
-            return null;
-        }
-
-        return $constants[$name];
+        return $this->getReflectionConstants()[$name] ?? null;
     }
 
     /**
@@ -518,7 +512,7 @@ class ReflectionClass implements Reflection, \Reflector
      *
      * @param array $accumulator
      * @param Node\Stmt $stmt
-     * @return array
+     * @return ReflectionClassConstant[]
      */
     private function processReflectionConstants(array $accumulator, Node\Stmt $stmt) : array
     {
@@ -537,15 +531,15 @@ class ReflectionClass implements Reflection, \Reflector
      */
     public function getReflectionConstants() : array
     {
-        if (null === $this->cachedReflectionConstants) {
-            $this->cachedReflectionConstants = array_reduce(
-                $this->node->stmts,
-                [$this, 'processReflectionConstants'],
-                []
-            );
+        if (null !== $this->cachedReflectionConstants) {
+            return $this->cachedReflectionConstants;
         }
 
-        return $this->cachedReflectionConstants;
+        return $this->cachedReflectionConstants = array_reduce(
+            $this->node->stmts,
+            [$this, 'processReflectionConstants'],
+            []
+        );
     }
 
     /**
--- a/src/Reflection/ReflectionClass.php
+++ b/src/Reflection/ReflectionClass.php
@@ -136,14 +136,13 @@ class ReflectionClass implements Reflection, \Reflector
         $buildConstants = function (array $items, $indentLevel = 4) {
             $str = '';
 
-            foreach ($items as $name => $value) {
+            /**
+             * @var string $name
+             * @var ReflectionClassConstant $const
+             */
+            foreach ($items as $name => $const) {
                 $str .= "\n" . str_repeat(' ', $indentLevel);
-                $str .= sprintf(
-                    'Constant [ %s %s ] { %s }',
-                    gettype($value),
-                    $name,
-                    $value
-                );
+                $str .= trim((string)$const);
             }
 
             return $str;
@@ -160,8 +159,8 @@ class ReflectionClass implements Reflection, \Reflector
             $this->getFileName(),
             $this->getStartLine(),
             $this->getEndLine(),
-            count($this->getConstants()),
-            $buildConstants($this->getConstants()),
+            count($this->getReflectionConstants()),
+            $buildConstants($this->getReflectionConstants()),
             count($staticProperties),
             $buildString($staticProperties),
             count($staticMethods),
--- a/src/Reflection/ReflectionClass.php
+++ b/src/Reflection/ReflectionClass.php
@@ -5,7 +5,6 @@ namespace Roave\BetterReflection\Reflection;
 use phpDocumentor\Reflection\Types\Object_;
 use PhpParser\Node;
 use PhpParser\Node\Stmt\Class_ as ClassNode;
-use PhpParser\Node\Stmt\ClassConst;
 use PhpParser\Node\Stmt\ClassConst as ConstNode;
 use PhpParser\Node\Stmt\ClassLike as ClassLikeNode;
 use PhpParser\Node\Stmt\ClassMethod;
@@ -523,7 +522,7 @@ class ReflectionClass implements Reflection, \Reflector
      */
     private function processReflectionConstants(array $accumulator, Node\Stmt $stmt) : array
     {
-        if ($stmt instanceof ClassConst) {
+        if ($stmt instanceof ConstNode) {
             $const = ReflectionClassConstant::createFromNode($this->reflector, $stmt, $this);
             $accumulator[$const->getName()] = $const;
         }
--- a/src/Reflection/ReflectionClass.php
+++ b/src/Reflection/ReflectionClass.php
@@ -525,14 +525,7 @@ class ReflectionClass implements Reflection, \Reflector
     private function processReflectionConstants(array $accumulator, Node\Stmt $stmt) : array
     {
         if ($stmt instanceof ClassConst) {
-            $const = ReflectionClassConstant::createFromConstFlagsNameAndValue(
-                $stmt->flags,
-                $stmt->consts[0]->name,
-                (new CompileNodeToValue())->__invoke(
-                    $stmt->consts[0]->value,
-                    new CompilerContext($this->reflector, $this)
-                )
-            );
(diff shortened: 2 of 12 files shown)


src/Reflection/ReflectionClass.php
```diff
@@ -11,20 +23,10 @@
...
-use PhpParser\Node\Stmt\Class_ as ClassNode;
-use PhpParser\Node\Stmt\Trait_ as TraitNode;
-use PhpParser\Node\Stmt\Interface_ as InterfaceNode;
-use PhpParser\Node\Stmt\ClassConst as ConstNode;
-use PhpParser\Node\Stmt\Property as PropertyNode;
-use PhpParser\Node\Stmt\TraitUse;
 
 class ReflectionClass implements Reflection, \Reflector
 {
+
```
Ocramius: Revert

src/Reflection/ReflectionClass.php
```diff
@@ -490,6 +497,56 @@ public function hasConstant(string $name) : bool
...
+     *
+     * @param string $name
+     * @return ReflectionClassConstant|null
+     */
+    public function getReflectionConstant(string $name): ?ReflectionClassConstant
+    {
+        $constants = $this->getReflectionConstants();
+
+        if (!isset($constants[$name])) {
+            return null;
```
Ocramius: I'd rather have an exception here. `null` makes it a silent failure

src/Reflection/ReflectionClass.php
```diff
@@ -490,6 +497,56 @@ public function hasConstant(string $name) : bool
...
+     * Returns null if not specified.
+     *
+     * @param string $name
+     * @return ReflectionClassConstant|null
+     */
+    public function getReflectionConstant(string $name): ?ReflectionClassConstant
+    {
+        $constants = $this->getReflectionConstants();
+
+        if (!isset($constants[$name])) {
```
Ocramius: `array_key_exists()`

src/Reflection/ReflectionClass.php
```diff
@@ -490,6 +497,56 @@ public function hasConstant(string $name) : bool
...
+     */
+    public function getReflectionConstants(): array
+    {
+        if (null !== $this->cachedReflectionConstants) {
+            return $this->cachedReflectionConstants;
+        }
+
+        $constants = [];
+
+        foreach ($this->node->stmts as $stmt) {
```
Ocramius: Use an array filter here. A private method where you do it would be best

src/Reflection/ReflectionClass.php
```diff
@@ -490,6 +497,56 @@ public function hasConstant(string $name) : bool
...
+                    (new CompileNodeToValue())->__invoke(
+                        $stmt->consts[0]->value,
+                        new CompilerContext($this->reflector, $this)
+                    )
+                );
+                $constants[$const->getName()] = $const;
+            }
+        }
+
+        $this->cachedReflectionConstants = $constants;
```
Ocramius: Can `return` directly on this assignment

src/Reflection/ReflectionClassConstant.php
```diff
@@ -0,0 +1,128 @@
+<?php
+
+namespace Roave\BetterReflection\Reflection;
+
+use PhpParser\Node\Stmt\Class_;
+
+class ReflectionClassConstant implements \Reflector
+{
+
+    private $name;
```
Ocramius: Please add docblocks

src/Reflection/ReflectionClassConstant.php
```diff
@@ -0,0 +1,128 @@
...
+     *
+     * @param int $flags
+     * @param string $name
+     * @param null $value
+     * @return ReflectionClassConstant
+     */
+    public static function createFromConstFlagsNameAndValue(int $flags, string $name, $value = null)
+    {
+        $ref = new self();
+        $ref->flags = $flags === 0 ? Class_::MODIFIER_PUBLIC : $flags;
```
Ocramius: Please assert on the flags validity. Constants only allow visibility changes

src/Reflection/ReflectionClassConstant.php
```diff
@@ -0,0 +1,128 @@
...
+
+    /**
+     * Create a reflection of a class's constant by it's flags, name and value
+     *
+     * @param int $flags
+     * @param string $name
+     * @param null $value
+     * @return ReflectionClassConstant
+     */
+    public static function createFromConstFlagsNameAndValue(int $flags, string $name, $value = null)
```
Ocramius: `$value` shouldn't be optional

src/Reflection/ReflectionClassConstant.php
```diff
@@ -0,0 +1,128 @@
...
+    {
+        return $this->value;
+    }
+
+    /**
+     * Constant is public
+     *
+     * @return bool
+     */
+    public function isPublic()
```
Ocramius: Return hint. Docblock is redundant

src/Reflection/ReflectionClassConstant.php
```diff
@@ -0,0 +1,128 @@
...
+    {
+        return (bool)($this->flags & Class_::MODIFIER_PUBLIC);
+    }
+
+    /**
+     * Cosnstant is private
+     *
+     * @return bool
+     */
+    public function isPrivate()
```
Ocramius: Return hint. Docblock is redundant

src/Reflection/ReflectionClassConstant.php
```diff
@@ -0,0 +1,128 @@
...
+    {
+        return (bool)($this->flags & Class_::MODIFIER_PRIVATE);
+    }
+
+    /**
+     * Constant is protected
+     *
+     * @return bool
+     */
+    public function isProtected()
```
Ocramius: Return hint. Docblock is redundant

src/Reflection/ReflectionClassConstant.php
```diff
@@ -0,0 +1,128 @@
...
+    {
+        return (bool)($this->flags & Class_::MODIFIER_PROTECTED);
+    }
+
+    private function getVisibility()
+    {
+        if ($this->isPublic()) {
+            return '<public>';
+        } elseif ($this->isPrivate()) {
```
Ocramius: `else` not needed

src/Reflection/ReflectionClassConstant.php
```diff
@@ -0,0 +1,128 @@
...
+            return '<private>';
+        } elseif ($this->isProtected()) {
```
Ocramius: Same

src/Reflection/ReflectionClassConstant.php
```diff
@@ -0,0 +1,128 @@
...
+            return '<protected>';
+        }
+        return '<unknown>';
```
Ocramius: Why the brackets?

test/unit/Reflection/ReflectionClassConstantTest.php
```diff
@@ -0,0 +1,48 @@
...
+
+use Roave\BetterReflection\Reflector\ClassReflector;
+use Roave\BetterReflection\SourceLocator\Type\ComposerSourceLocator;
+use Roave\BetterReflectionTest\Fixture\ExampleClass;
+
+class ReflectionClassConstantTest extends \PHPUnit_Framework_TestCase
+{
+    private function getComposerLocator() : ComposerSourceLocator
+    {
+        global $loader;
```
Ocramius: Really smelly...

src/Reflection/ReflectionClassConstant.php
TomasVotruba: Would you be for adding coding standard for these?

Ocramius: If a return type can be specified, it must be specified

Ocramius:
> Note: I did not understand how testReflectionObjectOverridesAllMethodsInReflectionClass works, therefore I missed it.

It's just a test that checks that all method calls go through to an underlying better-reflection class. It's just there for API compatibility: if php-src defines new reflection methods, this test catches them.

src/Reflection/ReflectionClassConstant.php
TomasVotruba: I agree! There is a sniff for that that can check (and complete) that for you.

src/Reflection/ReflectionClass.php
POPSuL: In original php-reflection this method returns false:
https://github.com/php-src/php/blob/master/ext/reflection/php_reflection.c#L4547

Ocramius: Urgh, that's just horrible :-(

test/unit/Reflection/ReflectionClassConstantTest.php
POPSuL: Yep, is really smelly code, but...
https://github.com/Roave/BetterReflection/blob/master/test/unit/Reflection/ReflectionClassTest.php#L44 :(

Ocramius: Can you try `$loader = require __DIR__ . '/../../../lots-of-../../vendor/autoload.php';`?

POPSuL:
Also need:
```
::getModifiers() : int;
::getDeclaringClass() : ReflectionClass
::getDocComment() : ?string
```

I'll add it later

POPSuL:
Huh, I think I'm done.

src/Reflection/ReflectionClass.php
```diff
@@ -490,6 +495,60 @@ public function hasConstant(string $name) : bool
...
+
+    /**
+     * Get an array reflection object of the defined constants in this class.
+     *
+     * @return ReflectionClassConstant[]
+     */
+    public function getReflectionConstants() : array
+    {
+        if (null === $this->cachedReflectionConstants) {
+            $this->cachedReflectionConstants = array_reduce(
```
Ocramius: Just noticed that this code does not consider parent classes and traits - anything that we can do there?

POPSuL: @Ocramius native reflection not provide access for constants inherited from parents. Current behavior is exactly the same as in the native reflection.
```
class A
{
    const A = 1;
}
class B
{
    const B = 2;
}
foreach ([A::class, B::class] as $class) {
    echo 'Native ', $class, ':', PHP_EOL;
    $ref = new ReflectionClass($class);
    foreach ($ref->getReflectionConstants() as $const) {
        echo $const;
    }

    echo 'Better ', $class, ':', PHP_EOL;
    $ref = $reflector->reflect($class);
    foreach ($ref->getReflectionConstants() as $const) {
        echo $const;
    }
}
```
Output:
```
Native A:
Constant [ public integer A ] { 1 }
Better A:
Constant [ public integer A ] { 1 }
Native B:
Constant [ public integer B ] { 2 }
Better B:
Constant [ public integer B ] { 2 }
```

And traits doesn't support constants.

Ocramius: > @Ocramius native reflection not provide access for constants inherited from parents. Current behavior is exactly the same as in the native reflection.

We should provide a "clean" implementation in our own code. For instance, we have [`Roave\BetterReflection\Reflection\ReflectionClass#getImmediateMethods()`](https://github.com/Roave/BetterReflection/blob/8bdd09ce8feadb027a93468e2da5a7f281eba2e5/src/Reflection/ReflectionClass.php#L378) for methods implemented only at the current inheritance level (without considering ancestors), and then we provide a [reflection-compatible API](https://github.com/Roave/BetterReflection/blob/8bdd09ce8feadb027a93468e2da5a7f281eba2e5/src/Reflection/Adapter/ReflectionClass.php#L132-L144) that retrieves all methods, exactly like internal reflection. The same should apply here.

 > And traits doesn't support constants.

Didn't know this, thanks for clarifying!

POPSuL: @Ocramius I added new methods and class to reflection-compat API.
But new features for class constants is not related to this PR, IMHO.

Ocramius: > But new features for class constants is not related to this PR, IMHO.

It's just in line with the package style: it's **Better** Reflection, not just reflection ;-)

TomasVotruba:
@Ocramius As we depend on this PR in ApiGen and after 2 weeks there is probably something missing, what needs to be done here?

Ocramius:
@TomasVotruba I simply didn't see the pushes. A rebase is needed, then another round of reviews

TomasVotruba:
@Ocramius Allright, thanks for update.

@POPSuL Could you look on it please?

POPSuL:
Hmmm... Strange conflict... I check this tomorrow.

POPSuL:
@Ocramius rebased 3 days ago ;)

src/Reflection/ReflectionClassConstant.php
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    /**
+     * Returns the doc comment for this constant
+     *
+     * @return string
+     */
+    public function getDocComment() : string
+    {
+        if (!$this->node->hasAttribute('comments')) {
```
Ocramius: Please see https://github.com/Roave/BetterReflection/pull/284 about the contents of this method

src/Reflection/Adapter/ReflectionClassConstant.php
```diff
@@ -0,0 +1,110 @@
+<?php
+
+namespace Roave\BetterReflection\Reflection\Adapter;
+
+use \ReflectionClassConstant as CoreReflectionClassConstant;
+use Roave\BetterReflection\Reflection\ReflectionClassConstant as BetterReflectionClassConstant;
+
+class ReflectionClassConstant extends CoreReflectionClassConstant
```
Ocramius: :+1:

src/Reflection/Adapter/ReflectionClassConstant.php
```diff
@@ -0,0 +1,110 @@
...
+{
+    private $betterClassConstant;
+
+    public function __construct(BetterReflectionClassConstant $betterClassConstant)
+    {
+        $this->betterClassConstant = $betterClassConstant;
+    }
+
```
Ocramius: Empty line to be removed

src/Reflection/Adapter/ReflectionClassConstant.php
```diff
@@ -0,0 +1,110 @@
+<?php
+
+namespace Roave\BetterReflection\Reflection\Adapter;
+
+use \ReflectionClassConstant as CoreReflectionClassConstant;
+use Roave\BetterReflection\Reflection\ReflectionClassConstant as BetterReflectionClassConstant;
+
+class ReflectionClassConstant extends CoreReflectionClassConstant
+{
+    private $betterClassConstant;
```
Ocramius: Docblock

src/Reflection/ReflectionClass.php
```diff
@@ -491,6 +495,60 @@ public function hasConstant(string $name) : bool
...
+     */
+    public function getReflectionConstant(string $name) : ?ReflectionClassConstant
+    {
+        $constants = $this->getReflectionConstants();
+
+        if (!array_key_exists($name, $constants)) {
+            return null;
+        }
+
+        return $constants[$name];
```
Ocramius: `return $this->getReflectionConstants()[$name] ?? null`

src/Reflection/ReflectionClass.php
```diff
@@ -491,6 +495,60 @@ public function hasConstant(string $name) : bool
...
+    }
+
+    /**
+     * Returns array of constants for this class
+     *
+     * @param array $accumulator
+     * @param Node\Stmt $stmt
+     * @return array
```
Ocramius: Could you be more specific here? `ReflectionClassConstant[]`?

src/Reflection/ReflectionClass.php
```diff
@@ -491,6 +495,60 @@ public function hasConstant(string $name) : bool
...
+     */
+    private function processReflectionConstants(array $accumulator, Node\Stmt $stmt) : array
+    {
+        if ($stmt instanceof ConstNode) {
+            $const = ReflectionClassConstant::createFromNode($this->reflector, $stmt, $this);
```
Ocramius: Inline with the next statement

src/Reflection/ReflectionClass.php
```diff
@@ -491,6 +495,60 @@ public function hasConstant(string $name) : bool
...
+    }
+
+    /**
+     * Get an array reflection object of the defined constants in this class.
+     *
+     * @return ReflectionClassConstant[]
+     */
+    public function getReflectionConstants() : array
+    {
+        if (null === $this->cachedReflectionConstants) {
```
Ocramius: Could you flip this `if/else` around, so the processing becomes the default scenario?

src/Reflection/ReflectionClassConstant.php
```diff
@@ -0,0 +1,202 @@
...
+     */
+    private $valueCached = false;
+
+    /**
+     * @var mixed const value
+     */
+    private $value;
+
+    /**
+     * @var
```
Ocramius: ? :-P

src/Reflection/ReflectionClassConstant.php
```diff
@@ -0,0 +1,202 @@
...
+     */
+    private $reflector;
+    /**
```
Ocramius: Spacing

src/Reflection/ReflectionClassConstant.php
```diff
@@ -0,0 +1,202 @@
...
+
+use PhpParser\Node\Stmt\Class_;
+use PhpParser\Node\Stmt\ClassConst;
+use Roave\BetterReflection\NodeCompiler\CompileNodeToValue;
+use Roave\BetterReflection\NodeCompiler\CompilerContext;
+use Roave\BetterReflection\Reflector\Reflector;
+
+class ReflectionClassConstant implements \Reflector
+{
+
```
Ocramius: Remove this empty line

src/Reflection/ReflectionClassConstant.php
```diff
@@ -0,0 +1,202 @@
...
+    /**
+     * @var bool value of const was cached?
```
Ocramius: Description redundant

src/Reflection/ReflectionClassConstant.php
```diff
@@ -0,0 +1,202 @@
...
+     */
+    private $valueCached = false;
```
Ocramius: Consider renaming to `$valueWasCached` to remove the need for a description

src/Reflection/ReflectionClassConstant.php
```diff
@@ -0,0 +1,202 @@
...
+
+    /**
+     * @var mixed const value
```
Ocramius: Constants are not mixed: they can contain `int|float|array|string|bool|null` - all related docblocks need alignment to this too

src/Reflection/ReflectionClassConstant.php
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    /**
+     * Returns constant value
+     *
+     * @return mixed
+     */
+    public function getValue()
+    {
+        if (false === $this->valueCached) {
```
Ocramius: Consider flipping over the conditional, so the processing comes after an early return

src/Reflection/ReflectionClassConstant.php
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    /**
+     * Returns a bitfield of the access modifiers for this constant
+     *
+     * @return int
+     */
+    public function getModifiers() : int
+    {
+        return $this->node->flags === 0 ? Class_::MODIFIER_PUBLIC : $this->node->flags;
```
Ocramius: Is `0` considered `public` at all times?

Ocramius: Maybe we should store the correct modifier if no modifier is defined (public) instead? How much does that go against the original reflection API?

src/Reflection/ReflectionClassConstant.php
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    /**
+     * Get the declaring class
+     *
+     * @return ReflectionClass
+     */
+    public function getDeclaringClass() : ReflectionClass
+    {
+        return $this->owner;
```
Ocramius: What about non-class constants?

```php
<?php

namespace Foo {
    const BAR = 'baz';
}

namespace {
    var_dump(\Foo\BAR);
}
```

https://3v4l.org/C0hEF

src/Reflection/ReflectionClassConstant.php
```diff
@@ -0,0 +1,202 @@
...
+        if ($this->isPublic()) {
+            return 'public';
+        }
+        if ($this->isPrivate()) {
+            return 'private';
+        }
+        if ($this->isProtected()) {
+            return 'protected';
+        }
+        return '';
```
Ocramius: Should we even consider this scenario?

src/Reflection/ReflectionClassConstant.php
POPSuL: If modifier not provided (just `const A = 'b'`) visibility is public.
Original behavior I'll check tomorrow.

src/Reflection/ReflectionClass.php
POPSuL: Hmmm... Sorry, how?

src/Reflection/Adapter/ReflectionClassConstant.php
POPSuL: Resolved

src/Reflection/ReflectionClassConstant.php
POPSuL: PHP doesn't have reflection for constants.
ReflectionClassConstant can be instantiated only on ReflectionClass and only for constant related to "this" class.

src/Reflection/ReflectionClass.php
TomasVotruba: Is this resolved somehow?

Ocramius: @POPSuL variable not needed, that's what I mean

POPSuL: @Ocramius why not needed? You suggest
```
$accumulator[ReflectionClassConstant::createFromNode($this->reflector, $stmt, $this)->getName()] = ReflectionClassConstant::createFromNode($this->reflector, $stmt, $this);
```
?
Or...
```
$accumulator[$stmt->consts[0]->value] = ReflectionClassConstant::createFromNode($this->reflector, $stmt, $this);
```
?

Ocramius: Oh, dang,  I just noticed that it is indeed used twice. Sorry, that's me misreading stuff :-\

POPSuL:
+1 week... :(

Ocramius:
@POPSuL patience - you're lucky it's not one of the "3 year old PRs" in my backlog ;-)

@asgrim did we decide something about namespaced constants?

```php
namespace Foo {
    const BAR = 'baz';
}
```

asgrim:
@Ocramius this is specifically to do with class constants, so not worried about `define`d or `namespace`d constants here I think.

My onl concern here is if a new API is introduced into PHP that allows reflection of constants (whether specific to class constants like this) or even all "forms" of constants. This API would then conflict with this, so if that were the case, we'd need to release a new major as BC would be broken. Apart from that :+1: from me.

test/unit/Fixture/ExampleClassExport.txt
```diff
@@ -1,9 +1,12 @@
 Class [ <user> class Roave\BetterReflectionTest\Fixture\ExampleClass ] {
-  @@ %s/test/unit/Fixture/ExampleClass.php 7-36
+  @@ %s/test/unit/Fixture/ExampleClass.php 7-43
 
-  - Constants [2] {
-    Constant [ integer MY_CONST_1 ] { 123 }
-    Constant [ integer MY_CONST_2 ] { 234 }
+  - Constants [5] {
+    Constant [ public integer MY_CONST_1 ] { 123 }
```
asgrim: Does this information exist when exporting a reflection in core PHP?

src/Reflection/ReflectionClass.php
```diff
@@ -491,6 +495,54 @@ public function hasConstant(string $name) : bool
...
+     */
+    public function getReflectionConstants() : array
+    {
+        if (null !== $this->cachedReflectionConstants) {
+            return $this->cachedReflectionConstants;
+        }
+
+        return $this->cachedReflectionConstants = array_reduce(
+            $this->node->stmts,
+            [$this, 'processReflectionConstants'],
```
asgrim: This method can just be inlined here I reckon, it's only used the once and I can't see it being re-usable?

test/unit/Fixture/ExampleClassExport.txt
POPSuL: Yes

asgrim: :+1:

src/Reflection/ReflectionClass.php
POPSuL: Done

Ocramius:
> My onl concern here is if a new API is introduced into PHP that allows reflection of constants (whether specific to class constants like this) or even all "forms" of constants. 

We would simply *NOT* release that part as a `Roave\BetterReflection\Reflection\Adapter\*` class (that's the bit that is affected by BC issues if PHP-SRC releases new reflection functionality)

asgrim:
Actually, that's a good point. The Adapter class for this is totally unnecessary at this time :)

Ocramius:
@asgrim can you create an issue for the namespaced (non-class) constant then? Not to be tackled here at this point.

src/Reflection/ReflectionClass.php
```diff
@@ -491,6 +495,44 @@ public function hasConstant(string $name) : bool
...
+    {
+        return $this->getReflectionConstants()[$name] ?? null;
+    }
+
+    /**
+     * Get an array reflection object of the defined constants in this class.
+     *
+     * @return ReflectionClassConstant[]
+     */
+    public function getReflectionConstants() : array
```
Ocramius: Note to self: this block is to be changed at merge time (not a reduce operation)

src/Reflection/ReflectionClassConstant.php
Ocramius: This was further discussed later on. Indeed, further work on non-class constants will not land in this patch :+1:

src/Reflection/ReflectionClassConstant.php
```diff
@@ -0,0 +1,205 @@
...
+            $value
+        );
+    }
+
+    /**
+     * Exports
+     *
+     * @link http://php.net/manual/en/reflector.export.php
+     * @return string
+     * @since 5.0
```
Ocramius: Note to self: fix comment on merge

src/Reflection/ReflectionClassConstant.php
```diff
@@ -0,0 +1,205 @@
...
+        // default visibility always is public
+        return 'public';
+    }
+
+    /**
+     * To string
+     *
+     * @link http://php.net/manual/en/reflector.tostring.php
+     * @return string
+     * @since 5.0
```
Ocramius: Note to self: fix comment on merge

Ocramius:
Just referencing it for confused folks: `ReflectionClassConstant` exists - see https://bugs.php.net/bug.php?id=74261

asgrim:
Ah yes, confused me indeed. Couldn't find documentation, so disregard my earlier comment about removing the Adapter. Does the API definitely match? How do we verify compatibility without checking the src?

Ocramius:
@asgrim there is a `ReflectionClassConstantTest` - I'm verifying it.

Ocramius:
Fixed the nitpicks and merged into `master` at Roave/BetterReflection@44fdc0a0195677cb874370e7f47a678aaa90f0b7, thanks @POPSuL!

PR #288, opened 2017-06-30
added ReflectionProperty::getDefaultValueAsString()

vlastavesely:


asgrim:
I'm not sure why the addition of `ValueDumper` is useful. It's called statically so is in reality no better than just using `var_export` directly in each location... Just introducing unnecessary complexity IMO, which not keen on. Is there a specific reason this is useful please?

vlastavesely:
I can imagine one. In the future you can decide to replace, for example, `var_export`'s `array()` to `[]`. If it is wrapped in the dumper, it is easy, when it would be on many places (I can imagine class constants also), it would be a little bit harder to keep it all the same.

asgrim:
I don't think that's really useful here, I'm afraid :/ the new endpoint is fine but can we just use `var_export` directly

vlastavesely:
So should it be removed?

src/Reflection/ReflectionParameter.php
```diff
@@ -317,7 +318,7 @@ public function getDefaultValue()
      */
     public function getDefaultValueAsString() : string
     {
-        return var_export($this->getDefaultValue(), true);
+        return ValueDumper::dump($this->getDefaultValue());
```
Ocramius: `ValueDumper` is not needed here: `var_export()` is portable, stable, security-checked and does exactly what is needed here

src/Reflection/ReflectionParameter.php
```diff
@@ -317,7 +318,7 @@ public function getDefaultValue()
      */
     public function getDefaultValueAsString() : string
```
Ocramius: @asgrim why do we even have this API? :-\

src/Util/ValueDumper.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+namespace Roave\BetterReflection\Util;
+
+/**
+ * @internal
+ */
+final class ValueDumper
```
Ocramius: This class is not really needed - the static coupling is there anyway

vlastavesely:
requested changes already pushed.

asgrim:
Thank you @vlastavesely :)

PR #289, opened 2017-07-14
Added caching for PhpParser result

POPSuL:
Please, see full description there https://github.com/ApiGen/ApiGen/issues/951#issuecomment-315245749

### DIFF
--- a/src/SourceLocator/Ast/CachedLocator.php
+++ b/src/SourceLocator/Ast/CachedLocator.php
@@ -1,0 +1,56 @@
+<?php
+
+namespace Roave\BetterReflection\SourceLocator\Ast;
+
+use Roave\BetterReflection\Identifier\IdentifierType;
+use Roave\BetterReflection\Reflector\Reflector;
+use Roave\BetterReflection\SourceLocator\Located\LocatedSource;
+
+/**
+ * @internal
+ */
+class CachedLocator extends Locator
+{
+    /**
+     * @var array|string[]
+     */
+    private  $cache = [];
+
+    /**
+     * Get an array of reflections found in some code.
+     *
+     * @param Reflector $reflector
+     * @param LocatedSource $locatedSource
+     * @param IdentifierType $identifierType
+     * @return \Roave\BetterReflection\Reflection\Reflection[]
+     * @throws Exception\ParseToAstFailure
+     */
+    public function findReflectionsOfType(
+        Reflector $reflector,
+        LocatedSource $locatedSource,
+        IdentifierType $identifierType
+    ) : array {
+        try {
+            $id = $locatedSource->getFileName();
+            if ($id === null) {
+                // if source is native stub
+                $id = sha1($locatedSource->getSource());
+            }
+
+            if (!array_key_exists($id, $this->cache)) {
+                $this->cache[$id] = $this->parser->parse($locatedSource->getSource());
+            }
+
+            return $this->findReflectionsInTree->__invoke(
+                $reflector,
+                $this->cache[$id],
+                $identifierType,
+                $locatedSource
+            );
+        } catch (\Exception $exception) {
+            throw Exception\ParseToAstFailure::fromLocatedSource($locatedSource, $exception);
+        } catch (\Throwable $exception) {
+            throw Exception\ParseToAstFailure::fromLocatedSource($locatedSource, $exception);
+        }
+    }
+}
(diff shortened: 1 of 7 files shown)


src/SourceLocator/Ast/Locator.php
```diff
@@ -27,6 +27,11 @@ class Locator
      */
     private $parser;
 
+    /**
+     * @var array|string[]
+     */
+    private static $cache = [];
```
Ocramius: Move this to a `CachedLocator`. Adding a cache, and ESPECIALLY a static one is a no-go here.

src/SourceLocator/Ast/CachedLocator.php
```diff
@@ -0,0 +1,56 @@
...
+                $this->cache[$id] = $this->parser->parse($locatedSource->getSource());
+            }
+
+            return $this->findReflectionsInTree->__invoke(
+                $reflector,
+                $this->cache[$id],
+                $identifierType,
+                $locatedSource
+            );
+        } catch (\Exception $exception) {
```
Ocramius: Not needed

src/SourceLocator/Ast/CachedLocator.php
```diff
@@ -0,0 +1,56 @@
...
+namespace Roave\BetterReflection\SourceLocator\Ast;
+
+use Roave\BetterReflection\Identifier\IdentifierType;
+use Roave\BetterReflection\Reflector\Reflector;
+use Roave\BetterReflection\SourceLocator\Located\LocatedSource;
+
+/**
+ * @internal
+ */
+class CachedLocator extends Locator
```
Ocramius: Let's extract an interface instead of using inheritance

src/SourceLocator/Ast/Locator.php
```diff
@@ -20,12 +20,12 @@ class Locator
     /**
      * @var FindReflectionsInTree
      */
-    private $findReflectionsInTree;
+    protected $findReflectionsInTree;
```
Ocramius: This should be reverted after we extract the interface

src/SourceLocator/Ast/Locator.php
```diff
@@ -20,12 +20,12 @@ class Locator
...
 
     /**
      * @var Parser
      */
-    private $parser;
+    protected $parser;
```
Ocramius: This should be reverted after we extract the interface

src/SourceLocator/Type/ComposerSourceLocator.php
```diff
@@ -21,9 +22,9 @@ class ComposerSourceLocator extends AbstractSourceLocator
      */
     private $classLoader;
 
-    public function __construct(ClassLoader $classLoader)
+    public function __construct(ClassLoader $classLoader, AstLocator $locator = null)
```
Ocramius: We can make the argument required. What is the impact of that?

Ocramius: As on: how much useeland code would need changing?

src/SourceLocator/Ast/Locator.php
burzum: Why does it have to be private?

src/SourceLocator/Ast/Locator.php
burzum: Why does it have to be private?

burzum:
I hope this gets finished and merged any time soon, because it's a great improvement!

Ocramius:
@burzum I think it will not land in `2.0.0`. For now, we focused on features supported, while we may focus on performance for `2.1.0` or `3.0.0` if BC breaks are required for that to happen.

TomasVotruba:
@POPSuL Will you have time to process the notes? Thank you.

TomasVotruba:
I guess this will be replaced by: https://github.com/Roave/BetterReflection/pull/312

Ocramius:
@TomasVotruba I'm still not sure about the approach. I will likely end building a `BetterReflectionKernel` that implements a PSR-11 interface or equivalent, and then is passed around through components.

I definitely need to get rid of the staticness in the lib itself.

TomasVotruba:
@ocramius That would be great. I use this approach in my app for years and I'm surprised no other package is using DI (is that antipattern? :))

Was that some intention on this packages?

I had to register and setup all services manually using Symfony DI. If that would be by default here, that would be awesome.

Ocramius:
Replaced (after much much re-design work) in https://github.com/Roave/BetterReflection/pull/338

TomasVotruba:
Thanks for the update

PR #292, opened 2017-07-22
added ReflectionClassConstant::getValueAsString()

vlastavesely:


src/Reflection/ReflectionClassConstant.php
```diff
@@ -91,6 +91,16 @@ public function getValue()
     }
 
     /**
+     * Get the value represented as a string.
+     *
+     * @return string
+     */
+    public function getValueAsString() : string
```
Ocramius: What's the advantage of this API? Can't the consumer of `getValue()` call a `var_export()` on it manually?

vlastavesely: It could be. But it already was in property reflection (or method parameter... I am not sure).

It is about consistence - the function should be on every reflection where default value is handled or nowhere. Do not you think?

Ocramius: @vlastavesely I see - I think @asgrim missed my comment: https://github.com/Roave/BetterReflection/pull/288#discussion_r125128644

IMO, the API should be completely removed. The consumer decides what to do with the value.

Ocramius:
@asgrim assigning to you for review. In my opinion, the method should be removed from both here and `ReflectionProperty`.

Ref: https://github.com/Roave/BetterReflection/pull/292#discussion_r128891524
Ref: https://github.com/Roave/BetterReflection/pull/288#discussion_r125128644

src/Reflection/ReflectionClassConstant.php
vlastavesely: OK, it is not problem. If you both agreed on the removing it, I will do it.

Ocramius: @vlastavesely thanks for your understanding :-)

Also, this is the right time to do it, as this code is not yet in a stable release :+1:

vlastavesely:
It is also in `ReflectionParameter::getDefaultValueAsString()` and that is where I saw it for first time so I thought it is part of the API for the function is `public`.

Ocramius:
Argh, that is in `1.2.0`, which means BC break to be documented if we remove it: https://github.com/Roave/BetterReflection/commit/a37e53e0c2428720fe1261d9fa6ba9d4eddf7926

Ocramius:
We should at least mark it as `@deprecated`

asgrim:
Yeah, looking back to when this was originally added (a37e53e0c2428720fe1261d9fa6ba9d4eddf7926) this probably shouldn't have been a public method. Let's remove these as indeed a `var_export` is probably more useful anyway.

asgrim:
I'll PR to `@deprecate` the `ReflectionParameter` one, and remove the `ReflectionProperty`.

asgrim:
Added in #293

asgrim:
Closing this therefore! Thank you anyway @vlastavesely, sorry about my misunderstanding!

PR #294, opened 2017-07-24
Support for PHP 7.2's `object` hint

Ocramius:
This patch ensures that the `object` type hint introduced by PHP 7.2 is correctly analysed and treated as a special type.

Also, I imported the current state of php-src reflection tests


src/Reflection/ReflectionParameter.php
```diff
@@ -541,11 +541,17 @@ public function getClass() : ?ReflectionClass
             return null;
         }
 
+        $fqsen = $hint->getFqsen();
+
+        if (! $fqsen) {
```
Ocramius: the cyclomatic complexity:

too-damn-high

test/core/002.phpt
```diff
@@ -50,14 +50,14 @@ var_dump($r->bar);
 ===DONE===
 --EXPECTF--
 ReflectionMethodEx::__construct
-%unicode|string%(26) "ReflectionFunctionAbstract"
-%unicode|string%(7) "getName"
-%unicode|string%(3) "xyz"
+string(26) "ReflectionFunctionAbstract"
```
asgrim: Why changing core `phpt` tests here? They're not strictly relevant, can this be done in a separate PR please if they need updating (which they probably do, to be fair)...

test/unit/Reflection/ReflectionMethodTest.php
```diff
@@ -203,6 +204,25 @@ public function someMethod() {}
...
+    {
+        $php = '<?php
+        namespace A;
+        class Foo {
+            public function someMethod() : object {}
+        }
+        ';
+
+        $returnType = (new ClassReflector(new StringSourceLocator($php)))
+            ->reflect('A\Foo')
```
asgrim: `::class` can be used

test/unit/Reflection/ReflectionParameterTest.php
```diff
@@ -595,4 +595,23 @@ public function testGetClassFromParentTypeHintedProperty() : void
         self::assertInstanceOf(ReflectionClass::class, $hintedClassReflection);
         self::assertSame('stdClass', $hintedClassReflection->getName());
     }
+
+    public function testGetClassFromObjectTypeHintedProperty() : void
+    {
+        $content = '<?php class Foo { public function myMethod(object $param) {} }';
+
+        $parameter = (new ClassReflector(new StringSourceLocator($content)))
+            ->reflect('Foo')
```
asgrim: `::class`

test/core/002.phpt
Ocramius: I was actually looking for 7.2 tests added around `object`, and figured that replacing them as per current state would be much easier

Ocramius:
Merging as per @asgrim's review.

PR #297, opened 2017-08-01
ReflectionParameter: return full name if default value is class' constant

vlastavesely:
In short: `AClass::SOME` !== `BClass::SOME` !== `SOME`

In reflection of this method

```php
public function method($param1 = AClass::SOME, $param2 = BClass::SOME);
```

method `getDefaultValueConstantName()` cannot return just `SOME` for it is not precise description of what constant it really is.

### DIFF
--- a/src/Reflection/ReflectionParameter.php
+++ b/src/Reflection/ReflectionParameter.php
@@ -235,6 +235,9 @@ class ReflectionParameter implements \Reflector
         );
     }
 
+    /**
+     * @throws \LogicException
+     */
     private function findParentClassDeclaringConstant(string $constantName): string
     {
         $class = $this->function->getDeclaringClass();
@@ -243,6 +246,8 @@ class ReflectionParameter implements \Reflector
                 return $class->getName();
             }
         } while ($class = $class->getParentClass());
+
+        throw new LogicException("Failed to find parent class of constant '$constantName'.");
     }
 
     /**
--- a/src/Reflection/ReflectionParameter.php
+++ b/src/Reflection/ReflectionParameter.php
@@ -8,6 +8,7 @@ use Roave\BetterReflection\Reflector\Reflector;
 use Roave\BetterReflection\TypesFinder\FindParameterType;
 use Roave\BetterReflection\TypesFinder\FindTypeFromAst;
 use phpDocumentor\Reflection\Types;
+use phpDocumentor\Reflection\Types\Self_;
 use PhpParser\Node\Param as ParamNode;
 use PhpParser\Node;
 use phpDocumentor\Reflection\Type;
@@ -208,8 +209,14 @@ class ReflectionParameter implements \Reflector
         $defaultValueNode = $this->node->default;
 
         if ($defaultValueNode instanceof Node\Expr\ClassConstFetch) {
+            $className = $this->getRealClassName(implode('\\', $defaultValueNode->class->parts));
+
+            if ($className instanceof Self_) {
+                $className = $this->findParentClassDeclaringConstant($defaultValueNode->name);
+            }
+
             $this->isDefaultValueConstant = true;
-            $this->defaultValueConstantName = $defaultValueNode->name;
+            $this->defaultValueConstantName = ltrim($className, '\\') . '::' . $defaultValueNode->name;
             $this->defaultValueConstantType = self::CONST_TYPE_CLASS;
         }
 
@@ -226,6 +233,16 @@ class ReflectionParameter implements \Reflector
             $defaultValueNode,
             new CompilerContext($this->reflector, $this->getDeclaringClass())
         );
+    }
+
+    private function findParentClassDeclaringConstant(string $constantName): string
+    {
+        $class = $this->function->getDeclaringClass();
+        do {
+            if ($class->hasConstant($constantName)) {
+                return $class->getName();
+            }
+        } while ($class = $class->getParentClass());
     }
 
     /**
@@ -388,12 +405,20 @@ class ReflectionParameter implements \Reflector
      */
     public function getTypeHint() : ?Type
     {
+        return $this->getRealClassName($this->node->type);
+    }
+
+    /**
+     * @param string|null|Node
+     */
+    private function getRealClassName($className) : ?Type
+    {
         $namespaceForType = $this->function instanceof ReflectionMethod
             ? $this->function->getDeclaringClass()->getNamespaceName()
             : $this->function->getNamespaceName();
 
         return (new FindTypeFromAst())->__invoke(
-            $this->node->type,
+            $className,
             $this->function->getLocatedSource(),
             $namespaceForType
         );
--- a/test/unit/Fixture/ClassWithConstantsAsDefaultValues.php
+++ b/test/unit/Fixture/ClassWithConstantsAsDefaultValues.php
@@ -3,6 +3,9 @@
 namespace Roave\BetterReflectionTest\Fixture {
 
     use Roave\BetterReflectionTest\FixtureOther\OtherClass;
+    use Roave\BetterReflectionTest\FixtureOther\OTHER_NAMESPACE_CONST;
+
+    const THIS_NAMESPACE_CONST = 'this_namespace';
 
     class ParentClassWithConstant
     {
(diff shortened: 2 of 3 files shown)


test/unit/Reflection/ReflectionParameterTest.php
```diff
@@ -465,7 +467,7 @@ public function testIsDefaultValueConstantAndGetDefaultValueConstantName() : voi
         $method = $classInfo->getMethod('methodWithConstAsDefault');
         $constDefault = $method->getParameter('constDefault');
         self::assertTrue($constDefault->isDefaultValueConstant());
-        self::assertSame('SOME_CONST', $constDefault->getDefaultValueConstantName());
+        self::assertSame('self::SOME_CONST', $constDefault->getDefaultValueConstantName());
```
Ocramius: This should be resolved to a FQCN, not to `self`. `self` is an alias with no meaning when put out of context.

test/unit/Reflection/ReflectionParameterTest.php
```diff
@@ -479,6 +481,24 @@ public function testIsDefaultValueConstantAndGetDefaultValueConstantName() : voi
...
+    public function testGetDefaultValueConstantNameAcrossClasses() : void
+    {
+        $reflector = new ClassReflector(new SingleFileSourceLocator(
+            __DIR__ . '/../Fixture/ClassWithConstantsAsDefaultValues.php')
+        );
+        $classInfo = $reflector->reflect(ClassWithConstantsAsDefaultValues::class);
+        $method = $classInfo->getMethod('method');
+
+        $param1 = $method->getParameter('param1');
+        self::assertSame('self::MY_CONST', $param1->getDefaultValueConstantName());
```
Ocramius: Same as above

test/unit/Fixture/ClassWithConstantsAsDefaultValues.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+namespace Roave\BetterReflectionTest\Fixture {
+
+	use Roave\BetterReflectionTest\FixtureOther\OtherClass;
+
+	class ClassWithConstantsAsDefaultValues
+	{
+		public const MY_CONST = 'my';
```
Ocramius: Also note: a class constant could be defined multiple times at multiple levels of the inheritance. `self` always resolves to "this class", but a test is needed to check that behavior too.

vlastavesely:
Changes pushed.

src/Reflection/ReflectionParameter.php
```diff
@@ -208,8 +209,20 @@ private function parseDefaultValueNode() : void
         $defaultValueNode = $this->node->default;
 
         if ($defaultValueNode instanceof Node\Expr\ClassConstFetch) {
```
Ocramius: The contents of this block should likely be moved to a private method

src/Reflection/ReflectionParameter.php
```diff
@@ -208,8 +209,20 @@ private function parseDefaultValueNode() : void
...
+            $className = $this->getRealClassName(implode('\\', $defaultValueNode->class->parts));
+
+            if ($className instanceof Self_) {
+                $class = $this->function->getDeclaringClass();
+                do {
+                    if ($class->hasConstant($defaultValueNode->name)) {
+                        $className = $class->getName();
+                        break;
```
Ocramius: This `do/while` should be in a private method, with this particular statement being a `return`

src/Reflection/ReflectionParameter.php
```diff
@@ -388,12 +401,17 @@ public function getPosition() : int
      */
     public function getTypeHint() : ?Type
     {
+        return $this->getRealClassName($this->node->type);
+    }
+
+    private function getRealClassName($className) : ?Type
```
Ocramius: Is `$className` a string? Do type-hint it please

src/Reflection/ReflectionParameter.php
```diff
@@ -208,8 +209,20 @@ private function parseDefaultValueNode() : void
...
+                    if ($class->hasConstant($defaultValueNode->name)) {
+                        $className = $class->getName();
+                        break;
+                    }
+                } while ($class = $class->getParentClass());
+            }
+
             $this->isDefaultValueConstant = true;
-            $this->defaultValueConstantName = $defaultValueNode->name;
+            $this->defaultValueConstantName = ltrim($className, '\\') . '::' . $defaultValueNode->name;
```
Ocramius: When can `$className` contain a leading `\\`? Can you check that with a test?

test/unit/Fixture/ClassWithConstantsAsDefaultValues.php
```diff
@@ -0,0 +1,27 @@
...
+        }
+    }
+}
+
+namespace Roave\BetterReflectionTest\FixtureOther {
+    class OtherClass
+    {
+        public const MY_CONST = 'other';
+    }
+}
```
Ocramius: One scenario needs to be added here:

```php
namespace Roave\BetterReflectionTest\FixtureConst {
    const MY_CONST = 'constant outside a class';
}
```

Ocramius:
@vlastavesely by the way, I know I'm picky and my reviews sound discouraging, but you are doing quite a good job here: keep going! :+1:

src/Reflection/ReflectionParameter.php
vlastavesely: Everytime when working with output of 
```php
return (new FindTypeFromAst())->__invoke(
            $className,
            $this->function->getLocatedSource(),
            $namespaceForType
        );
```

Ocramius: So we are casting to string and prepending `\\` ourselves? That's a bug that we should solve in a separate issue then :-\

src/Reflection/ReflectionParameter.php
vlastavesely: It is enough to add a type-hint into doc comment? It can be a string, NULL, or PhpParser's node.

Ocramius: If there are 3 possible types, then a docblock is OK, although that's a bit of a smell.

test/unit/Fixture/ClassWithConstantsAsDefaultValues.php
vlastavesely: BR does not support reflection of a constant outside a class at the time, does it?

Ocramius: Correct, it doesn't. It would also be acceptable to just write a test that is skipped for now, but we need to be able to resolve to those global or namespaced constants regardless.

src/Reflection/ReflectionParameter.php
vlastavesely: After changes, there was only one condition added so I think it could remain inside `parseDefaultValueNode()`

Ocramius: Seems good then :+1:

src/Reflection/ReflectionParameter.php
```diff
@@ -228,6 +235,16 @@ private function parseDefaultValueNode() : void
...
     }
 
+    private function findParentClassDeclaringConstant(string $constantName): string
+    {
+        $class = $this->function->getDeclaringClass();
+        do {
+            if ($class->hasConstant($constantName)) {
+                return $class->getName();
+            }
+        } while ($class = $class->getParentClass());
```
Ocramius: If the end of the loop is reached, an exception should be thrown, I suppose?

Ocramius: Possibly dead code, but from an semantic perspective, that looks possible

test/unit/Reflection/ReflectionParameterTest.php
```diff
@@ -479,6 +483,41 @@ public function testIsDefaultValueConstantAndGetDefaultValueConstantName() : voi
...
+            __DIR__ . '/../Fixture/ClassWithConstantsAsDefaultValues.php')
+        );
+        $classInfo = $reflector->reflect(ClassWithConstantsAsDefaultValues::class);
+        $method = $classInfo->getMethod('method');
+
+        $param4 = $method->getParameter('param4');
+        self::assertSame('Roave\BetterReflectionTest\Fixture\THIS_NAMESPACE_CONST', $param4->getDefaultValueConstantName());
+
+        $param5 = $method->getParameter('param5');
+        self::assertSame('Roave\BetterReflectionTest\FixtureOther\OTHER_NAMESPACE_CONST', $param5->getDefaultValueConstantName());
```
Ocramius: Oh, it worked \o/ Nice job!

src/Reflection/ReflectionParameter.php
vlastavesely: Yes, I also think it is a dead code but I added it.
With it, it seems better to me 👍

Ocramius:
:ship: \o/

vlastavesely:
Thanks :)

PR #298, opened 2017-08-07
CompileNodeToValue: added support for __CLASS__ magic const

vlastavesely:


src/NodeCompiler/CompileNodeToValue.php
```diff
@@ -273,6 +281,30 @@ private function compileDirConstant(Node\Scalar\MagicConst\Dir $node, CompilerCo
...
+    /**
+     * Compiles magic constant __CLASS__
+     *
+     * @param Node\Scalar\MagicConst\Class_ $node
+     * @param CompilerContext $context
+     * @return string
+     */
+    private function compileClassConstant(Node\Scalar\MagicConst\Class_ $node, CompilerContext $context): string
+    {
+        return $context->hasSelf() ? $context->getSelf()->getName() : '';
```
asgrim: This matches PHP's behaviour (https://3v4l.org/aIRNG), but this feels wrong... should we throw an exception if context doesn't `hasSelf()`? It doesn't make sense to use `__CLASS__` outside of a class context...

src/NodeCompiler/CompilerContext.php
```diff
@@ -41,16 +54,40 @@ public function getSelf() : ?ReflectionClass
...
     public function getReflector() : Reflector
     {
         return $this->reflector;
     }
 
+    /**
+     * @return ReflectionClass|ReflectionFunction
+     * @throws \RuntimeException
+     */
+    public function getContext()
```
asgrim: This is the same as already defined `getSelf()` apart from signature. I see it's done to avoid BC breaks, but it makes me uneasy to duplicate up like this :/ not sure best approach here (this may already be the best approach)

src/NodeCompiler/CompilerContext.php
```diff
@@ -13,26 +14,38 @@ class CompilerContext
...
+     * @param Reflector $reflector
+     * @param ReflectionClass|ReflectionFunction|null $context
+     * @throws \RuntimeException
+     */
+    public function __construct(Reflector $reflector, $context)
     {
         $this->reflector = $reflector;
-        $this->self = $self;
+
+        if ($context instanceof ReflectionClass || $context instanceof ReflectionFunction || $context == null) {
```
asgrim: Hmm, it looks like this PR actually introduces two new features:

 * `__CLASS__` and `__NAMESPACE__` support
 * Allowing context to be a `ReflectionFunction` (which is required to make `__NAMESPACE__` work for a function, as I understand)

Maybe separate them? (especially as the latter could potentially break BC)

test/unit/NodeCompiler/CompileNodeToValueTest.php
```diff
@@ -382,4 +383,50 @@ class Bar {
...
+        function baz($parameter = __CLASS__)
+        {
+        }
+        ';
+
+        $reflector = new ClassReflector(new StringSourceLocator($phpCode));
+        $classInfo = $reflector->reflect('Foo\Bar');
+        self::assertSame('Foo\Bar', $classInfo->getProperty('property')->getDefaultValue());
+
+        $reflector = new FunctionReflector(new StringSourceLocator($phpCode));
```
asgrim: Function support should be in separate tests please

test/unit/NodeCompiler/CompileNodeToValueTest.php
```diff
@@ -382,4 +383,50 @@ class Bar {
...
+    public function testNamespaceMagicConstantAsDefaultValue() : void
+    {
+        $phpCode = '<?php
+        namespace Foo;
+        
+        class Bar {
+            public $property = __NAMESPACE__;
+        }
+        
+        function baz($parameter = __NAMESPACE__)
```
asgrim: Function support should be in separate tests please

src/NodeCompiler/CompileNodeToValue.php
vlastavesely: I think it should match PHP's behaviour no matter how strange it may seem or feel.

src/NodeCompiler/CompilerContext.php
vlastavesely: OK, I will separate it.

asgrim:
Nice, thanks @vlastavesely!

TomasVotruba:
Thank you @vlastavesely  and @asgrim :+1:

PR #299, opened 2017-08-09
TypeFinder: add support for indexed @param annotations

vlastavesely:
Many projects, including Nette Framework, are ommiting variable name for `@param` annotation and are using the annotation's index.

```php
/**
 * @param string
 * @param int
 */
function xxx($str, $int)
{
}
```

Coould we add fallback support for this behaviour?

### DIFF
--- a/src/TypesFinder/FindParameterType.php
+++ b/src/TypesFinder/FindParameterType.php
@@ -8,6 +8,7 @@ use phpDocumentor\Reflection\Types\Context;
 use phpDocumentor\Reflection\Types\ContextFactory;
 use PhpParser\Node\Param as ParamNode;
 use phpDocumentor\Reflection\DocBlock;
+use phpDocumentor\Reflection\DocBlock\Tags\Param;
 use phpDocumentor\Reflection\Type;
 use Roave\BetterReflection\Reflection\ReflectionFunctionAbstract;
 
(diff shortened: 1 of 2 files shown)


Ocramius:
@vlastavesely this seems to be a bad practice:

 > phpDocumentor supports @param tags which omit the name, this is NOT RECOMMENDED but provided for compatibility with existing projects.

https://www.phpdoc.org/docs/latest/references/phpdoc/tags/param.html

I don't think it should be supported here.

vlastavesely:
I consider it to be matter of a coding style in the first place. Why to cut down support for reflection of project with alternative coding style that - I agree with you - may not be in alignment with best practices but it is established for many years?

Any tips how to support this cases?

Ocramius:
> but it is established for many years?

It doesn't really matter: what matters is the set annotation standard. For instance, this library doesn't parse nor understand doctrine annotations, even if they exist as a "de-facto standard" on their own. If this is something that could make it to PSR-5, then it could likely be accepted: I suggest poking that tree first.

Maybe @jaapio can help further here?

jaapio:
In phpdocumentor we don't really care about the order of the `@param` which is the case in the example. When switching the annotations you will change the expected argument type? I'm not sure if PSR-5 describes something about the order of tags. I think it doesn't only the order of
`summary`, `description` followed by none or many tags. 

I think the current phpdocumentor version has some support for this. But I don't know how this is rendered in the documentation. A future version will not support omitting the name. Since it is not part of PSR-5.

Ocramius:
> A future version will not support omitting the name. Since it is not part of PSR-5.

And that clears the case for me.

Closing as `won't fix`.

@vlastavesely please ping me again (or re-open) if this becomes part of PSR-5 at any future point in time.

PR #302, opened 2017-08-16
ReflectionClass::getConstants() and ::getReflectionConstants() should return inherited constants

kukulich:
PHP works the same way: https://3v4l.org/l0NQZ

### DIFF
--- a/src/Reflection/ReflectionClass.php
+++ b/src/Reflection/ReflectionClass.php
@@ -14,8 +14,6 @@ use PhpParser\Node\Stmt\Namespace_ as NamespaceNode;
 use PhpParser\Node\Stmt\Property as PropertyNode;
 use PhpParser\Node\Stmt\Trait_ as TraitNode;
 use PhpParser\Node\Stmt\TraitUse;
-use Roave\BetterReflection\NodeCompiler\CompileNodeToValue;
-use Roave\BetterReflection\NodeCompiler\CompilerContext;
 use Roave\BetterReflection\Reflection\Exception\NotAClassReflection;
 use Roave\BetterReflection\Reflection\Exception\NotAnInterfaceReflection;
 use Roave\BetterReflection\Reflection\Exception\NotAnObject;
@@ -46,11 +44,6 @@ class ReflectionClass implements Reflection, \Reflector
      * @var ClassLikeNode
      */
     private $node;
-
-    /**
-     * @var mixed[]|null
-     */
-    private $cachedConstants;
 
     /**
      * @var ReflectionClassConstant[]|null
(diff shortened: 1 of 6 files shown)


kukulich:
The falling test looks like a bug in PHP itself: https://3v4l.org/1hsD4
I really don't know how to fix it.

src/Reflection/ReflectionClass.php
```diff
@@ -438,30 +431,28 @@ public function hasMethod(string $methodName) : bool
     }
 
     /**
+     * Get an array of the defined constants for this specific class (i.e. do not search
+     * up parent classes etc.)
+     *
+     * @return mixed[]
```
Ocramius: Constants can be `string[]|int[]|float[]|array[]|bool[]|null[]`. Can you remove `mixed[]` and replace it, while we're here?

Ocramius: Also, are they indexed by name? That should be specified in the comment

src/Reflection/ReflectionClass.php
```diff
@@ -542,6 +534,35 @@ function (ReflectionClassConstant $constant) : string {
...
 
     /**
+     * Get an array reflection object of the defined constants in this class.
+     *
+     * @return ReflectionClassConstant[]
+     */
+    public function getReflectionConstants() : array
+    {
+        return array_merge(
+            $this->getImmediateReflectionConstants(),
```
Ocramius: The immediate reflection constants have precedence over the inherited ones: please do add a test with same constant names, but different values throughout the inheritance.

src/Reflection/ReflectionClass.php
kukulich: Done.

src/Reflection/ReflectionClass.php
kukulich: I've improved the the already existed test.

I've also improved the implementation so the keys order in array should be same as in internal implementation: https://3v4l.org/lheaK

Ocramius:
We should probably open an issue on bugs.php.net and skip that class whilst commenting with the bug reference.

src/Reflection/ReflectionClass.php
```diff
@@ -542,6 +535,45 @@ function (ReflectionClassConstant $constant) : string {
...
+     * Get an associative array of the defined constants in this class,
+     * with keys as constant names and values as {@see ReflectionClassConstant} objects.
+     *
+     * @return ReflectionClassConstant[]
+     */
+    public function getReflectionConstants() : array
+    {
+        $allReflectionConstants = array_merge(
+            array_values($this->getImmediateReflectionConstants()),
+            ...array_map(
```
Ocramius: Can't remember if `array_merge()` accepts a single argument. If not, then this code will crash with no ancestors nor interfaces. The simplest solution is to add an empty array as second parameter

Ocramius: Single argument [is fine](https://3v4l.org/PiQ5B), so ignore this comment :-)

kukulich:
@Ocramius 

> You still lack a test that shows what happens with collisions in constant names:

I think it's there. See constants `D` and `F` in https://github.com/Roave/BetterReflection/pull/302/commits/a23d41acf151e74d013fbb0c2829d9949ada15d2#diff-9474cd79fe0ea7163ee0651f53f79b67

I should have used better contant values maybe :)

kukulich:
> We should probably open an issue on bugs.php.net and skip that class whilst commenting with the bug reference.

Done.

src/Reflection/ReflectionClass.php
```diff
@@ -543,6 +536,45 @@ function (ReflectionClassConstant $constant) : string {
...
 
     /**
+     * Get an associative array of the defined constants in this class,
+     * with keys as constant names and values as {@see ReflectionClassConstant} objects.
+     *
+     * @return ReflectionClassConstant[]
+     */
+    public function getReflectionConstants() : array
+    {
+        $allReflectionConstants = array_merge(
```
Ocramius: Note: I tried simplifying this code by removing the merging oddities:

```php
return array_merge(
            ...array_map(
                function (ReflectionClass $ancestor) {
                    return array_filter(
                        $ancestor->getReflectionConstants(),
                        function (ReflectionClassConstant $classConstant) {
                            return ! $classConstant->isPrivate();
                        }
                    );
                },
                array_filter([$this->getParentClass()])
            ),
            ...array_map(
                function (ReflectionClass $interface) {
                    return $interface->getReflectionConstants();
                },
                array_values($this->getInterfaces())
            ),
            ...[$this->getImmediateReflectionConstants()]
        );
```

The sorting does not respect the correct expectations though:

```php
<?php

interface Foo {
    const A = 'a';
}

interface Boo {
    public const B = 'b';
}

class Baz implements Foo {
    public const C = 'c';

    protected const D = 'd';

    private const E = 'e';
}

abstract class Qux extends Baz implements Boo {
    public const F = 'f';

    protected const D = 'dd';
}

abstract class Next extends Qux {
    public const F = 'ff';
}

var_dump((new ReflectionClass(Next::class))->getConstants());
```

Which produces following output, [according to 3v4l.org](https://3v4l.org/oTheI):

```
array(5) { ["F"]=> string(2) "ff" ["D"]=> string(2) "dd" ["C"]=> string(1) "c" ["A"]=> string(1) "a" ["B"]=> string(1) "b" }
```

Therefore this can't be simplified further, I fear :-(

Ocramius:
:ship: thanks @kukulich!

PR #300, opened 2017-08-16
Support for anonymous classes

kukulich:

PR #310, opened 2017-08-18
Improved AST traversing - all levels are inspected now

kukulich:
Needed for https://github.com/Roave/BetterReflection/pull/300

### DIFF
--- a/src/SourceLocator/Ast/FindReflectionsInTree.php
+++ b/src/SourceLocator/Ast/FindReflectionsInTree.php
@@ -3,11 +3,12 @@ declare(strict_types=1);
 
 namespace Roave\BetterReflection\SourceLocator\Ast;
 
+use PhpParser\NodeTraverser;
+use PhpParser\NodeVisitorAbstract;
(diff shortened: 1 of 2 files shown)


Ocramius:
@kukulich do you have an example where the previous approach fails? The fact that this change comes without test additions is worrisome, and a blocker.

kukulich:
@Ocramius As I've mention it's needed for https://github.com/Roave/BetterReflection/pull/300.

There was no bug currently because namespaces had special handling in previous code. So it's improvement not a bug.

I've just added a test for anonymous class.

Ocramius:
@kukulich thanks for adding the test - it demonstrates that the code works as expected, and that's what makes me happy :+1: 

I manually merged after applying nitpicks. Specifically, the by-ref stuff is gone, as well as some if/else cleanup that was long overdue.

kukulich:
@Ocramius Thanks! I especially happy you removed the by-ref stuff. I was very sad I didn't find better solution. It's perfect you find it 👍

PR #316, opened 2017-08-19
PHPStan integration (Level 5 with exclusions) and CI stages

Ocramius:


src/SourceLocator/Type/PhpInternalSourceLocator.php
```diff
@@ -38,9 +38,7 @@ protected function createLocatedSource(Identifier $identifier) : ?LocatedSource
              * @todo this code path looks never used, and disagrees with the contract anyway...?
```
lcobucci: I think that this comment is not relevant anymore

Ocramius: It is still relevant - I only replaced the type, but it is still dead code that will need fixing later on :-\

Ocramius:
:ship:

Issue #317, opened 2017-08-20
Incompatibility between `ReflectionClass#newInstance()` on PHP 7.1 vs PHP 7.2


Ocramius:
As detected in the [latest travis-ci builds](https://travis-ci.org/Roave/BetterReflection/jobs/266398155#L549-L560), the method signature of `ReflectionClass#newInstance()` and `ReflectionObject#newInstance()` changed between PHP 7.1 and PHP 7.2.

We may need to release BetterReflection 2.0.0 with only PHP 7.1 support, and BetterReflection 3.0.0 with only PHP 7.2 support, as there is no way to support both in a sane way.

asgrim:
As this is fixed as per #318, I'm moving the milestone for this to 3.0.0 now.

Ocramius:
Handled in #342

PR #318, opened 2017-08-20
#317 locking onto PHP 7.1 due to reflection API changes that cannot be made compatible with both versions

Ocramius:
See #317


asgrim:
Urgh, nasty, but fair enough :/

Issue #228, opened 2016-12-02
Implement internal parsing of closures


asgrim:
At the moment we rely on a tiny piece of SuperClosure library to reflect on closures. However, it wouldn't be too difficult to bring that in house. Reason being, we're at the mercy of SuperClosure to update dependency on PhpParaer all the time. It's happened twice now and blocked us, and for no good reason.

jeremeamia:
I would actually _encourage_ you to decouple from SuperClosure. It's not something I'm going to put much effort into going forward. It almost seems like BetterReflection would be a logically better home for the part of SuperClosure you are using anyway.

asgrim:
Thanks for feedback @jeremeamia! I mostly used it because "well, it does what we want, and I'm lazy" reasons, but I agree; coupling to a library for just that small feature was a mistake ;) thanks!

asgrim:
Fixed in #321

Issue #152, opened 2015-12-01
ClosureSourceLocator::locateIdentifiersByType - investigate implementation


asgrim:
Currently this is unimplemented and throws an exception. Investigate to see if it's possible to implement.

Ref #37 and #151

asgrim:
Fixed in #321

PR #321, opened 2017-08-20
ClosureSourceLocator improvements

kukulich:
https://github.com/Roave/BetterReflection/issues/228 and https://github.com/Roave/BetterReflection/issues/152 should be solved.


src/SourceLocator/Type/ClosureSourceLocator.php
```diff
@@ -4,100 +4,159 @@
...
+                if (isset($closureNodesDataOnSameLine[1])) {
+                    throw TwoClosuresOnSameLine::create($this->fileName, $this->startLine);
+                }
+
+                return $closureNodesDataOnSameLine[0];
+            }
+        };
+
+        $fileContents = \file_get_contents($fileName);
+        $ast          = (new ParserFactory())->create(ParserFactory::PREFER_PHP7)->parse($fileContents);
```
Ocramius: Note to self: one more parser to inject from an outer scope

Ocramius: Move this call to the constructor, as I may need to move the parser to an injected dependency shortly

src/SourceLocator/Type/ClosureSourceLocator.php
```diff
@@ -4,100 +4,159 @@
...
+
+            /**
+             * @var int
+             */
+            private $startLine;
+
+            /**
+             * @var Node\Expr\Closure[]
+             */
+            private $closureNodesData = [];
```
Ocramius: Just `closureNodes`

kukulich: Done.

src/SourceLocator/Type/ClosureSourceLocator.php
kukulich: Done.

asgrim:
Thanks @kukulich! I've been wanting to get this done for ages but never found the time :grinning:

PR #323, opened 2017-08-22
Method isSubclassOf() of ReflectionClass adapter should check also implemented interfaces to be compatible with internal reflection

kukulich:


Ocramius:
Thanks @kukulich!

PR #324, opened 2017-08-22
Method getFileName() of adapters should return false where no filename to be compatible with internal reflection

kukulich:


Ocramius:
Thanks @kukulich!

PR #328, opened 2017-08-23
Coding standard

kukulich:
I made a lot of commits so it's easy to delete the commits you would not like :)

I tried to be as much as possible compatible with in https://github.com/Roave/BetterReflection/blob/master/.php_cs.php


Ocramius:
@kukulich due to the latest changes, this is in need of a rebase.

kukulich:
@Ocramius Done.

composer.json
```diff
@@ -11,7 +11,9 @@
         "roave/signature": "^1.0"
     },
     "require-dev": {
-        "phpunit/phpunit": "^6.3.0"
+        "phpunit/phpunit": "^6.3.0",
+        "slevomat/coding-standard": "^3.2.1",
+        "squizlabs/php_codesniffer": "^3.0.1"
```
Ocramius: I suggest explicitly using `composer require slevomat/coding-standard:^3.2.1 squizlabs/php_codesniffer:^3.0.1` as a CLI command in the build pipeline step.

src/Reflection/ReflectionType.php
```diff
@@ -90,7 +90,7 @@ public function __toString() : string
             case $this->type instanceof Types\Void_:
                 return 'void';
             default:
-                return \ltrim((string) $this->type, '\\');
+                return \ltrim((string)$this->type, '\\');
```
Ocramius: This should be reverted: spacing after casts is welcome and required.

Ocramius:
See https://github.com/doctrine/coding-standard - I think we have some of this stuff in there.

composer.json
kukulich: It's really difficult to check CS before push then :(

Ocramius: Yeah, it's annoying, but since these are also reflection-ish tools, the dependencies can conflict at some point, making `--dev` installation differ too much from `--no-dev`. I think this is acceptable, as CS can be easily ffixed anyway.

kukulich: Done.

src/Reflection/ReflectionType.php
kukulich: Perfect! Fixed.

kukulich:
https://github.com/doctrine/coding-standard is a base of the CS now.

kukulich:
@Ocramius 

> Another addition that would be very useful is spacing (one empty line) around logical blocks and before return.

I may write the sniff someday :) However first sniff on my todo list is to check that all global functions and constant are used with `\` :)

Ocramius:
@kukulich that would be nice to have indeed, but first this one is going in as-is :+1:

kukulich:
@Ocramius Superb 👍

Issue #327, opened 2017-08-23
Property from Trait should have actual class as declaring class


kukulich:
See https://3v4l.org/sKOMg

Should I fix it only for adapter or should it be fixed in `BetterReflectionClass`?

I will have to add public method `BetterReflectionClass::getReflector()` for fix it only in adapter.

Ocramius:
Not sure about this one, but I feel that the internal reflection API is kinda bad in this case. There should be separate endpoints to get the declaring class and the implementing class, IMO.

Ocramius:
Handled in #330

PR #330, opened 2017-08-25
Adapters compatibility for properties and methods from traits

kukulich:
Fix of https://github.com/Roave/BetterReflection/issues/327.

### DIFF
--- a/test/unit/Reflection/ReflectionPropertyTest.php
+++ b/test/unit/Reflection/ReflectionPropertyTest.php
@@ -3,7 +3,9 @@ declare(strict_types=1);
 
 namespace Roave\BetterReflectionTest\Reflection;
 
+use ClassWithPropertiesAndTraitProperties;
 use Exception;
+use ExtendedClassWithPropertiesAndTraitProperties;
 use InvalidArgumentException;
 use phpDocumentor\Reflection\Types;
 use PhpParser\Node\Stmt\Class_;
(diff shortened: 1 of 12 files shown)


test/unit/Fixture/ClassWithProperties.php
```diff
@@ -0,0 +1,13 @@
+<?php
```
Ocramius: This file name does not reflect the intended usage - this should be `ClassWithPropertiesAndTraitProperties`

test/unit/Fixture/ClassWithMethods.php
```diff
@@ -0,0 +1,17 @@
+<?php
```
Ocramius: This file name does not reflect the intended usage - this should be `ClassWithMethodsAndTraitMethods`

test/unit/Reflection/ReflectionPropertyTest.php
```diff
@@ -362,4 +367,44 @@ public function testGetStartColumnAndEndColumn(string $php, int $startColumn, in
         self::assertEquals($startColumn, $constantReflection->getStartColumn());
         self::assertEquals($endColumn, $constantReflection->getEndColumn());
     }
+
+    public function testGetAst() : void
+    {
+        $php = '<?php
```
Ocramius: Can I suggest using `<<<'PHP'` here?

Ocramius: (maybe a sniff for anything above 3 `"\n"` in a string? Unsure how complex that would be)

kukulich:
> Another test is needed

I'm not sure if I understand it well but I added another test.

test/unit/Fixture/ClassWithProperties.php
kukulich: Done.

test/unit/Fixture/ClassWithMethods.php
kukulich: Done.

test/unit/Reflection/ReflectionPropertyTest.php
kukulich: Done.

Ocramius:
@kukulich the tests you added are exactly what I meant. While they may sound obvious, they remove ambiguity around that API :+1:

Ocramius:
:ship: \o/

Issue #325, opened 2017-08-22
Class with name "Scalar" is not resolved right


kukulich:
Issue reported at https://github.com/phpDocumentor/TypeResolver/issues/43

Ocramius:
Handled in #331

PR #331, opened 2017-08-28
 Used PhpParser\NodeVisitor\NameResolver to simplify code

kukulich:
- The code is more simple
- The code is faster because there's no second parsing in `phpDocumentor\Reflection\Types\ContextFactory`
- It fixes bug https://github.com/Roave/BetterReflection/issues/325
- ~~It fixes support of some missing built-in types eg. `object`, `iterable`, see https://github.com/Roave/BetterReflection/blob/master/src/Reflection/ReflectionType.php#L77~~


test/unit/Reflection/ReflectionTypeTest.php
```diff
@@ -15,49 +13,41 @@ class ReflectionTypeTest extends TestCase
...
+        self::assertSame('string', (string) ReflectionType::createFromType('string', false));
+        self::assertSame('array', (string) ReflectionType::createFromType('array', false));
+        self::assertSame('callable', (string) ReflectionType::createFromType('callable', false));
+        self::assertSame('bool', (string) ReflectionType::createFromType('bool', false));
+        self::assertSame('float', (string) ReflectionType::createFromType('float', false));
+        self::assertSame('void', (string) ReflectionType::createFromType('void', false));
+        self::assertSame('object', (string) ReflectionType::createFromType('object', false));
+        self::assertSame('iterable', (string) ReflectionType::createFromType('iterable', false));
+
+        self::assertSame('Foo\Bar\Baz', (string) ReflectionType::createFromType('\Foo\Bar\Baz', false));
```
Ocramius: Is the prefixing `\` necessary? A test without it is also needed

test/unit/Reflection/ReflectionTypeTest.php
```diff
@@ -15,49 +13,41 @@ class ReflectionTypeTest extends TestCase
...
-        self::assertTrue(ReflectionType::createFromType(new Types\Integer(), false)->isBuiltin());
-        self::assertTrue(ReflectionType::createFromType(new Types\Array_(), false)->isBuiltin());
-        self::assertTrue(ReflectionType::createFromType(new Types\Object_(), false)->isBuiltin());
-        self::assertFalse(ReflectionType::createFromType(new Types\Object_(new Fqsen('\foo')), false)->isBuiltin());
+        self::assertTrue(ReflectionType::createFromType('string', false)->isBuiltin());
+        self::assertTrue(ReflectionType::createFromType('int', false)->isBuiltin());
+        self::assertTrue(ReflectionType::createFromType('array', false)->isBuiltin());
+        self::assertTrue(ReflectionType::createFromType('object', false)->isBuiltin());
+        self::assertTrue(ReflectionType::createFromType('iterable', false)->isBuiltin());
+        self::assertFalse(ReflectionType::createFromType('\foo', false)->isBuiltin());
```
Ocramius: Need to test also without prefixing `\`

src/Reflection/ReflectionType.php
```diff
@@ -22,30 +33,15 @@ private function __construct()
     {
     }
 
-    /**
-     * @param Type $type
-     * @param bool $allowsNull
-     * @return ReflectionType
-     */
-    public static function createFromType(Type $type, bool $allowsNull) : self
+    public static function createFromType(string $type, bool $allowsNull) : self
```
Ocramius: This is a BC break, as this thing was [already in `1.2.0`](https://github.com/Roave/BetterReflection/blob/5d656518b027d3303dee197cec10a7fcb5b712ab/src/Reflection/ReflectionType.php#L29)

src/Reflection/ReflectionType.php
```diff
@@ -3,11 +3,22 @@
 
 namespace Roave\BetterReflection\Reflection;
 
-use phpDocumentor\Reflection\Type;
-use phpDocumentor\Reflection\Types;
-
 class ReflectionType
 {
+    private const BUILT_IN_TYPES = [
+        'int'      => true,
```
Ocramius: Use `null` as value, since boolean involves two possible values, and all we want is existence/non-existence

composer.json
```diff
@@ -4,7 +4,7 @@
     "license": "MIT",
     "require": {
         "php": "7.1.*",
-        "nikic/php-parser": "^3.0.5",
+        "nikic/php-parser": "^3.1.0",
         "phpdocumentor/reflection-docblock": "^3.2.0",
         "phpdocumentor/type-resolver": "^0.4.0",
```
Ocramius: Are these deps still needed then?

src/NodeCompiler/CompileNodeToValue.php
```diff
@@ -122,19 +118,10 @@ private function compileConstFetch(Node\Expr\ConstFetch $constNode)
      */
     private function compileClassConstFetch(Node\Expr\ClassConstFetch $node, CompilerContext $context)
     {
-        $className = \implode('\\', $node->class->parts);
+        $className = $node->class->toString();
```
Ocramius: Overall, this is sane - the `NameVisitor` does indeed simplify things, but we also need to verify that it was run. Any way to do that?

src/TypesFinder/FindTypeFromAst.php
```diff
@@ -1,61 +0,0 @@
...
-
-namespace Roave\BetterReflection\TypesFinder;
-
-use phpDocumentor\Reflection\Type;
-use phpDocumentor\Reflection\Types\ContextFactory;
-use PhpParser\Node\Name;
-use PhpParser\Node\NullableType;
-use Roave\BetterReflection\SourceLocator\Located\LocatedSource;
-
-class FindTypeFromAst
```
Ocramius: [`FindTypesFromAst` is present in `1.2.0`](https://github.com/Roave/BetterReflection/blob/5d656518b027d3303dee197cec10a7fcb5b712ab/src/TypesFinder/FindTypeFromAst.php), so this is a BC break

src/Reflection/ReflectionParameter.php
```diff
@@ -415,35 +416,6 @@ public function getPosition() : int
...
 
     /**
-     * Get the type hint declared for the parameter. This is the real type hint
-     * for the parameter, e.g. `method(closure $someFunc)` defined by the
-     * method itself, and is separate from the DocBlock type hints.
-     *
-     * @see getDocBlockTypes()
-     * @return Type|null
-     */
-    public function getTypeHint() : ?Type
```
Ocramius: Possible BC break

src/Reflection/ReflectionParameter.php
```diff
@@ -469,19 +447,17 @@ public function getType() : ?ReflectionType
...
     /**
      * Set the parameter type declaration.
      *
-     * You must use the phpDocumentor reflection type classes as the parameter.
-     *
-     * @param Type $newParameterType
+     * @param string $newParameterType
      */
-    public function setType(Type $newParameterType) : void
+    public function setType(string $newParameterType) : void
```
Ocramius: BC break

src/Reflection/ReflectionType.php
```diff
@@ -22,30 +33,15 @@ private function __construct()
...
         $reflectionType->allowsNull = $allowsNull;
         return $reflectionType;
     }
 
     /**
-     * Get a PhpDocumentor type object for this type
-     *
-     * @return Type
-     */
-    public function getTypeObject() : Type
```
Ocramius: Bc break

composer.json
kukulich: @Ocramius Yes, there are needed for https://github.com/Roave/BetterReflection/blob/master/src/Reflection/ReflectionFunctionAbstract.php#L427 etc.

src/NodeCompiler/CompileNodeToValue.php
kukulich: @Ocramius I've been thinking about this all day and did not find any way to verify it was run :(

Ocramius: That's alright: I don't expect random folks to jump in and use custom AST nodes with this.

Ocramius: Just needs documentation

Ocramius:
@kukulich the BC breaks are acceptable, but they need to be documented

kukulich:
@Ocramius Just pushed some documentation changes (and fixes). Only documentation of `NameVisitor` is not done because I'm not sure where to put it.

Ocramius:
> Only documentation of `NameVisitor` is not done because I'm not sure where to put it.

@kukulich anything that accepts a `Node` that requires to be passed in as already processed by the `NameVisitor` needs at least a clarification on the parameter hint

kukulich:
@Ocramius Ok, I hope it's ok now.

src/Reflection/ReflectionFunctionAbstract.php
```diff
@@ -442,11 +442,14 @@ public function getReturnType() : ?ReflectionType
             return null;
         }
 
+        $allowsNull = false;
+
         if ($returnType instanceof NullableType) {
             $returnType = $returnType->type;
+            $allowsNull = true;
```
Ocramius: This should be an early return: overwriting variables is generally bad.

Ocramius:
@kukulich the various BC breaks still need documenting in `README.md`

kukulich:
@Ocramius This is not enough?

```
* `FindTypeFromAst` was removed
* `ReflectionParameter::getTypeHint()` was removed
* `ReflectionType::getTypeObject()` was removed
* `ReflectionType::createFromType()` now requires `string` for `$type` parameter
```

Ocramius:
@kukulich totally missed it, sorry! Squashed commits are not my forte ;-)

src/Reflection/ReflectionFunctionAbstract.php
kukulich: Done.

Ocramius:
:ship:

Thanks @kukulich!

asgrim:
@kukulich @Ocramius Hmm, what's the reasoning for removing `FindTypeFromAst`? Pretty sure it's used by some consumers of this library.

asgrim:
Specifically, my concern is that there is no documented upgrade path for consumers of that type resolver.

asgrim:
Actually, created issue #347 to discuss this further, please can you respond there :)

PR #333, opened 2017-08-29
`PhpParser\BuilderFactory` instead of `Zend\Code\Generator\ClassGenerator` in the `SourceStubber`

kukulich:
There's already PhpParser so we can use its `BuilderFactory` to remove dependency on `zendframework/zend-code`.

It should be probably a little faster too because this code has been removed: https://github.com/Roave/BetterReflection/blob/master/src/SourceLocator/Reflection/SourceStubber.php#L51

### DIFF
--- a/src/SourceLocator/Reflection/SourceStubber.php
+++ b/src/SourceLocator/Reflection/SourceStubber.php
@@ -3,13 +3,30 @@ declare(strict_types=1);
 
 namespace Roave\BetterReflection\SourceLocator\Reflection;
 
-use PhpParser\Node\Stmt\Class_;
-use PhpParser\Node\Stmt\Interface_;
-use PhpParser\Node\Stmt\Trait_;
+use PhpParser\Builder\Class_;
+use PhpParser\Builder\Declaration;
+use PhpParser\Builder\Interface_;
+use PhpParser\Builder\Method;
+use PhpParser\Builder\Property;
+use PhpParser\Builder\Trait_;
+use PhpParser\BuilderAbstract;
+use PhpParser\BuilderFactory;
+use PhpParser\Comment\Doc;
+use PhpParser\Node\Const_;
+use PhpParser\Node\Name;
+use PhpParser\Node\Name\FullyQualified;
+use PhpParser\Node\NullableType;
+use PhpParser\Node\Stmt\Class_ as ClassNode;
+use PhpParser\Node\Stmt\ClassConst;
+use PhpParser\Node\Stmt\TraitUse;
+use PhpParser\Node\Stmt\TraitUseAdaptation;
+use PhpParser\NodeAbstract;
 use PhpParser\PrettyPrinter\Standard;
-use Roave\BetterReflection\SourceLocator\Ast\PhpParserFactory;
-use Zend\Code\Generator\ClassGenerator;
-use Zend\Code\Reflection\ClassReflection;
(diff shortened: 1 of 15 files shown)


src/SourceLocator/Reflection/SourceStubber.php
```diff
@@ -30,57 +43,238 @@
...
-     * @param ClassReflection $reflection
-     *
-     * @return string
-     */
-    public function __invoke(ClassReflection $reflection) : string
+    public function __invoke(CoreReflectionClass $classReflection) : string
     {
-        $stubCode    = ClassGenerator::fromReflection($reflection)->generate();
-        $isInterface = $reflection->isInterface();
+        if ($classReflection->isTrait()) {
```
Ocramius: Please move this triplet of `if/else` to a private method

src/SourceLocator/Reflection/SourceStubber.php
```diff
@@ -30,57 +43,238 @@
...
+            $classNode = $this->builderFactory->trait($classReflection->getShortName());
+        } elseif ($classReflection->isInterface()) {
+            $classNode = $this->builderFactory->interface($classReflection->getShortName());
+        } else {
+            $classNode = $this->builderFactory->class($classReflection->getShortName());
+        }
+
+        if ( ! $classReflection->isInterface() && $classReflection->isAbstract()) {
```
Ocramius: Move to a private method `addClassModifiers()`

src/SourceLocator/Reflection/SourceStubber.php
```diff
@@ -30,57 +43,238 @@
...
+            }
+            foreach ($classReflection->getTraits() as $trait) {
+                if ($trait->hasProperty($propertyReflection->getName())) {
+                    continue 2;
+                }
+            }
+
+            $propertyNode = $this->builderFactory->property($propertyReflection->getName());
+
+            if ($propertyReflection->isStatic()) {
```
Ocramius: Move property visibility setting to a private method

src/SourceLocator/Reflection/SourceStubber.php
```diff
@@ -30,57 +43,238 @@
...
+                continue;
+            }
+            foreach ($classReflection->getTraits() as $trait) {
+                if ($trait->hasMethod($methodReflection->getName())) {
+                    continue 2;
+                }
+            }
+
+            $methodNode = $this->builderFactory->method($methodReflection->getName());
+            if ($methodReflection->isFinal()) {
```
Ocramius: Move method flags to a private method and use early return statements as much as possible

kukulich:
@Ocramius I hope I fixed everything.

kukulich:
@Ocramius the test is already there in https://github.com/Roave/BetterReflection/blob/master/test/unit/SourceLocator/Type/PhpInternalSourceLocatorTest.php

Should I move it to SourceStubberTest?

Ocramius:
> @Ocramius the test is already there in https://github.com/Roave/BetterReflection/blob/master/test/unit/SourceLocator/Type/PhpInternalSourceLocatorTest.php

Ah yep, nothing to do then. I'll just grab this patch and split up a few of the complex methods in it, then merge :+1:

kukulich:
Ok, thank you 👍

Ocramius:
:ship:

Issue #182, opened 2016-03-16
[BC Break] Separate AST modification out


asgrim:
Features like #180 and #141 are methods that allow modification of the reflection (via manipulating the AST).

For 2.0 we should change the API so that the modification stuff is separated.

asgrim:
Suggestion:

```php
<?php

// In current BR:
$reflectionMethod = $reflectionClass->getMethod('myMethod');
$reflectionMethod->removeReturnType();

// In 2.0
$newReflectionMethod = (new \BetterReflection\Mutation\RemoveReturnType())($reflectionMethod);
```

Ocramius:
Partially implemented in #339. Closing here, as further discussion should be continued on the PR.

PR #337, opened 2017-08-30
BetterReflection entry-point class

Ocramius:
This patch introduces an "entry point" that can be used to get started with `BetterReflection`.

The usage is as following:

```php
$class = (new \Roave\BetterReflection\Configuration)->classReflector()->reflect(ClassName::class)
```

This also allows us to have clean DI on each of the internal components, since the orchestration is organised by this class.

I will provide detailed info about the BC breaks and decisions to be taken on this below

PR #339, opened 2017-08-30
Mutators and mutations

kukulich:
Implemented https://github.com/Roave/BetterReflection/issues/182

I hope I didn't miss anything and documented all BC breaks.

### DIFF
--- a/docs/features.md
+++ b/docs/features.md
@@ -55,6 +55,6 @@ $functionInfo = (new RemoveFunctionReturnType())($functionInfo);
 You can do similar things with parameter types also:
 
 ```
-$parameterInfo->setType('int');
+(new SetParameterType())($parameterInfo, 'int');
 (new RemoveParameterType())($parameterInfo);
 ```
--- a/docs/features.md
+++ b/docs/features.md
@@ -49,7 +49,7 @@ if (count($returnTypes) === 1) {
 }
 
 // Remove the return type declaration
-$functionInfo->removeReturnType();
+$functionInfo = (new RemoveFunctionReturnType())($functionInfo);
 ```
 
 You can do similar things with parameter types also:
--- a/docs/features.md
+++ b/docs/features.md
@@ -40,12 +40,12 @@ declaration to that defined in the PHP DocBlock.
 
 ```
 // Change a function to ensure it returns an integer
-$functionInfo->setReturnType('int');
+$functionInfo = (new SetFunctionReturnType())($functionInfo, 'int');
 
 // If there is only one type defined in the DocBlock, set it as the return type
 $returnTypes = $functionInfo->getDocBlockReturnTypes();
 if (count($returnTypes) === 1) {
-    $functionInfo->setReturnType((string) $returnTypes[0]);
+    $functionInfo = (new SetFunctionReturnType())($functionInfo, (string) $returnTypes[0], false);
 }
 
 // Remove the return type declaration
--- a/docs/features.md
+++ b/docs/features.md
@@ -56,5 +56,5 @@ You can do similar things with parameter types also:
 
 ```
 $parameterInfo->setType('int');
-$parameterInfo->removeType();
+(new RemoveParameterType())($parameterInfo);
 ```
--- a/docs/features.md
+++ b/docs/features.md
@@ -38,7 +38,14 @@ written for PHP 7 work in PHP 5 for example, and setting new types might
 do the opposite. For instance, you might want to set the PHP 7 return type
 declaration to that defined in the PHP DocBlock.
 
-```
+```php
+<?php
+
+use Roave\BetterReflection\Reflection\Mutation\RemoveFunctionReturnType;
+use Roave\BetterReflection\Reflection\Mutation\RemoveParameterType;
+use Roave\BetterReflection\Reflection\Mutation\SetFunctionReturnType;
+use Roave\BetterReflection\Reflection\Mutation\SetParameterType;
+
 // Change a function to ensure it returns an integer
 $functionInfo = (new SetFunctionReturnType())($functionInfo, 'int');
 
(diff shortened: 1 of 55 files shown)


kukulich:
Hmm, I always create new instance  however I should probably change the instance directly to enable monkey patching. Am I right?

docs/features.md
```diff
@@ -40,16 +40,16 @@ declaration to that defined in the PHP DocBlock.
 
 ```
 // Change a function to ensure it returns an integer
-$functionInfo->setReturnType('int');
+$functionInfo = (new SetFunctionReturnType())($functionInfo, 'int');
```
Ocramius: Please import these right above, since there's no auto-completion in the docs

docs/features.md
```diff
@@ -40,16 +40,16 @@ declaration to that defined in the PHP DocBlock.
...
 
 // If there is only one type defined in the DocBlock, set it as the return type
 $returnTypes = $functionInfo->getDocBlockReturnTypes();
 if (count($returnTypes) === 1) {
-    $functionInfo->setReturnType((string) $returnTypes[0]);
+    $functionInfo = (new SetFunctionReturnType())($functionInfo, (string) $returnTypes[0], false);
```
Ocramius: Please import these right above, since there's no auto-completion in the docs

docs/features.md
```diff
@@ -40,16 +40,16 @@ declaration to that defined in the PHP DocBlock.
...
 }
 
 // Remove the return type declaration
-$functionInfo->removeReturnType();
+$functionInfo = (new RemoveFunctionReturnType())($functionInfo);
```
Ocramius: Please import these right above, since there's no auto-completion in the docs

src/Reflection/Mutation/AddClassMethod.php
```diff
@@ -0,0 +1,29 @@
...
+class AddClassMethod
+{
+    /**
+     * @var ReflectionClassMutator
+     */
+    private $mutator;
+
+    public function __construct()
+    {
+        $this->mutator = new ReflectionClassMutator();
```
Ocramius: Don't really like this lack of constructor injection: it has bitten us before, so I'd prefer to have an entry point like in https://github.com/Roave/BetterReflection/pull/337

src/Reflection/Mutation/AddClassProperty.php
```diff
@@ -0,0 +1,53 @@
...
+
+    public function __construct()
+    {
+        $this->mutator = new ReflectionClassMutator();
+    }
+
+    public function __invoke(
+        ReflectionClass $classReflection,
+        string $propertyName,
+        int $visibility = CoreReflectionProperty::IS_PUBLIC,
```
Ocramius: Would suggest removing default values overall

src/Reflection/Mutation/AddClassProperty.php
```diff
@@ -0,0 +1,53 @@
...
+        bool $static = false
```
Ocramius: Would suggest removing default values overall

src/Reflection/Mutation/AddClassProperty.php
```diff
@@ -0,0 +1,53 @@
...
+    ) : ReflectionClass {
+        $type = 0;
+        switch ($visibility) {
+            case CoreReflectionProperty::IS_PRIVATE:
+                $type |= Class_::MODIFIER_PRIVATE;
+                break;
+            case CoreReflectionProperty::IS_PROTECTED:
+                $type |= Class_::MODIFIER_PROTECTED;
+                break;
+            default:
```
Ocramius: If the value is out of range, an exception should be thrown

src/Reflection/Mutation/AddClassMethod.php
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    public function __invoke(ReflectionClass $classReflection, string $methodName) : ReflectionClass
```
Ocramius: No visibility? No parameters? What's going on here?

Ocramius (changes requested):
Overall, I really like the patch and the idea, but I don't know if we should pursue it here, or split it out to a separate library. Thoughts, @asgrim?

src/Reflection/Mutation/AddClassMethod.php
kukulich: I implemented the same version: https://github.com/Roave/BetterReflection/blob/master/src/Reflection/ReflectionClass.php#L1407

I think the idea was that you should you other methods (now mutations) to add parameters, body etc.

However it would be good to add modifiers here at least.

src/Reflection/Mutation/AddClassProperty.php
kukulich: Perfect, I was thinking about it but I implemented the same way that already exist at the end.

src/Reflection/Mutation/AddClassMethod.php
kukulich: I marked the mutators as internal but ok :)

kukulich:
Everything fixed now.

Ocramius:
@kukulich I think this patch should be deferred for now, as the primary focus of the library is not monkey-patching.

I'm moving the milestone, but please don't work on it further until we get `2.0.0` out, as it will only be wasted rebase time :-)

kukulich:
@Ocramius Ok, I have some more PR in progress so now problem to postpone this :)

asgrim:
Nice! Been wanting to get stuck into this for a while. I do like the patch, but the suggestion of a separate library is interesting. One thing I'd point out, is that in two of my talks, I basically pitch BR as a way to do monkey patching, but separating out that "mutation" component would actually remove some stuff. It can still use BR at the end of the day. Maybe I can set up a new repo `Roave\BetterReflectionMutation` or something? Discussion should continue, but I agree with not including this in `2.0.0` at least as it's quite a huge BC ;)

Ocramius:
Pushing down the milestone.

I'd also add that all the caching, reflection & co really only work if the reflections are immutable, so we need to add some careful testing around the mutation and mutability of the original symbols.

asgrim:

Issue #20, opened 2015-07-02
Validate identifier name in Identifier value object


asgrim:
Reported by @Ocramius in #3 ([here](https://github.com/Roave/BetterReflection/pull/3#discussion_r33802573))

> Consider validating `$name` against the regex `'/([a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]+)(\\\\[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]+)*/'`, as that's what a valid class name can refer to.
> 
> Only one segment for functions/methods.
> 
> See https://github.com/Ocramius/ProxyManager/blob/7badedb5f8afaaccd496a096daadb19d7bf2b374/tests/ProxyManagerTest/Inflector/ClassNameInflectorTest.php#L127

Ocramius:
Handled in #341

PR #341, opened 2017-08-30
Check if identifier name is valid

kukulich:
Implemented https://github.com/Roave/BetterReflection/issues/20

I've slightly modifed the regexp:

Before: `/([a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]+)(\\\\[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]+)*/`
After:  `'/([a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*)(\\\\[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*)*/'`

The first one did not support names with only one letter.

### DIFF
--- a/src/Identifier/Identifier.php
+++ b/src/Identifier/Identifier.php
@@ -3,8 +3,16 @@ declare(strict_types=1);
 
 namespace Roave\BetterReflection\Identifier;
 
+use Roave\BetterReflection\Identifier\Exception\InvalidIdentifierName;
+use Roave\BetterReflection\Reflection\ReflectionClass;
+use Roave\BetterReflection\Reflection\ReflectionFunctionAbstract;
+
 class Identifier
 {
+    public const WILDCARD = '*';
+
+    private const VALID_NAME_REGEXP = '/([a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*)(\\\\[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*)*/';
+
     /**
      * @var string
      */
@@ -15,13 +23,31 @@ class Identifier
      */
     private $type;
 
+    /**
+     * @param string $name
+     * @param IdentifierType $type
+     *
+     * @throws InvalidIdentifierName
+     */
     public function __construct(string $name, IdentifierType $type)
     {
         $this->type = $type;
 
+        if (self::WILDCARD === $name
+            || ReflectionFunctionAbstract::CLOSURE_NAME === $name
+            || 0 === \strpos($name, ReflectionClass::ANONYMOUS_CLASS_NAME_PREFIX)
+        ) {
+            $this->name = $name;
+            return;
+        }
+
         $name = \ltrim($name, '\\');
-        // @todo validate the name somehow (see issue #20)
-        $this->name = (string) $name;
+
+        if ( ! \preg_match(self::VALID_NAME_REGEXP, $name)) {
+            throw InvalidIdentifierName::fromInvalidName($name);
+        }
+
+        $this->name = $name;
     }
 
     public function getName() : string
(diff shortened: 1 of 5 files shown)


src/Identifier/Identifier.php
```diff
@@ -15,13 +21,31 @@ class Identifier
...
+        ) {
+            $this->name = $name;
+            return;
+        }
+
         $name = \ltrim($name, '\\');
-        // @todo validate the name somehow (see issue #20)
-        $this->name = (string) $name;
+
+        if ( ! \preg_match('/([a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*)(\\\\[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*)*/', $name)) {
```
Ocramius: Private constant plz :-)

Ocramius: This pattern makes `\\` valid - At least the first block should be a `+` instead of a `*`

test/unit/Identifier/IdentifierTest.php
```diff
@@ -44,4 +47,28 @@ public function testIsTypesForFunction() : void
...
+
+    public function testGetNameForWildcard() : void
+    {
+        $identifier = new Identifier(Identifier::WILDCARD, new IdentifierType(IdentifierType::IDENTIFIER_CLASS));
+        self::assertSame(Identifier::WILDCARD, $identifier->getName());
+    }
+
+    public function testThrowExceptionForInvalidName() : void
+    {
+        self::expectException(InvalidIdentifierName::class);
```
Ocramius: I think this is an instance method

test/unit/Identifier/IdentifierTest.php
```diff
@@ -44,4 +47,28 @@ public function testIsTypesForFunction() : void
...
+        new Identifier('1234567890', new IdentifierType(IdentifierType::IDENTIFIER_CLASS));
```
Ocramius: Move the invalid data to a data provider, try a few invalid strings

src/Identifier/Identifier.php
kukulich: No, there's still `[a-zA-Z_\x7f-\xff]` at the beginning.

test/unit/Identifier/IdentifierTest.php
kukulich: Done.

test/unit/Identifier/IdentifierTest.php
kukulich: Done.

src/Identifier/Identifier.php
Ocramius: Dang, I need some sleep @_@

test/unit/Identifier/IdentifierTest.php
```diff
@@ -44,4 +47,73 @@ public function testIsTypesForFunction() : void
...
+    {
+        $identifier = new Identifier(Identifier::WILDCARD, new IdentifierType(IdentifierType::IDENTIFIER_CLASS));
+        self::assertSame(Identifier::WILDCARD, $identifier->getName());
+    }
+
+    public function validNamesProvider() : array
+    {
+        return [
+            ['Foo', 'Foo'],
+            ['\Foo', 'Foo'],
```
Ocramius: Should we even accept `\\` as a leading character?

I think it would be nice to start educating users about this, but it's a possible BC break.

Thoughts?

kukulich: The leading `\\` is accepted, there's `ltrim` already.

Ocramius: Yes, I was wondering whether we should tell users not to use it.

In strings, we are always talking about FQCN, so `\\` is redundant, and it can cause harm if not filtered.

We do it here, but the user should likely not keep passing it around that way in their own domain either.

kukulich: Or do you think to accept only names with leading `\\`? It would not be compatible with internal reflection and other libraries.

Ocramius: > It would not be compatible with internal reflection and other libraries.

That is a fair point, and I think it clears the discussion point.

Ocramius:
:ship:

PR #343, opened 2017-08-31
Implemented ReflectionProperty::setAccessible(), getValue() and setValue() and ReflectionClass::getStaticProperties()

kukulich:


src/Reflection/ReflectionClass.php
```diff
@@ -1287,36 +1289,28 @@ public function __clone()
...
         if ( ! $property || ! $property->isStatic()) {
-            throw new PropertyDoesNotExist('Property does not exist on class or is not static');
-        }
-
-        // PHP behaviour is to simply say "property does not exist" if accessing
-        // protected or private values. Here we be a little more explicit in
-        // reasoning...
-        if ( ! $property->isPublic()) {
-            throw new PropertyNotPublic('Property is not public');
+            throw PropertyDoesNotExistOrIsNotStatic::fromName($propertyName);
```
Ocramius: You know if it does not exist or if it is static, so throwing a `PropertyDoesNotExistOrIsNotStatic` feels wrong.

A better way to handle this is having `PropertyDoesNotExistOrIsNotStatic` be an abstract class, and then have two subtypes for the actual failure: that would make it much easier to deal with for a consumer, and would read more sensibly.

src/Reflection/Adapter/ReflectionClass.php
```diff
@@ -400,23 +403,36 @@ public function isSubclassOf($class)
...
     public function getStaticPropertyValue($name, $default = null)
     {
-        throw new Exception\NotImplemented('Not implemented');
+        try {
+            return $this->betterReflectionClass->getStaticPropertyValue($name);
+        } catch (PropertyDoesNotExistOrIsNotStatic $e) {
+            if (2 === \func_num_args()) {
+                return $default;
+            }
+            throw new CoreReflectionException($e->getMessage());
```
Ocramius: Spacing please: spaces around logical blocks :-)

src/Reflection/Adapter/ReflectionClass.php
```diff
@@ -400,23 +403,36 @@ public function isSubclassOf($class)
...
+        } catch (Throwable $e) {
```
Ocramius: When can this happen?

src/Reflection/Adapter/ReflectionClass.php
```diff
@@ -400,23 +403,36 @@ public function isSubclassOf($class)
...
+            throw new CoreReflectionException($e->getMessage());
```
Ocramius: Shouldn't we populate the previous exception?

src/Reflection/Adapter/ReflectionClass.php
Ocramius: Shouldn't we populate the previous exception?

src/Reflection/Adapter/ReflectionClass.php
```diff
@@ -400,23 +403,36 @@ public function isSubclassOf($class)
...
     /**
      * {@inheritDoc}
      */
     public function setStaticPropertyValue($name, $value)
     {
-        throw new Exception\NotImplemented('Not implemented');
+        try {
+            $this->betterReflectionClass->setStaticPropertyValue($name, $value);
+        } catch (Throwable $e) {
+            throw new CoreReflectionException($e->getMessage());
```
Ocramius: Shouldn't we populate the previous exception?

src/Reflection/Adapter/ReflectionObject.php
```diff
@@ -370,23 +373,36 @@ public function isSubclassOf($class)
...
     public function getStaticPropertyValue($name, $default = null)
     {
-        throw new Exception\NotImplemented('Not implemented');
+        try {
+            return $this->betterReflectionObject->getStaticPropertyValue($name);
+        } catch (PropertyDoesNotExistOrIsNotStatic $e) {
+            if (2 === \func_num_args()) {
+                return $default;
+            }
+            throw new CoreReflectionException($e->getMessage());
```
Ocramius: Spacing

src/Reflection/Adapter/ReflectionObject.php
```diff
@@ -370,23 +373,36 @@ public function isSubclassOf($class)
...
+        } catch (Throwable $e) {
+            throw new CoreReflectionException($e->getMessage());
```
Ocramius: Shouldn't we populate the previous exception?

src/Reflection/Adapter/ReflectionObject.php
Ocramius: Shouldn't we populate the previous exception?

src/Reflection/Adapter/ReflectionObject.php
```diff
@@ -370,23 +373,36 @@ public function isSubclassOf($class)
...
     /**
      * {@inheritDoc}
      */
     public function setStaticPropertyValue($name, $value)
     {
-        throw new Exception\NotImplemented('Not implemented');
+        try {
+            $this->betterReflectionObject->setStaticPropertyValue($name, $value);
+        } catch (Throwable $e) {
+            throw new CoreReflectionException($e->getMessage());
```
Ocramius: Shouldn't we populate the previous exception?

src/Reflection/Adapter/ReflectionProperty.php
```diff
@@ -47,15 +51,27 @@ public function getName()
      */
     public function getValue($object = null)
     {
-        throw new Exception\NotImplemented('Not implemented');
+        try {
+            return $this->betterReflectionProperty->getValue($object);
+        } catch (NoObjectProvided | NotAnObject $e) {
+            return null;
```
Ocramius: I think this was fixed somewhen... 7.1 maybe?

src/Reflection/Adapter/ReflectionProperty.php
```diff
@@ -47,15 +51,27 @@ public function getName()
...
+        } catch (Throwable $e) {
+            throw new CoreReflectionException($e->getMessage());
```
Ocramius: Shouldn't we populate the previous exception?

src/Reflection/Adapter/ReflectionProperty.php
```diff
@@ -47,15 +51,27 @@ public function getName()
...
      */
     public function setValue($object, $value = null)
     {
-        throw new Exception\NotImplemented('Not implemented');
+        try {
+            $this->betterReflectionProperty->setValue($object, $value);
+        } catch (NoObjectProvided | NotAnObject $e) {
+            return null;
+        } catch (Throwable $e) {
+            throw new CoreReflectionException($e->getMessage());
```
Ocramius: Shouldn't we populate the previous exception?

src/Reflection/ReflectionProperty.php
```diff
@@ -361,4 +372,120 @@ public function __clone()
...
+     * @throws ClassDoesNotExist
+     * @throws NoObjectProvided
+     * @throws NotAnObject
+     * @throws ObjectNotInstanceOfClass
+     */
+    public function getValue($object = null)
+    {
+        $this->assertAccessibility();
+
+        $declaringClassName = $this->getDeclaringClass()->getName();
```
Ocramius: Scenario to check: https://3v4l.org/8fBDe

src/Reflection/ReflectionProperty.php
```diff
@@ -361,4 +372,120 @@ public function __clone()
...
+     * @return mixed
+     * @throws PropertyNotAccessible
+     * @throws ClassDoesNotExist
+     * @throws NoObjectProvided
+     * @throws NotAnObject
+     * @throws ObjectNotInstanceOfClass
+     */
+    public function getValue($object = null)
+    {
+        $this->assertAccessibility();
```
Ocramius: IMO, our own implementation should drop the `setAccessible()` bullcrap

src/Reflection/ReflectionProperty.php
```diff
@@ -361,4 +372,120 @@ public function __clone()
...
+
+        $declaringClassName = $this->getDeclaringClass()->getName();
+
+        if ($this->isStatic()) {
+            $this->assertClassExist($declaringClassName);
+
+            return Closure::bind(function (string $propertyName) {
+                return self::${$propertyName};
```
Ocramius: Instead of `self::`, use `$declaringClassName`

src/Reflection/ReflectionProperty.php
```diff
@@ -361,4 +372,120 @@ public function __clone()
...
+            }, null, $declaringClassName)->__invoke($this->getName());
+        }
+
+        $this->assertObject($object);
+
+        return Closure::bind(function (string $propertyName) {
+            return $this->{$propertyName};
```
Ocramius: Instead of `$this`, use a parameter `$object` and call the closure with that

Ocramius: (this is to confuse the static analysers *less*

src/Reflection/Adapter/ReflectionProperty.php
kukulich: Nope, it throws a warning and returns `NULL`.

src/Reflection/ReflectionClass.php
kukulich: Ok. I just used the code that was already here :) https://github.com/Roave/BetterReflection/blob/master/src/Reflection/ReflectionClass.php#L1309

src/Reflection/Adapter/ReflectionProperty.php
Ocramius: Dang :-\

src/Reflection/ReflectionClass.php
Ocramius: Aye - I know my comments sometimes sound annoying and out of context, but I really just see the diffs only ^_^

kukulich: It's ok :) I hope I will someday prepare a pull request that can be merged straight away :)

src/Reflection/ReflectionProperty.php
kukulich: Done. `setAccessible()` implemented only in adapters.

kukulich:
@Ocramius I hope everything is ok now :)

Ocramius:
@kukulich merged, thanks! I tried to improve something in https://github.com/Roave/BetterReflection/commit/9e82778965aa1177d2bd7936b9314e8be5a62ebf, but failed due to how the test is (reasonably) designed.

:ship: and thanks!

kukulich:
@Ocramius I'm only on mobile now but if I understand it well then remove this line https://github.com/Roave/BetterReflection/blob/master/test/unit/Reflection/Adapter/ReflectionClassTest.php#L90 and implement special test for "getStaticProperties". Some other methods are already tested in similar way.

Ocramius:
That's alright: the naming is not the end of the world here, as it would
probably be more confusing to have a different one

Marco Pivetta

http://twitter.com/Ocramius

http://ocramius.github.com/

PR #101, opened 2015-08-14
#87 Loads of stubs

Ocramius:
See #87 

 - [x] parse phpdoc files from the php-src documentation
 - [x] generate stubs for classes with methods with default parameters in the phpdoc
 - [x] handle quirks in documentation
 - [x] write/commit stubs
 - [ ] ~~convert to method-only stubs (the stub is just the method, nothing else, eventually consider handling parameter-only stubs)~~ - not worth the amount of effort atm: we can easily do this by reproducing methods via `Zend\Code` tho.
 - [ ] ~~convert the source locator to replace "magically" reflected method stubs with the method stubs that are in the repository~~ - strictly related to previous point: not relevant right now.

PR #344, opened 2017-09-01
Fixed stubs searching

kukulich:
Build is failing because a lot of stubs is out of date...

The stubs searching is broken since https://github.com/Roave/BetterReflection/commit/3f9a4b317a6c1bc29ab207e4a6b259622bca5813 so it doesn't work almost two years.

May I remove the stubs functionality?

### DIFF
--- a/src/SourceLocator/Type/PhpInternalSourceLocator.php
+++ b/src/SourceLocator/Type/PhpInternalSourceLocator.php
@@ -96,7 +96,7 @@ final class PhpInternalSourceLocator extends AbstractSourceLocator
             return null;
         }
 
-        return __DIR__ . '/../../stub/' . $className . '.stub';
+        return __DIR__ . '/../../../stub/' . $className . '.stub';
     }
 
     /**
--- a/stub/mysqli_result.stub
+++ b/stub/mysqli_result.stub
@@ -1,5 +1,3 @@
-<?php
-
 class mysqli_result implements Traversable
 {
 


Ocramius:
> May I remove the stubs functionality?

The stubs are relevant for reflecting any internal classes, where internal reflection usually is a piece of garbage.

I put a lot of work in it, so it would be quite sad to see it go. Wouldn't it be better to split it out to a separate repo and have a regular test routine check against the PHP internals classes and the XML on http://docs.php.net/ ?

See https://github.com/Roave/BetterReflection/pull/101 for more details about this painful part of the implementation.

kukulich:
I know internal reflection is a garbage. I tried to implement reflection for internal function and it's impossible without stubs :(

Split stubs to separate repo sounds like a good step. 

Where is the XML on http://docs.php.net/ ? I'm blind probably and not able to find it :(

Ocramius:
See http://doc.php.net/tutorial/local-setup.php - that's where I used that XSLT mumbo-jumbo in the code snippet I wrote in #101

Ocramius:
@kukulich my first suggestion is to simply "hold off" for now.

asgrim:
BTW - I did try and investigate using the XML from the docs, but honestly it's so inconsistent :/

Ocramius:
> it's so inconsistent 

Yeah, there are no proper tests on their side either.

Ocramius:
In the past, I voted for not using them, but given the amount of work needed, what do you folks think about dropping our stubs, splitting stubbing logic out to a separate repo and using https://github.com/JetBrains/phpstorm-stubs instead?

asgrim:
> In the past, I voted for not using them

What was the reason for not using them, I don't recall? It looks like a sensible option IMO for managing the stubs.

Ocramius:
I think the main reason was that they are not 1:1 stubs, but rather IDE-friendly ones. They don't reflect the real API 100%.

That said, it would move the problem of "keeping things in sync" to a third party that benefits from bug reports as well.

Ocramius:
Also, more work for @GeeH, which is a big plus :+1:

kukulich:
> In the past, I voted for not using them, but given the amount of work needed, what do you folks think about dropping our stubs, splitting stubbing logic out to a separate repo and using https://github.com/JetBrains/phpstorm-stubs instead?

Sounds good to me!

Ocramius:
I manually fixed the stubs for now, and merged this patch after a rebase.

When misalignment between stubs and core happens again in the future, we should prioritize moving away from our own stubs.

Ocramius:
Thanks @kukulich!

kukulich:
I will probably prepare some PR later this year. The stubs espeacially for internal functions would really help PHPStan.

Ocramius:
Absolutely, although don't expect them to be accurate. Also, https://github.com/JetBrains/phpstorm-stubs is not tagged: that's a problem :-(

kukulich:
https://twitter.com/phpstorm/status/905744643534589952

PR #345, opened 2017-09-01
Introduced PHPBench with sample benchmark

dantleech:
Added PHPBench and a sample benchmark.

The sample benchmark reflects the PHPUnit `TestCase` class - this isn't ideal as it can obviously change, but a class of that magnitude is a good candidate.

I compared the memonising PR with the master branch below:

```
phpbench report --uuid=latest --uuid=latest-1 --report='{extends: aggregate, break: [ benchmark ], cols: [ mem_peak, best, mean, mode, worst, stdev ]}'
benchmark: PhpUnitTestCaseBench
+-------------+---------------+---------------+---------------+---------------+---------+
| mem_peak    | best          | mean          | mode          | worst         | stdev   |
+-------------+---------------+---------------+---------------+---------------+---------+
| 15,263,800b | 269,072.000μs | 269,072.000μs | 269,072.000μs | 269,072.000μs | 0.000μs |
| 15,247,312b | 414,300.000μs | 414,300.000μs | 414,300.000μs | 414,300.000μs | 0.000μs |
+-------------+---------------+---------------+---------------+---------------+---------+
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -89,6 +89,11 @@ jobs:
         - vendor/bin/phpunit --coverage-text --coverage-clover=clover.xml
         - php ocular.phar code-coverage:upload --format=php-clover clover.xml
 
+    - stage: Run benchmarks
+      php: 7.1
+      env: DEPENDENCIES=""
+      script: ./vendor/bin/phpbench run --progress=dots --iterations=1
+
   allow_failures:
     - php: 7.2
 
--- a/test/benchmark/PhpUnitTestCaseBench.php
+++ b/test/benchmark/PhpUnitTestCaseBench.php
@@ -8,6 +8,9 @@ use Roave\BetterReflection\Reflector\ClassReflector;
 use Roave\BetterReflection\Reflection\ReflectionMethod;
 use Roave\BetterReflection\Reflection\ReflectionParameter;
 
+/**
+ * @Iterations(10)
+ */
 class PhpUnitTestCaseBench
 {
     /**
--- a/test/benchmark/PhpUnitTestCaseBench.php
+++ b/test/benchmark/PhpUnitTestCaseBench.php
@@ -1,0 +1,42 @@
+<?php
+
+namespace Roave\BetterReflectionBenchmark;
+
+use PHPUnit\Framework\TestCase;
+use Roave\BetterReflection\BetterReflection;
+use Roave\BetterReflection\Reflector\ClassReflector;
+use Roave\BetterReflection\Reflection\ReflectionMethod;
+use Roave\BetterReflection\Reflection\ReflectionParameter;
+
+class PhpUnitTestCaseBench
+{
+    /**
+     * @var ClassReflector
+     */
+    private $reflector;
+
+    public function __construct()
+    {
+        $reflection = new BetterReflection();
+        $this->reflector = $reflection->classReflector();
+    }
+
+    /**
+     * @Subject()
+     */
+    public function reflect_phpunit_test_case()
+    {
+        $reflection = $this->reflector->reflect(TestCase::class);
+
+        /** @var $method ReflectionMethod */
+        foreach ($reflection->getMethods() as $method) {
+            $method->hasReturnType() ? $method->getReturnType()->__toString() : null;
+
+            /** @var $parameter ReflectionParameter */
+            foreach ($method->getParameters() as $parameter) {
+                $parameter->hasType() ? $parameter->getType()->__toString() : null;
+            }
+        }
+    }
+}
+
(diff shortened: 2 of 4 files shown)


test/benchmark/PhpUnitTestCaseBench.php
```diff
@@ -0,0 +1,45 @@
...
+        $reflection = new BetterReflection();
+        $this->reflector = $reflection->classReflector();
+    }
+
+    /**
+     * @Subject()
+     */
+    public function reflect_phpunit_test_case()
+    {
+        $reflection = $this->reflector->reflect(TestCase::class);
```
Ocramius: Any way to isolate parsing operations from reflection ones?

dantleech: I think you would need a profiler for that, there is also the [XDebug](http://phpbench.readthedocs.io/en/latest/extensions/xdebug.html#function-tracing) extension which can help.

Ocramius: I was asking to split the `$this->reflector->reflect(TestCase::class);` call into a preparation step, while the actual looping over methods/parameters should happen in here

dantleech: Well, you are still going to get contamination think? - `TestCase` extends `Assert` - not sure if it will trigger any more loading.

Ocramius: It will, but what bit are you benchmarking? 

```php
$reflection = $this->reflector->reflect(TestCase::class);
```

or 

```php
$method->hasReturnType() ? $method->getReturnType()->__toString() : null;
```

or

```php
$parameter->hasType() ? $parameter->getType()->__toString() : null;
```

dantleech: The context is that I encountered the performance issue wjen writing an auto-complete tool, and the performance was particularly bad on the `TestCase`. The tool iterates over methods, parameters, properties, etc. So this is kind-of reproducing that real-world scenario.

Happy to add smaller methods, but I think having the scenario is good (it should probably iterate over properties too). (

btw I already wrote similar tests for [here](https://github.com/phpactor/worse-reflection/tree/master/tests/Benchmarks).

Ocramius: > Happy to add smaller methods, but I think having the scenario is good (it should probably iterate over properties too).

I would really want to isolate the performance issues here, rather than grouping them.
Sorry if I sound annoying :-\

Ocramius:
@dantleech can you rebase as well? Rebasing already yields a 50% performance improvement:

```
./vendor/bin/phpbench rep
ort --file=original.xml --file=rebased.xml --report=compare
benchmark: PhpUnitTestCaseBench, subject: reflect_phpunit_test_case
+--------+--------+------+-----------------------------------------------------------+-------------------------------------------------------------------+
| groups | params | revs | vcs_branch:fix/#345-type-context-parsing-performance:mean | vcs_branch:fix/#345-type-context-parsing-performance-rebased:mean |
+--------+--------+------+-----------------------------------------------------------+-------------------------------------------------------------------+
|        | []     | 1    | 409,972.300μs                                             | 267,363.400μs                                                     |
+--------+--------+------+-----------------------------------------------------------+-------------------------------------------------------------------+
```

dantleech:
Have split:

```

File: .travis.yml [hunk 2]
```diff
@@ -89,6 +89,11 @@ jobs:
         - vendor/bin/phpunit --coverage-text --coverage-clover=clover.xml
         - php ocular.phar code-coverage:upload --format=php-clover clover.xml
 
+    - stage: Run benchmarks
+      php: 7.1
+      env: DEPENDENCIES=""
+      script: ./vendor/bin/phpbench run --progress=dots
```
dantleech: Just run the benchmarks so that they are maintained...

Ocramius:
checked this, and it is mostly re-parsing done by phpDocumentor:

selection_063

Even if we cache it, just calling the thing only once is a performance nightmare.

Given that we already have the parsed `namespace` and `use` nodes that are parent to the current function/method/property, I think that phasing out phpDocumentor is a possibility, although it is super-annoying to do so because of stupid PHP features such as group use statements and similar.

@jaapio thoughts? Anything we can do to reuse the already parsed `use` and `namespace` nodes in the decision logic?

Ocramius:
Another thing we can do is to create a `phpDocumentor\Reflection\Types\Context` manually, since the overhead is mostly in the `ContextFactory#createForNamespace()` logic.

Ocramius:
I am currently working on an implementation that doesn't use the `ContextFactory`, and the results are very promising :+1: 

Implementation is also quite simple =)

Ocramius:
See #346

Ocramius:
Closing here: all commits were imported into #346

PR #348, opened 2017-09-03
Migrate BC Breaks documentation to migration path docs

Ocramius:


kukulich:
Nice job!

Ocramius:
Merging as per @kukulich's proof-reading.

README.md
```diff
@@ -32,70 +32,33 @@ $ composer require roave/better-reflection
 ## Usage
 
 ```php
-use Roave\BetterReflection\Reflection\ReflectionClass;
+<?php
 
-$classInfo = ReflectionClass::createFromName('Foo\Bar\MyClass');
```
asgrim: Do we no longer support static instantiation at all? This was one of my "pitches" during talks; as in "the API is almost similar". I'd say that `ReflectionClass::createFromX` methods are still easier for the majority of people (at least for the `README.md`)

UPGRADE.md
```diff
@@ -0,0 +1,251 @@
...
+from the `Roave\BetterReflection\BetterReflection` kernel:
+
+```php
+<?php
+
+use Roave\BetterReflection\BetterReflection;
+$astLocator = (new BetterReflection())->astLocator();
+```
+
+### `Roave\BetterReflection\Reflection\Reflection*::createFromNode()` changes
```
asgrim: Should we mark all these methods as `@internal`?

Ocramius: @asgrim possibly a good idea

README.md
Ocramius: It is supported, it just isn't the best way anymore.

`(new BetterReflection)-><CTRL+SPACE>` works better right now

asgrim: I don't want to end up bikeshedding, but surely (for the most use cases):

```php
$r = ReflectionClass::createFromName(Foo::class);
```
is preferable to:
```php
$r = (new BetterReflection)->classReflector()->reflect(Foo::class);
```

Reasons to use `::createFrom*`:
 * It's shorter
 * Closer to PHP's core reflection API (i.e. `$r = new ReflectionClass(Foo::class);`). I think it's a smaller leap from core reflection for beginners (who will read the `README.md` first)

Reasons to use `new BetterReflection`:
 * Better understanding of how it's working under the hood (although steeper learning curve)
 * More flexible, easy path to starting to use custom locators and all sorts of magic nonsense

Ocramius: @asgrim both are still supported, but having `new BetterReflection()` as entry point makes it easier for folks that will at some point need to customize it, because they can just look into the call.

They are equivalent, but I'm just not advertising the other one anymore

asgrim: OK :+1:

UPGRADE.md
asgrim: see #358

PR #350, opened 2017-09-03
Fixes and improvements

kukulich:


src/Reflection/ReflectionClass.php
```diff
@@ -319,35 +324,12 @@ public function inNamespace() : bool
...
-                                $this->declaringNamespace,
-                                $trait,
-                                $this
-                            );
-                        }, $trait->getMethods());
-                    },
-                    $this->getTraits()
-                )
-            ),
+        // Note: methods are not merged via their name as array index, since internal PHP method
```
Ocramius: Split to a private method, put the comment on that private method

src/Reflection/ReflectionFunction.php
```diff
@@ -102,4 +102,16 @@ public function isDisabled() : bool
     {
         return false;
     }
+
+    /**
+     * Note - we cannot reflect on internal functions (as there is no PHP source
+     * code we can access. This means, at present, we can only EVER return null
```
Ocramius: "can't" = fail = throw, no?

src/Reflection/ReflectionMethod.php
```diff
@@ -94,8 +94,22 @@ public function getPrototype() : self
                 continue;
             }
 
-            if ($i->hasMethod($this->getName()) && $i->getMethod($this->getName())->isAbstract()) {
-                return $i->getMethod($this->getName());
+            if ($i->hasMethod($this->getName())) {
+                $method = $i->getMethod($this->getName());
```
Ocramius: I think `$i` needs a rename

src/Reflection/ReflectionMethod.php
```diff
@@ -94,8 +94,22 @@ public function getPrototype() : self
...
+
+                if ($method->isAbstract()) {
```
Ocramius: Maybe these conditions should bein a private method that we can call on the other instance? This includes the exception

src/SourceLocator/Located/InternalLocatedSource.php
```diff
@@ -8,12 +8,19 @@
...
+    /**
+     * @var string
+     */
+    private $extensionName;
+
     /**
      * {@inheritDoc}
      */
-    public function __construct(string $source)
+    public function __construct(string $source, string $extensionName)
```
Ocramius: BC break to be documented: you may consider waiting for the other patch to be merged though

src/SourceLocator/Located/InternalLocatedSource.php
```diff
@@ -8,12 +8,19 @@
...
     {
         parent::__construct($source, null);
+        $this->extensionName = $extensionName;
```
Ocramius: Empty line before this line

src/SourceLocator/Type/PhpInternalSourceLocator.php
```diff
@@ -37,18 +37,20 @@ protected function createLocatedSource(Identifier $identifier) : ?LocatedSource
...
              */
-            return new InternalLocatedSource("<?php\n\n" . $stub);
+            return new InternalLocatedSource("<?php\n\n" . $stub, $extensionName);
         }
 
         $stubber = $this->stubber;
 
         return new InternalLocatedSource(
-            "<?php\n\n" . $stubber($classReflection)
+            "<?php\n\n" . $stubber($classReflection), $extensionName
```
Ocramius: Not CS compliant

src/Reflection/ReflectionFunction.php
kukulich: I think that `NULL` is valid return value for non-internal functions.

The same comment is already there for some more methods: https://github.com/Roave/BetterReflection/blob/master/src/Reflection/ReflectionFunctionAbstract.php#L287

src/Reflection/ReflectionClass.php
kukulich: Is it necessary? The comment looks same as here: https://github.com/Roave/BetterReflection/commit/ca631dd343a940e9fc939c566ba02051dd311898

And I think there's already enough private methods :) `getMethods()` calls private method `getMethodsIndexedByName()` and this method calls private method `scanMethods()` :)

Ocramius: @kukulich yes - the number of private methods is not scary, but their singular size is. The fact that there is a comment here should move the comment to the method, rather than inline

src/Reflection/ReflectionFunction.php
Ocramius: Yeah, but is this viable also for classes that we know to be internal? Shouldn't we throw there?

kukulich: I'm probably a little dumb today :) I'm sorry I don't get how you mean it.

src/Reflection/ReflectionMethod.php
kukulich: Renamed in new commit.

src/Reflection/ReflectionMethod.php
kukulich: Refactored in new commit https://github.com/Roave/BetterReflection/pull/350/commits/a70d5b5700399759a7a669cc082aaa2fcd970e38

However I'm really not happy with the name `findPrototype()`. I hope someone find better name.

src/SourceLocator/Located/InternalLocatedSource.php
kukulich: Yes, I will fix this after https://github.com/Roave/BetterReflection/pull/348 is merged.

src/Reflection/ReflectionClass.php
kukulich: I've tried to improve it in special commit: https://github.com/Roave/BetterReflection/pull/350/commits/b9046e81537a13613cb484c6d1660a33067ab7f4

src/Reflection/ReflectionFunction.php
Ocramius: A `ReflectionFunction` that is defined and is internal can be reflected on the fly, IMO

src/SourceLocator/Located/InternalLocatedSource.php
```diff
@@ -9,11 +9,18 @@
...
     /**
+     * @var string
+     */
+    private $extensionName;
+
+    /**
      * {@inheritDoc}
      */
-    public function __construct(string $source)
+    public function __construct(string $source, string $extensionName)
```
Ocramius: @kukulich remember to add this BC break to the `UPGRADE.md`

Ocramius:
@kukulich also this one needs another rebase :-\

Ocramius:
Rebased and manually applied via Roave/BetterReflection@9ac8efa5e00b9afeba90c453a460fda62f220f51 :ship:

kukulich:
Thanks :)

PR #352, opened 2017-09-04
Cleanup: generic type/static analysis fixes

Ocramius:


src/NodeCompiler/CompileNodeToValue.php
```diff
@@ -277,4 +199,114 @@ private function getConstantDeclaringClass(string $constantName, ReflectionClass
 
         return $parentClass ? $this->getConstantDeclaringClass($constantName, $parentClass) : null;
     }
+
+    /**
+     * @return callable[] indexed by node class name
+     */
+    private static function loadEvaluators() : array
```
Ocramius: I should probably split this out to an external class...

Ocramius: Thinking twice: no, that's probably not such a good idea. Having it split out for the sake of it would probably just increase complexity right now, and cause a cyclic dependency.

src/Reflection/ReflectionObject.php
```diff
@@ -66,30 +64,6 @@ public static function export($instance = null) : string
     }
 
     /**
-     * Cannot instantiate this way, use ReflectionObject::createFromInstance
-     *
-     * @throws \LogicException
-     */
-    public static function createFromNode(
```
Ocramius: Note: dropped because it is not an instance method, so an override does not really serve any purpose. The user would get a `ReflectionClass` instance anyway.

src/Reflection/ReflectionObject.php
```diff
@@ -66,30 +64,6 @@ public static function export($instance = null) : string
...
-    ) : ReflectionClass {
-        throw new LogicException('Cannot create a ReflectionObject from node - use ReflectionObject::createFromInstance');
-    }
-
-    /**
-     * Cannot instantiate this way, use ReflectionObject::createFromInstance
-     *
-     * @throws \LogicException
-     */
-    public static function createFromName(string $className) : ReflectionClass
```
Ocramius: Note: dropped because it is not an instance method, so an override does not really serve any purpose. The user would get a `ReflectionClass` instance anyway.

src/Reflection/ReflectionObject.php
Ocramius: Also: not really a BC break since this method was only throwing

src/Reflection/ReflectionObject.php
Ocramius: Also: not really a BC break since this method was only throwing

test/unit/NodeCompiler/CompileNodeToValueTest.php
```diff
@@ -151,6 +150,9 @@ public function nodeProvider() : array
             ['3 <= 2', false],
             ['PHP_INT_MAX', \PHP_INT_MAX],
             ['PHP_EOL', \PHP_EOL],
+            ['1 <=> 4', -1],
```
Ocramius: Not sure why the spaceship operator was previously skipped

test/unit/Reflection/ReflectionFunctionAbstractTest.php
```diff
@@ -71,25 +68,6 @@ public function testExportThrowsException() : void
         ReflectionFunctionAbstract::export();
     }
 
-    public function testPopulateFunctionAbstractThrowsExceptionWithInvalidNode() : void
```
Ocramius: No longer possible, as the type is enforced by the type hint. The reflectionish approach of the test is no longer valid either.

test/unit/Reflection/ReflectionMethodTest.php
```diff
@@ -334,21 +330,6 @@ public function testGetPrototype(string $class, string $method, ?string $expecte
         self::assertSame($expectedPrototype, $b->getDeclaringClass()->getName());
     }
 
-    public function testGetMethodNodeFailsWhenNodeIsNotClassMethod() : void
```
Ocramius: Also here: no longer possible, as this change enforces the constraints via type hints.

Ocramius:
Paging @kukulich as well

src/SourceLocator/FileChecker.php
```diff
@@ -19,15 +19,15 @@ public static function assertReadableFile(string $filename) : void
         }
 
         if ( ! \file_exists($filename)) {
```
kukulich: I suggest to remove this check and use only `is_file`.

kukulich:
LGTM 👍

Ocramius:
Shipping as per @kukulich's review.

PR #356, opened 2017-09-04
Replacing internal `assertObject` return type with `object`, removing therefore the implicit upcast

Ocramius:
Just hunting for static analysis issues. 7.2 will make this more pleasant in the future.


Ocramius:
All checks :green_apple:

:ship:

PR #357, opened 2017-09-05
Implemented missing functions in ReflectionFunction

kukulich:


src/Reflection/ReflectionFunction.php
```diff
@@ -114,4 +116,77 @@ public function getExtensionName() : ?string
...
+     * @param mixed[] $args
+     *
+     * @return mixed
+     *
+     * @throws NotImplemented
+     * @throws FunctionDoesNotExist
+     */
+    public function invokeArgs(array $args = [])
+    {
+        $this->assertIsNoClosure();
```
Ocramius: Why can't you call it on a function?

kukulich: I think that the only possible solution for `invokeArgs` working for closures is to build PHP code from AST and use `eval`. And it will fail anyway for code like this:

```php
<?php

class Foo
{
    private $foo = 'foo';
    
    public function boo()
    {
        return (function () {
            return $this->foo;
        })();
    }
}

Ocramius: Ah, got it. Didn't consider that the closure wasn't loaded yet.

Ocramius:
Thanks @kukulich!

PR #367, opened 2017-09-08
Improved Closure stub

kukulich:
Not sure hot to fix the failing test...


Ocramius:
@kukulich this needs to be excluded from `PHP_VERSION_ID < 70200`

kukulich:
The method is already in PHP 7.1 but internal reflection reports callable parameter as not callable https://3v4l.org/EeHXS

Just added ugly if.

Ocramius:
Oh nice! (not)

PR #371, opened 2017-09-21
MemoizingParser: Fix parent interface misstype of parse() method

TomasVotruba:
I believe this is PHP 7.2 feature. The parsing fails otherwise

image

### DIFF
--- a/src/SourceLocator/Ast/Parser/MemoizingParser.php
+++ b/src/SourceLocator/Ast/Parser/MemoizingParser.php
@@ -32,7 +32,7 @@ final class MemoizingParser implements Parser
      *
      * @phpcsSuppress SlevomatCodingStandard.TypeHints.TypeHintDeclaration.MissingParameterTypeHint
      */
-    public function parse($code, ?ErrorHandler $errorHandler = null) : ?array
+    public function parse(string $code, ?ErrorHandler $errorHandler = null) : ?array
     {
         // note: this code is mathematically buggy by default, as we are using a hash to identify
         //       cache entries. The string length is added to further reduce likeliness (although


TomasVotruba:
Oh, I have dev-master version of PHPParser and this class extends it :/
It has `string` typehint but this class doesn't.

Any idea how to workaround it?

Ocramius:
Can't workaround for now: if PHP-Parser has breaking changes, we'll have to adapt ourselves too

TomasVotruba:
Well, it's only in `dev-master` and God knows, when that's gonna be tagged.

I'll use https://github.com/cweagans/composer-patches for now

Ocramius:
@TomasVotruba anything in particular that you rely upon?

TomasVotruba:
This patch fixed it: https://github.com/RectorPHP/Rector/pull/44/commits/2c53ce3f8e5f8055095fc304aab4b0588ed4cc5f#diff-13073b040fd8a4bec22c435bd7392dc9

Ocramius:
I was actually wondering why you needed PHP-Parser dev-master 🤔

TomasVotruba:
Because there is [format-preserving AST printer](https://github.com/nikic/PHP-Parser/issues/344#issuecomment-298162516), that I need to refactor the code without making it ugly.

Thanks to that I'll be able to partially upgrade app from Symfony 3.0 to 4.0 while keeping the original formatting.

Ocramius:
Makes sense, although we won't be able to support both versions of PHP-Parser for now :-\

TomasVotruba:
No problem, I solved this way I'm fine with :+1:

PR #372, opened 2017-09-21
Added missing parent::setUp() in tests

kukulich:


TomasVotruba:
Just wondering, that is the motivation for this? 

Is there any code in parent call?

Ocramius:
No, but parent calls should always be there unless the child implementation
can fully guarantee to initialise the parent's state.

The parent implementation can also change, so it isn't a fixed target.

PR #375, opened 2017-10-16
Implement new instance args

### LINKED ISSUE #142: ReflectionClass::newInstance proposal
It would basically do [this](https://github.com/Roave/BetterReflection/pull/141/files#diff-264ad048c579cda8edbfa19b80318e02R22).

If class NOT already loaded:

```php
public function newInstance(/*...*/)
{
    $classCode = (new \PhpParser\PrettyPrinter\Standard())->prettyPrint($this->getAst());
    file_put_contents('/tmp/foo.php', '<?php ' . $classCode);
    require_once('/tmp/foo.php');
    unlink('/tmp/foo.php');
    return new MyClass(/*...*/);
}
```

However if class has already been loaded, we must do one of:

* Throw an exception - cannot create instance of this class (we may have modified the operation, which could lead to unintended issues
* Create a temporary wrapper class instance that extends the already-loaded class, i.e.:
```php
public function newInstance(/*...*/)
{
    $this->setName('MyClass_Fake_Reflection_123'); // doesn't exist yet
    $this->setClassFromWhichThisExtends('MyClass'); // doesn't exist, choose a better name
    $classCode = (new \PhpParser\PrettyPrinter\Standard())->prettyPrint($this->getAst());
    file_put_contents('/tmp/foo.php', '<?php ' . $classCode);
    require_once('/tmp/foo.php');
    unlink('/tmp/foo.php');
    return new MyClass_Fake_Reflection_123(/*...*/);
}
```
* Something else might be better.

asgrim:
Fixes #142


src/Reflection/Adapter/ReflectionClass.php
```diff
@@ -366,10 +367,15 @@ public function newInstanceWithoutConstructor()
 
     /**
      * {@inheritDoc}
+     * @throws \LogicException
      */
-    public function newInstanceArgs(?array $args = null)
+    public function newInstanceArgs(array $args = [], ClassLoader $classLoader = null)
```
Ocramius: This... is ugly. Especially since we want to maintain reflection compatibility.

I don't think this is the right way to go with this implementation.

asgrim: It's not nice indeed; other option is to provide a default setup for `ClassLoader`, so like:

```php
public function newInstanceArgs(?array $args = null)
{
    $classLoader = new ClassLoader(new EvalLoader(new PhpPrettyPrinter()));
}
```

but I didn't want to get tied to that specific implementation... not sure how else to handle this :s

Ocramius: @asgrim add the `ClassLoader` to the `BetterReflection` core or such?

asgrim: That's kinda the problem, using the `EvalLoader` method is crap, the `FileCacheLoader` is much more preferable so should be encouraged, however that needs a folder path specified, so can't be default :/

Ocramius: @asgrim I think we should leave the signature be. It is better to not have this thing at all than to have a Frankenstein.

Ocramius:
Removing this from the `3.0.0` milestone - will be pushed down the line for later.

asgrim:

Ocramius:
Closing here: AFAIK we already implemented `newInstanceArgs()`, so this patch can be closed :)

PR #377, opened 2017-10-21
composer: allow PHP 7.1

### LINKED ISSUE #376: Incorrect PHP version constraint
https://github.com/Roave/BetterReflection/blob/master/composer.json#L6 uses `>7.1.0` instead of `>=7.1.0`. Looking at commit history, this does not seem to be intentional.

TomasVotruba:
closes #376


TomasVotruba:
:+1:

Issue #378, opened 2017-11-01
TypeError: ReflectionFunctionAbstract::getBodyAst() returns null for empty body


codeliner:
Got following error for an empty method body: 

`TypeError: Return value of Roave\BetterReflection\Reflection\ReflectionFunctionAbstract::getBodyAst() must be of the type array, null returned`

[PhpParser\Node\Stmt\ClassMethod](https://github.com/nikic/PHP-Parser/blob/master/lib/PhpParser/Node/Stmt/ClassMethod.php#L20) says `stmts` prop is always an array but seems not to be the case.

I'll check if I can provide a bugfix PR to php-parser repo. 
Just wanted to let you know about the bug. Tell me if I should provide a failing test + PR for BetterReflection, too.

Ocramius:
If the bug is with PHP-Parser then it probably should only be cascaded here
if it can't be fixed on that side.

codeliner:
Looks like it is wanted behavior: https://github.com/nikic/PHP-Parser/commit/73dc35cbbd8dd464cedbe83b9cd9c6e1d1ebf144
and https://github.com/nikic/PHP-Parser/issues/235#issuecomment-194505139
/cc @nikic

I'll submit a PR (and a small one for php-parser to update doc block of property)

Ocramius:
That would be sufficient to make phpstan crash in this project, leading to
test+fix on our side

PR #379, opened 2017-11-01
Return empty ast for interface method body

codeliner:
Fixes: #378 

Problem is caused when inspecting method of an interface. See also https://github.com/nikic/PHP-Parser/issues/235#issuecomment-194505139

### DIFF
--- a/src/Reflection/ReflectionFunctionAbstract.php
+++ b/src/Reflection/ReflectionFunctionAbstract.php
@@ -498,7 +498,7 @@ abstract class ReflectionFunctionAbstract implements CoreReflector
      */
     public function getBodyAst() : array
     {
-        return $this->node->stmts;
+        return $this->node->stmts ?: [];
     }
 
     /**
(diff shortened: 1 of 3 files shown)


codeliner:
:thinking: Not sure if the failing test is caused by my change. Cannot reproduce locally. I guess I would need to install intl extension first, right? 

But I did not change the public API of `ReflectionMethod` so why the hell should this one test fail now? Any ideas?

kukulich:
It’s probably caused by new PHP version.

kukulich:
I will fix the tesr later.

Ocramius:
The failure is on appveyor, which is as stable as walking on 🍌 peel.

Will re-trigger it once home.

Ocramius:
Ah yes, requires indeed an adaptation of the tests. Will apply shortly.

Ocramius:
Patch provided at https://github.com/Roave/BetterReflection/pull/380

Ocramius:
Waiting for builds, then tagging

PR #381, opened 2017-11-02
Fix wrong parameter name

tobemedia:
rename $myMethod to $myParameter as expected by getParameter('myParameter')


Ocramius:
@tobemedia thanks!

Issue #389, opened 2017-12-23
Throw ReflectionException when property not found?


muglug:
`Roave\BetterReflection\Reflection\Adapter\ReflectionClass::getProperty` returns a nullable object, but the method it's extending throws an exception if a given property isn't found. Should that behaviour and return type be matched?

```php
$class = new ReflectionClass('ReflectionClass');
$property = $class->getProperty('naame');
var_dump($property);
?>
```

Ocramius:
Yeah, that would probably be more correct: https://3v4l.org/pZqVl

kukulich:
I’ll prepare a PR in next few days.

kukulich:
Fixed in https://github.com/Roave/BetterReflection/pull/391

Ocramius:
Handled in #391

PR #391, opened 2018-01-02
Method getProperty() of adapters should throw exception when property doesn't exist

kukulich:

PR #394, opened 2018-01-05
Fix issues found by static analysis

muglug:
This attempts to adhere, as much as possible, to the base `Reflection*` signatures. In the case of https://github.com/Roave/BetterReflection/issues/389, I made the current different return type explicit so that changing will be a clear breaking change.

According to the docs, static `::export` methods can return a value, so I allowed that even if your model does not. You may want to add an explicit `return null`.

`Roave\BetterReflection\Reflection\Adapter\ReflectionMethod::getClosure` wasn't returning a `Closure`, now it should.

### DIFF
--- a/src/NodeCompiler/CompileNodeToValue.php
+++ b/src/NodeCompiler/CompileNodeToValue.php
@@ -130,10 +130,7 @@ class CompileNodeToValue
      */
     private function compileClassConstFetch(Node\Expr\ClassConstFetch $node, CompilerContext $context)
     {
-        if ($node->class instanceof Node\Expr || $node->name instanceof Node\Expr\Error) {
-            throw new \LogicException('Found unexpected value when evaluating class constant');
-        }
-
+        /** @var Node\Name $node->class */
         $className = $node->class->toString();
 
         if ($node->name === 'class') {
@@ -152,6 +149,7 @@ class CompileNodeToValue
             $classInfo = $context->getReflector()->reflect($className);
         }
 
+        /** @var string $node->name */
         $reflectionConstant = $classInfo->getReflectionConstant($node->name);
 
         return $this->__invoke(
--- a/src/NodeCompiler/CompileNodeToValue.php
+++ b/src/NodeCompiler/CompileNodeToValue.php
@@ -130,6 +130,10 @@ class CompileNodeToValue
      */
     private function compileClassConstFetch(Node\Expr\ClassConstFetch $node, CompilerContext $context)
     {
+        if ($node->class instanceof Node\Expr || $node->name instanceof Node\Expr\Error) {
+            throw new \LogicException('Found unexpected value when evaluating class constant');
+        }
+
         $className = $node->class->toString();
 
         if ($node->name === 'class') {
--- a/src/Reflection/Adapter/ReflectionClass.php
+++ b/src/Reflection/Adapter/ReflectionClass.php
@@ -184,9 +184,11 @@ class ReflectionClass extends CoreReflectionClass
     }
 
     /**
-     * {@inheritDoc}
+     * @param string $name
+     *
+     * @return ReflectionProperty|null
      */
-    public function getProperty($name) : ?ReflectionProperty
+    public function getProperty($name)
     {
         $betterReflectionProperty = $this->betterReflectionClass->getProperty($name);
 
--- a/src/Reflection/Adapter/ReflectionClass.php
+++ b/src/Reflection/Adapter/ReflectionClass.php
@@ -184,9 +184,7 @@ class ReflectionClass extends CoreReflectionClass
     }
 
     /**
-     * @param string $name
-     *
-     * @return ReflectionProperty|null
+     * {@inheritDoc}
      */
     public function getProperty($name)
     {
--- a/src/Reflection/Adapter/ReflectionClass.php
+++ b/src/Reflection/Adapter/ReflectionClass.php
@@ -186,7 +186,7 @@ class ReflectionClass extends CoreReflectionClass
     /**
      * {@inheritDoc}
      */
-    public function getProperty($name)
+    public function getProperty($name) : ?ReflectionProperty
     {
         $betterReflectionProperty = $this->betterReflectionClass->getProperty($name);
 
--- a/src/Reflection/Adapter/ReflectionFunction.php
+++ b/src/Reflection/Adapter/ReflectionFunction.php
@@ -27,7 +27,7 @@ class ReflectionFunction extends CoreReflectionFunction
      */
     public static function export($name, $return = null)
     {
-        BetterReflectionFunction::export(...func_get_args());
+        return BetterReflectionFunction::export(...func_get_args());
     }
 
     /**
--- a/src/Reflection/Adapter/ReflectionFunction.php
+++ b/src/Reflection/Adapter/ReflectionFunction.php
@@ -27,7 +27,7 @@ class ReflectionFunction extends CoreReflectionFunction
      */
     public static function export($name, $return = null)
     {
-        return BetterReflectionFunction::export(...func_get_args());
+        BetterReflectionFunction::export(...func_get_args());
     }
 
     /**
--- a/src/Reflection/Adapter/ReflectionMethod.php
+++ b/src/Reflection/Adapter/ReflectionMethod.php
@@ -34,7 +34,7 @@ class ReflectionMethod extends CoreReflectionMethod
      */
     public static function export($class, $name, $return = null)
     {
-        BetterReflectionMethod::export(...func_get_args());
+        return BetterReflectionMethod::export(...func_get_args());
     }
 
     /**
@@ -313,7 +313,7 @@ class ReflectionMethod extends CoreReflectionMethod
     public function getClosure($object = null)
     {
         try {
-            $this->betterReflectionMethod->getClosure($object);
+            return $this->betterReflectionMethod->getClosure($object);
         } catch (NoObjectProvided | NotAnObject $e) {
             return null;
         } catch (Throwable $e) {
--- a/src/Reflection/Adapter/ReflectionMethod.php
+++ b/src/Reflection/Adapter/ReflectionMethod.php
@@ -34,7 +34,7 @@ class ReflectionMethod extends CoreReflectionMethod
      */
     public static function export($class, $name, $return = null)
     {
-        return BetterReflectionMethod::export(...func_get_args());
+        BetterReflectionMethod::export(...func_get_args());
     }
 
     /**
--- a/src/Reflection/Adapter/ReflectionObject.php
+++ b/src/Reflection/Adapter/ReflectionObject.php
@@ -175,9 +175,11 @@ class ReflectionObject extends CoreReflectionObject
     }
 
     /**
-     * {@inheritDoc}
+     * @param string $name
+     *
+     * @return ReflectionProperty|null
      */
-    public function getProperty($name) : ?ReflectionProperty
+    public function getProperty($name)
     {
         $property = $this->betterReflectionObject->getProperty($name);
 
--- a/src/Reflection/Adapter/ReflectionObject.php
+++ b/src/Reflection/Adapter/ReflectionObject.php
@@ -177,7 +177,7 @@ class ReflectionObject extends CoreReflectionObject
     /**
      * {@inheritDoc}
      */
-    public function getProperty($name)
+    public function getProperty($name) : ?ReflectionProperty
     {
         $property = $this->betterReflectionObject->getProperty($name);
 
--- a/src/Reflection/Adapter/ReflectionObject.php
+++ b/src/Reflection/Adapter/ReflectionObject.php
@@ -175,9 +175,7 @@ class ReflectionObject extends CoreReflectionObject
     }
 
     /**
-     * @param string $name
-     *
-     * @return ReflectionProperty|null
+     * {@inheritDoc}
      */
     public function getProperty($name)
     {
--- a/src/Reflection/ReflectionClass.php
+++ b/src/Reflection/ReflectionClass.php
@@ -350,7 +350,6 @@ class ReflectionClass implements Reflection, CoreReflector
      */
     public function getImmediateMethods(?int $filter = null) : array
     {
-        /** @var \ReflectionMethod[] $methods */
         $methods = array_map(
             function (ClassMethod $methodNode) : ReflectionMethod {
                 return ReflectionMethod::createFromNode(
--- a/src/Reflection/ReflectionClass.php
+++ b/src/Reflection/ReflectionClass.php
@@ -350,6 +350,7 @@ class ReflectionClass implements Reflection, CoreReflector
      */
     public function getImmediateMethods(?int $filter = null) : array
     {
+        /** @var ReflectionMethod[] $methods */
         $methods = array_map(
             function (ClassMethod $methodNode) : ReflectionMethod {
                 return ReflectionMethod::createFromNode(
--- a/src/Reflection/ReflectionClassConstant.php
+++ b/src/Reflection/ReflectionClassConstant.php
@@ -188,7 +188,7 @@ class ReflectionClassConstant implements CoreReflector
     /**
      * {@inheritDoc}
      */
-    public static function export() : void
+    public static function export()
     {
         throw new Exception('Unable to export statically');
     }
--- a/src/Reflection/ReflectionClassConstant.php
+++ b/src/Reflection/ReflectionClassConstant.php
@@ -185,10 +185,7 @@ class ReflectionClassConstant implements CoreReflector
         return ReflectionClassConstantStringCast::toString($this);
     }
 
-    /**
-     * {@inheritDoc}
-     */
-    public static function export()
+    public static function export() : void
     {
         throw new Exception('Unable to export statically');
     }
--- a/src/Reflection/ReflectionFunctionAbstract.php
+++ b/src/Reflection/ReflectionFunctionAbstract.php
@@ -67,6 +67,9 @@ abstract class ReflectionFunctionAbstract implements CoreReflector
     {
     }
 
+    /**
+     * {@inheritDoc}
+     */
     public static function export()
     {
         throw new Exception('Unable to export statically');
@@ -515,7 +518,7 @@ abstract class ReflectionFunctionAbstract implements CoreReflector
      */
     public function setBodyFromString(string $newBody) : void
     {
-        /** @var Node[] */
+        /** @var Node[] $this->node->stmts */
         $this->node->stmts = $this->loadStaticParser()->parse('<?php ' . $newBody);
     }
 
--- a/src/Reflection/ReflectionFunctionAbstract.php
+++ b/src/Reflection/ReflectionFunctionAbstract.php
@@ -39,7 +39,7 @@ abstract class ReflectionFunctionAbstract implements CoreReflector
     public const CLOSURE_NAME = '{closure}';
 
     /**
-     * @var NamespaceNode
+     * @var ?NamespaceNode
      */
     private $declaringNamespace;
 
@@ -67,7 +67,7 @@ abstract class ReflectionFunctionAbstract implements CoreReflector
     {
     }
 
-    public static function export() : void
+    public static function export()
     {
         throw new Exception('Unable to export statically');
     }
@@ -75,7 +75,7 @@ abstract class ReflectionFunctionAbstract implements CoreReflector
     /**
      * Populate the common elements of the function abstract.
      *
-     * @param Node\Stmt\ClassMethod|Node\FunctionLike|Node\Stmt|Node $node Node has to be processed by the PhpParser\NodeVisitor\NameResolver
+     * @param Node\Stmt\ClassMethod|Node\Stmt\Function_|Node\Expr\Closure $node Node has to be processed by the PhpParser\NodeVisitor\NameResolver
      *
      * @throws \Roave\BetterReflection\Reflection\Exception\InvalidAbstractFunctionNodeType
      */
@@ -515,7 +515,7 @@ abstract class ReflectionFunctionAbstract implements CoreReflector
      */
     public function setBodyFromString(string $newBody) : void
     {
-        $this->node->stmts = $this->loadStaticParser()->parse('<?php ' . $newBody);
+        $this->node->stmts = $this->loadStaticParser()->parse('<?php ' . $newBody) ?: [];
     }
 
     /**
--- a/src/Reflection/ReflectionFunctionAbstract.php
+++ b/src/Reflection/ReflectionFunctionAbstract.php
@@ -67,10 +67,7 @@ abstract class ReflectionFunctionAbstract implements CoreReflector
     {
     }
 
-    /**
-     * {@inheritDoc}
-     */
-    public static function export()
+    public static function export() : void
     {
         throw new Exception('Unable to export statically');
     }
@@ -518,7 +515,6 @@ abstract class ReflectionFunctionAbstract implements CoreReflector
      */
     public function setBodyFromString(string $newBody) : void
     {
-        /** @var Node[] $this->node->stmts */
         $this->node->stmts = $this->loadStaticParser()->parse('<?php ' . $newBody);
     }
 
--- a/src/Reflection/ReflectionFunctionAbstract.php
+++ b/src/Reflection/ReflectionFunctionAbstract.php
@@ -515,7 +515,8 @@ abstract class ReflectionFunctionAbstract implements CoreReflector
      */
     public function setBodyFromString(string $newBody) : void
     {
-        $this->node->stmts = $this->loadStaticParser()->parse('<?php ' . $newBody) ?: [];
+        /** @var Node[] */
+        $this->node->stmts = $this->loadStaticParser()->parse('<?php ' . $newBody);
     }
 
     /**
--- a/src/Reflection/ReflectionFunctionAbstract.php
+++ b/src/Reflection/ReflectionFunctionAbstract.php
@@ -39,7 +39,7 @@ abstract class ReflectionFunctionAbstract implements CoreReflector
     public const CLOSURE_NAME = '{closure}';
 
     /**
-     * @var ?NamespaceNode
+     * @var NamespaceNode|null
      */
     private $declaringNamespace;
 
--- a/src/Reflection/ReflectionParameter.php
+++ b/src/Reflection/ReflectionParameter.php
@@ -78,7 +78,7 @@ class ReflectionParameter implements CoreReflector
     {
     }
 
-    public static function export() : void
+    public static function export()
     {
         throw new Exception('Unable to export statically');
     }
@@ -192,6 +192,14 @@ class ReflectionParameter implements CoreReflector
         $defaultValueNode = $this->node->default;
 
         if ($defaultValueNode instanceof Node\Expr\ClassConstFetch) {
+            if ($defaultValueNode->class instanceof Node\Expr) {
+                throw new LogicException('Class constant classes cannot be expressions');
+            }
+
+            if ($defaultValueNode->name instanceof Node\Expr\Error) {
+                throw new LogicException('Cannot deal with error nodes');
+            }
+
             $className = $defaultValueNode->class->toString();
 
             if ($className === 'self' || $className === 'static') {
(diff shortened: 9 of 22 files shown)


src/Reflection/Adapter/ReflectionFunction.php
```diff
@@ -26,7 +26,7 @@ public function __construct(BetterReflectionFunction $betterReflectionFunction)
      */
     public static function export($name, $return = null)
     {
-        BetterReflectionFunction::export(...\func_get_args());
+        return BetterReflectionFunction::export(...\func_get_args());
```
Ocramius: WOW that's bad O_O

src/Reflection/Adapter/ReflectionClass.php
```diff
@@ -180,7 +180,7 @@ public function hasProperty($name)
     /**
      * {@inheritDoc}
      */
-    public function getProperty($name)
+    public function getProperty($name) : ?ReflectionProperty
```
Ocramius: AFAIK, all adapters didn't really have return types. Wondering if we should change that...

src/NodeCompiler/CompileNodeToValue.php
```diff
@@ -121,6 +121,10 @@ private function compileConstFetch(Node\Expr\ConstFetch $constNode)
      */
     private function compileClassConstFetch(Node\Expr\ClassConstFetch $node, CompilerContext $context)
     {
+        if ($node->class instanceof Node\Expr || $node->name instanceof Node\Expr\Error) {
```
Ocramius: This would require also a test to reach this code location. This may as well be an assumption that is hard to code within the types we have available.

src/Reflection/ReflectionFunctionAbstract.php
```diff
@@ -62,15 +62,15 @@ protected function __construct()
     {
     }
 
-    public static function export() : void
+    public static function export()
```
Ocramius: This is indeed supposed to be `void`

src/Reflection/ReflectionFunctionAbstract.php
```diff
@@ -34,7 +34,7 @@
     public const CLOSURE_NAME = '{closure}';
 
     /**
-     * @var NamespaceNode
+     * @var ?NamespaceNode
```
Ocramius: `NamespaceNode|null`

src/Reflection/ReflectionFunctionAbstract.php
```diff
@@ -510,7 +510,7 @@ public function setBodyFromClosure(Closure $newBody) : void
      */
     public function setBodyFromString(string $newBody) : void
     {
-        $this->node->stmts = $this->loadStaticParser()->parse('<?php ' . $newBody);
+        $this->node->stmts = $this->loadStaticParser()->parse('<?php ' . $newBody) ?: [];
```
Ocramius: Hmm... the parser should never yield non-arrays...

src/Reflection/ReflectionParameter.php
```diff
@@ -184,6 +184,14 @@ private function parseDefaultValueNode() : void
         $defaultValueNode = $this->node->default;
 
         if ($defaultValueNode instanceof Node\Expr\ClassConstFetch) {
+            if ($defaultValueNode->class instanceof Node\Expr) {
```
Ocramius: Same scenario as above - this would require a test case

src/Reflection/ReflectionProperty.php
```diff
@@ -67,7 +67,7 @@ private function __construct()
     {
     }
 
-    public static function export() : void
+    public static function export()
```
Ocramius: This is indeed `void`

src/Reflection/ReflectionType.php
```diff
@@ -21,7 +21,7 @@ class ReflectionType
     ];
 
     /**
-     * @var $type
+     * @var string
```
Ocramius: :+1:

src/SourceLocator/Ast/FindReflectionsInTree.php
```diff
@@ -87,7 +87,7 @@ public function __construct(
                 $this->astConversionStrategy = $astConversionStrategy;
             }
 
-            public function enterNode(Node $node) : void
```
Ocramius: These should be kept

src/Reflection/Adapter/ReflectionClass.php
muglug: I can put it in the docblock?

src/Reflection/ReflectionFunctionAbstract.php
muglug: In *theory* it can yield `null` just based on the return type.

src/SourceLocator/Ast/FindReflectionsInTree.php
muglug: This method inherits from a method that has a non-`void` return type

src/Reflection/ReflectionProperty.php
muglug: This method inherits from a method that has a non-`void` return type

src/Reflection/ReflectionFunctionAbstract.php
muglug: This method inherits from a method that has a non-`void` return type

src/SourceLocator/Type/FileIteratorSourceLocator.php
```diff
@@ -25,7 +25,7 @@ class FileIteratorSourceLocator implements SourceLocator
     private $aggregateSourceLocator;
 
     /**
-     * @var \Iterator|\SplFileInfo[]
+     * @var \Iterator<\SplFileInfo>
```
kukulich: @muglug Is there an editor that can understand this? I don't think that eg. PHPStorm supports this syntax.

src/NodeCompiler/CompileNodeToValue.php
muglug: `$node->class` cannot ever be a `Node\Expr` here, but `Node\Expr\ClassConstFetch::$class` can be. I suppose the error node may be possible, if you allow errors in `PhpParser`.

src/Reflection/Adapter/ReflectionClass.php
muglug: PHP *allows* it...

src/NodeCompiler/CompileNodeToValue.php
muglug: I'll change to a `@var` docblock

src/Reflection/ReflectionFunctionAbstract.php
muglug: http://php.net/manual/en/reflector.export.php

muglug: If the parent were user-defined, a `void` return type would be a fatal error: https://3v4l.org/BXtvR

src/SourceLocator/Type/FileIteratorSourceLocator.php
muglug: I don't think so? But it's a logical syntax for an iterator that only contains `SplFileInfo`.

The union type that works in PHPStorm - `\Iterator|\SplFileInfo[]` - is *illogical*, since it implies the property can accept arrays.

kukulich: I agree it's logical but this syntax is not supported in editors and other tools (eg. PHPCS).

https://github.com/phpDocumentor/fig-standards/tree/master/proposed is still only a draft.

muglug: Yup, and until people stop using broken docblocks in their code, there'll be less momentum to fix it.

I have a `@psalm-var` annotation for non-phpdoc-compatible types, I could add it there instead.

src/SourceLocator/Ast/FindReflectionsInTree.php
muglug: I've added an `@inheritDoc` annotation instead

muglug: Btw this issue is documented here: https://github.com/phpstan/phpstan/issues/729#issuecomment-361737775

src/Reflection/Adapter/ReflectionFunction.php
```diff
@@ -27,7 +27,7 @@ public function __construct(BetterReflectionFunction $betterReflectionFunction)
      */
     public static function export($name, $return = null)
     {
-        BetterReflectionFunction::export(...func_get_args());
+        return BetterReflectionFunction::export(...func_get_args());
```
asgrim: Hmm, I know there's been a couple of comments on these, but `::export` methods really do return nothing:

https://3v4l.org/uLmk8

There's certainly a discrepancy here (phpdocs vs reality), so which do we pick? What the docs/types say, or how it actually behaves? So far we've been obeying the real-world behaviour (which I'd favour, but is of course incompatible with static analysis...)

/cc @Ocramius

src/Reflection/ReflectionClass.php
```diff
@@ -350,7 +350,6 @@ function (ReflectionMethod $method) use ($filter) {
      */
     public function getImmediateMethods(?int $filter = null) : array
     {
-        /** @var \ReflectionMethod[] $methods */
```
asgrim: why removing this?

src/Reflection/ReflectionFunctionAbstract.php
```diff
@@ -515,6 +518,7 @@ public function setBodyFromClosure(Closure $newBody) : void
      */
     public function setBodyFromString(string $newBody) : void
     {
+        /** @var Node[] $this->node->stmts */
```
asgrim: I haven't checked, but does PhpStorm (or other IDEs) actually support this syntax? o_O also, can this not be inferred anyway? And finally, why are we documenting the type when it's not actually used anywhere later?

src/SourceLocator/Reflection/SourceStubber.php
```diff
@@ -106,10 +115,7 @@ private function addDocComment(BuilderAbstract $node, CoreReflector $reflection)
         }
     }
 
-    /**
-     * @param Class_|Interface_|Trait_ $classNode
-     */
-    private function addClassModifiers(Declaration $classNode, CoreReflectionClass $classReflection) : void
+    private function addClassModifiers(Class_ $classNode, CoreReflectionClass $classReflection) : void
```
asgrim: Is `ClassLike` not more appropriate here? As this would no longer accept `Interface_` or `Trait_` I think, which I believe should be accepted so we can make them `abstract` (unless I'm mistaken)

src/SourceLocator/Reflection/SourceStubber.php
```diff
@@ -64,11 +64,20 @@ public function __invoke(CoreReflectionClass $classReflection) : string
     {
         $classNode = $this->createClass($classReflection);
 
-        $this->addClassModifiers($classNode, $classReflection);
+        if ($classNode instanceof Class_) {
+            $this->addClassModifiers($classNode, $classReflection);
```
asgrim: `Interface_` and `Trait_` should be passed here too

src/SourceLocator/Type/FileIteratorSourceLocator.php
asgrim: Yeah, until this can be fixed "properly", at least we have editor support with the `\SplFileInfo[]` hint, which is preferable to not figuring out types IMO :/

src/SourceLocator/Ast/FindReflectionsInTree.php
asgrim: Yeah, ok - makes sense to drop these at least for now :+1:

src/Reflection/Adapter/ReflectionFunction.php
Ocramius: Reality would be better overall.

src/Reflection/Adapter/ReflectionMethod.php
```diff
@@ -34,7 +34,7 @@ public function __construct(BetterReflectionMethod $betterReflectionMethod)
      */
     public static function export($class, $name, $return = null)
     {
-        BetterReflectionMethod::export(...func_get_args());
+        return BetterReflectionMethod::export(...func_get_args());
```
Ocramius: Same here: to be reverted

src/Reflection/Adapter/ReflectionMethod.php
```diff
@@ -313,7 +313,7 @@ public function isDestructor()
     public function getClosure($object = null)
     {
         try {
-            $this->betterReflectionMethod->getClosure($object);
+            return $this->betterReflectionMethod->getClosure($object);
```
Ocramius: Looks like we never tested this - that's needed here

src/Reflection/Adapter/ReflectionObject.php
```diff
@@ -175,7 +175,9 @@ public function hasProperty($name)
     }
 
     /**
-     * {@inheritDoc}
+     * @param string $name
```
Ocramius: We shouldn't change the parent API in the `Adapter` ns

src/Reflection/Adapter/ReflectionClass.php
```diff
@@ -184,7 +184,9 @@ public function hasProperty($name)
     }
 
     /**
-     * {@inheritDoc}
+     * @param string $name
```
Ocramius: We shouldn't change the documented API in the `Adapter` ns

src/Reflection/ReflectionClassConstant.php
```diff
@@ -188,7 +188,7 @@ public function __toString() : string
     /**
      * {@inheritDoc}
      */
-    public static function export() : void
+    public static function export()
```
Ocramius: ` : void` should be kept here

src/Reflection/ReflectionFunctionAbstract.php
Ocramius: No, this syntax is not supported

src/Reflection/ReflectionParameter.php
```diff
@@ -78,7 +78,10 @@ private function __construct()
     {
     }
 
-    public static function export() : void
+    /**
+     * {@inheritDoc}
+     */
+    public static function export()
```
Ocramius: `export()` is indeed ` : void`

src/Reflection/ReflectionProperty.php
```diff
@@ -71,7 +71,10 @@ private function __construct()
     {
     }
 
-    public static function export() : void
+    /**
+     * {@inheritDoc}
+     */
+    public static function export()
```
Ocramius: `export()` is indeed ` : void`

src/SourceLocator/Ast/FindReflectionsInTree.php
```diff
@@ -87,7 +87,10 @@ public function __construct(
...
+            /**
+             * {@inheritDoc}
+             */
+            public function enterNode(Node $node)
```
Ocramius: ` : void` is to be kept - would highlight changes in the parent API.

src/SourceLocator/Ast/FindReflectionsInTree.php
```diff
@@ -115,7 +118,10 @@ public function enterNode(Node $node) : void
                 }
             }
 
-            public function leaveNode(Node $node) : void
+            /**
+             * {@inheritDoc}
+             */
+            public function leaveNode(Node $node)
```
Ocramius: ` : void` is to be kept - would highlight changes in the parent API.

src/SourceLocator/Type/AnonymousClassObjectSourceLocator.php
```diff
@@ -123,7 +123,10 @@ public function __construct(string $fileName, int $startLine)
                 $this->startLine = $startLine;
             }
 
-            public function enterNode(Node $node) : void
+            /**
+             * {@inheritDoc}
+             */
+            public function enterNode(Node $node)
```
Ocramius: ` : void` is to be kept - would highlight changes in the parent API.

src/SourceLocator/Type/ClosureSourceLocator.php
```diff
@@ -112,7 +112,10 @@ public function __construct(string $fileName, int $startLine)
                 $this->startLine = $startLine;
             }
 
-            public function enterNode(Node $node) : void
+            /**
+             * {@inheritDoc}
+             */
+            public function enterNode(Node $node)
```
Ocramius: ` : void` is to be kept - would highlight changes in the parent API.

src/SourceLocator/Type/ClosureSourceLocator.php
```diff
@@ -125,7 +128,10 @@ public function enterNode(Node $node) : void
                 }
             }
 
-            public function leaveNode(Node $node) : void
+            /**
+             * {@inheritDoc}
+             */
+            public function leaveNode(Node $node)
```
Ocramius: ` : void` is to be kept - would highlight changes in the parent API.

src/SourceLocator/Type/FileIteratorSourceLocator.php
Ocramius: Syntax change indeed to be reverted here.

src/TypesFinder/PhpDocumentor/NamespaceNodeToReflectionTypeContext.php
```diff
@@ -35,17 +35,28 @@ public function __invoke(?Namespace_ $namespace) : Context
     private function aliasesToFullyQualifiedNames(Namespace_ $namespace) : array
     {
         // flatten(flatten(map(stuff)))
-        return array_merge([], ...array_merge([], ...array_map(function ($use) : array {
-            /** @var $use Use_|GroupUse */
+        return array_merge(
```
Ocramius: I don't understand what was changed here, since the alignment was changed. The operation was simply turning `array[][]` into `array`, but I don't see the diff nodes affected here.

src/Reflection/Adapter/ReflectionFunction.php
muglug: I'm confused. It *can* return e.g. https://3v4l.org/oolCP

src/Reflection/ReflectionFunctionAbstract.php
muglug: I'll remove

> also, can this not be inferred anyway

The method can return null here

src/SourceLocator/Reflection/SourceStubber.php
muglug: How can we have protected/private interfaces or traits?

src/SourceLocator/Reflection/SourceStubber.php
muglug: Traits/Interfaces can't be `abstract` though? Or `final`?

src/SourceLocator/Type/FileIteratorSourceLocator.php
muglug: Yup will do

src/TypesFinder/PhpDocumentor/NamespaceNodeToReflectionTypeContext.php
muglug: Changing `/** @var $use Use_|GroupUse */` to `/** @param Use_|GroupUse $use */`

src/Reflection/ReflectionClass.php
muglug: Because it can be inferred easily

src/Reflection/Adapter/ReflectionClass.php
muglug: You already have, though - documented [here](https://github.com/Roave/BetterReflection/issues/389)

src/Reflection/Adapter/ReflectionObject.php
muglug: See https://github.com/Roave/BetterReflection/issues/389

muglug:
@ocramius w/r/t `enterNode`/`leaveNode`, do you think a static analysis tool should treat a `void` return as equal to a `null` one?

Currently PHPStan [doesn’t check](https://github.com/phpstan/phpstan/issues/729) return type equality or covariance of docblock return types, allowing you to do whatever you want.

I could add a workaround that allows `: void` when the upstream return type is only in a docblock (and nullable). That would acknowledge that, as far as PHP is concerned, the function returns `null`.

Once PHPStan starts using BetterReflection I imagine these issues will pop up there, too.

Ocramius:
@muglug ` : void` is covariant with ` : null` (if we ever had `null`), so I think that it is correct to allow it.

src/SourceLocator/Ast/FindReflectionsInTree.php
muglug: Will revert this change though, per https://github.com/Roave/BetterReflection/pull/394#issuecomment-361956061

muglug:
@Ocramius patched [here](https://github.com/vimeo/psalm/commit/de9b90297a5b1b74fd5bd7f1e02e14ce9b15133c)

muglug:
I've removed the unnecessary `return` statements in `export` static methods that always threw exceptions anyway.

src/Reflection/Adapter/ReflectionClass.php
Ocramius: Yeah, but that bug is being fixed. This needs revert or we can wait for the other patch to go in before reverting

src/Reflection/Adapter/ReflectionObject.php
Ocramius: same as discussion in https://github.com/Roave/BetterReflection/pull/394/files#r164996850

src/Reflection/ReflectionClass.php
Ocramius: `array_map()`'s documented return value is `array` - unless we have generics, this isn't inferred by tooling that doesn't assume implementation details of `array_map`, so the change should be reverted.

src/TypesFinder/PhpDocumentor/NamespaceNodeToReflectionTypeContext.php
Ocramius: Seems related to https://github.com/Roave/BetterReflection/pull/394/files#r164988430 - type is being inferred over implementation of `array_map()`. LGTM.

src/Reflection/Adapter/ReflectionMethod.php
Ocramius: Yeah, this needs a test addition too

src/NodeCompiler/CompileNodeToValue.php
```diff
@@ -130,6 +130,7 @@ private function compileConstFetch(Node\Expr\ConstFetch $constNode)
      */
     private function compileClassConstFetch(Node\Expr\ClassConstFetch $node, CompilerContext $context)
     {
+        /** @var Node\Name $node->class */
```
Ocramius: This syntax is not really compatible with most tooling out there - what should we do about it?

src/NodeCompiler/CompileNodeToValue.php
```diff
@@ -148,6 +149,7 @@ private function compileClassConstFetch(Node\Expr\ClassConstFetch $node, Compile
             $classInfo = $context->getReflector()->reflect($className);
         }
 
+        /** @var string $node->name */
```
Ocramius: Same question about compatibility: what to do here?

src/Reflection/ReflectionClass.php
Ocramius: Hmm, errata - it can be inferred, it's just that most tooling won't do recursive inferring of types, so this can't really be removed unless the ecosystem evolves

src/Reflection/Adapter/ReflectionObject.php
Ocramius: Handled in #391

src/Reflection/Adapter/ReflectionClass.php
muglug: Reverted

src/Reflection/Adapter/ReflectionObject.php
muglug: Reverted

src/Reflection/Adapter/ReflectionMethod.php
muglug: Test added in 74ba5922c6910cb7954b09703f5086c2b29357d7

src/Reflection/ReflectionClass.php
muglug: evolution

muglug: I'll revert this

muglug: Oh this was also inaccurate, which is why I made the change in the first place (it shouldn't have the leading slash)

src/NodeCompiler/CompileNodeToValue.php
muglug: If it breaks anything I'm happy to remove it, or replace with an `assert`.

My hope is that the PSR-5 PHPDoc proposal is approved, and the people who make similar tools add support for this. It's a common-sense annotation.

muglug: This is also that could be fixed with a Psalm config, as it relates to behaviour of PhpParser that will never occur given how it's used in BetterReflection. https://github.com/vimeo/psalm/issues/482 would fix that internally.

Anyway, using an `assert` also fixes a PHPStan error at the same location so maybe that's the way to go here.

src/TypesFinder/PhpDocumentor/NamespaceNodeToReflectionTypeContext.php
muglug: Not totally related. Psalm has a `MissingDocblockType` that triggers for `@var $var Type` annotations. When I had a look at this, it made sense to not only fix the order by change it to a `@param`.

Psalm doesn't project types into closures, so it's not able to infer this, but it does check the boundaries – e.g. it sees the error in

```php
class A {
  public function foo() : void {}
}
class B {
  public function foo() : void {}
}

$as = [new A(), new B()];

$classes = array_map(
  function(B $a) : string {
    $a->foo();
    return "hello";
  },
  $as
);
```

(I don't have link generation set up, but you can always paste into [getpsalm.org](https://getpsalm.org/) to verify)

src/Reflection/ReflectionClass.php
Ocramius: I like the pic :D

I know the tool you built is doing it, but the majority still needs to catch up ;-)

src/NodeCompiler/CompileNodeToValue.php
Ocramius: Let's keep them for now 👍

src/NodeCompiler/CompileNodeToValue.php
Ocramius: Let's keep this for now

src/Reflection/Adapter/ReflectionMethod.php
Ocramius: Thanks!

src/SourceLocator/Reflection/SourceStubber.php
Ocramius: No fix needed

src/SourceLocator/Reflection/SourceStubber.php
Ocramius: No fix needed

src/TypesFinder/PhpDocumentor/NamespaceNodeToReflectionTypeContext.php
Ocramius: 👍

test/unit/Reflection/Adapter/ReflectionMethodTest.php
```diff
@@ -67,7 +69,7 @@ public function methodExpectationProvider() : array
             ['isUserDefined', null, true, []],
             ['getClosureThis', NotImplemented::class, null, []],
             ['getClosureScopeClass', NotImplemented::class, null, []],
-            ['getDocComment', null, '', []],
+            ['getDocComment', null, false, []],
```
Ocramius: Nice catch here 👍

muglug:
@Ocramius totally worth it for the improvements it forced me to make to Psalm. Thanks for taking your time with it!

PR #397, opened 2018-02-02
Fix behaviour of ReflectionClass::export/ReflectionObject::export to match behaviour of core classes

muglug:
`Roave\BetterReflection\Reflection\ReflectionClass::export` and `Roave\BetterReflection\Reflection\ReflectionObject::export` would always `return` values. However, `ReflectionClass::export` and `ReflectionObject::export` echo values to stdout by default,.

This PR changes the default behaviour of the BetterReflection classes to match up with the core ones. Now the `return` functionality is hidden behind an explicit `bool $return` argument.

It also allows `ReflectionClass::export` to take an object instance, like its core equivalent.

### DIFF
--- a/src/Reflection/ReflectionClass.php
+++ b/src/Reflection/ReflectionClass.php
@@ -46,6 +46,7 @@ use function get_class;
 use function implode;
 use function in_array;
 use function is_object;
+use function is_string;
 use function ltrim;
 use function sha1;
 use function sprintf;
@@ -101,6 +102,8 @@ class ReflectionClass implements Reflection, CoreReflector
 
     /**
      * Create a reflection and return the string representation of a named class
+     *
+     * @param string|object $argument
      */
     public static function export($argument, bool $return = false) : ?string
     {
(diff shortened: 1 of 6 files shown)


theofidry:
ping @asgrim

asgrim:
@Ocramius @theofidry indeed the `export` functions were written before we wrote the `Adapter` namespace, I'd also favour fixing/moving that behaviour into just the adapters and deprecating/removing the `export` functions from the Better Reflections.

With that said, this change looks good, but only if we can remove it from the main BR reflections and just do it in `export`. May be some scope difficulty to overcome, but I'm sure we can find our way aroudn things :D

asgrim:
I mean - we can implement the `export` methods in the adapters and just leave the actual BR implementations as exception throwing, that would work, just leaves an annoying unusable function lying aroud meh.

Ocramius:
@asgrim what about introducing a BC Break where we move the `Reflector` interface from the `BetterReflection\Reflection\*` instances (and only keep it in `BetterReflection\Reflection\Adapter\*`)?

asgrim:
@Ocramius actually yes that seems like a reasonable approach, it's a BC break regardless, so I'm +1 on that. I'll see if I can get to that at some point!

asgrim:
Only took like 4 months - opened a new PR in #479 as I don't think I can push to this one.

PR #400, opened 2018-02-03
PHPUnit & PHPStan updated

kukulich:


src/Reflection/ReflectionFunction.php
```diff
@@ -38,10 +41,13 @@ public static function createFromClosure(Closure $closure) : self
     {
         $configuration = new BetterReflection();
 
-        return (new FunctionReflector(
+        /** @var self $reflectionFunction */
```
muglug: I think you can revert this change and instead update the docblock of `FunctionReflector::reflect` instead to reflect that it always returns a `ReflectionFunction`. PHPStan [allows that syntax](https://phpstan.org/r/43781f0e8344ccf014130042a3a232af).

src/Reflection/ReflectionFunctionAbstract.php
```diff
@@ -75,7 +75,7 @@ public static function export() : void
     /**
      * Populate the common elements of the function abstract.
      *
-     * @param Node\Stmt\ClassMethod|Node\FunctionLike|Node\Stmt|Node $node Node has to be processed by the PhpParser\NodeVisitor\NameResolver
+     * @param Node\Stmt\ClassMethod|Node\Stmt\Function_|Node\Expr\Closure $node Node has to be processed by the PhpParser\NodeVisitor\NameResolver
```
muglug: Also here: https://github.com/Roave/BetterReflection/pull/394/files#diff-a24d529e5ad68aa3369134d27a9d51f3R78

src/Reflection/ReflectionObject.php
```diff
@@ -94,7 +94,10 @@ public static function createFromInstance($object) : ReflectionClass
             $reflector = (new BetterReflection())->classReflector();
         }
 
-        return new self($reflector, $reflector->reflect($className), $object);
+        /** @var ReflectionClass $reflectionClass */
```
muglug: As above, can be in the docblock `ClassReflector::reflect` instead

src/Reflection/ReflectionFunction.php
kukulich: You,re true, thanks.

src/Reflection/ReflectionParameter.php
```diff
@@ -490,7 +490,10 @@ public function getClass() : ?ReflectionClass
             ));
         }
 
-        return $this->reflector->reflect($className);
+        /** @var ReflectionClass $reflectionClass */
```
muglug: With the above ClassReflector change, you could instead change the `@var` to `@var ClassReflector $reflector`

src/Reflection/ReflectionType.php
```diff
@@ -25,7 +25,7 @@ class ReflectionType
     ];
 
     /**
-     * @var $type
+     * @var string
```
muglug: Here as well: https://github.com/Roave/BetterReflection/pull/394/files#diff-715bd899956ff9c25ee4bc334a997cacR28

src/NodeCompiler/CompileNodeToValue.php
```diff
@@ -140,15 +140,15 @@ private function compileClassConstFetch(Node\Expr\ClassConstFetch $node, Compile
         $classInfo = null;
 
         if ($className === 'self' || $className === 'static') {
-            $classInfo = $this->getConstantDeclaringClass($node->name, $context->getSelf());
+            $classInfo = $this->getConstantDeclaringClass((string) $node->name, $context->getSelf());
```
muglug: This cast is unnecessary as `$node->name` will always be a string here. Maybe an `assert(is_string($node->name))` above would be better, and would make the other cast below unnecessary

Ocramius:
@kukulich rebase needed as per #394 changes

kukulich:
@Ocramius Done.

Ocramius:
@kukulich thanks!

Issue #398, opened 2018-02-03
Trait method aliases


lookyman:
Aliasing trait methods doesn't seem to work:

```php
trait Foo
{
	public function a() {}
}

trait Bar
{
	use Foo { a as b; }
	public function a() { $this->b(); }
}

class Baz
{
	use Bar;
	public function c()
	{
		$this->a();
		$this->b();
	}
}

$refN = new \ReflectionClass('Bazz');
var_dump($refN->hasMethod('b')); // true

$refB = (new \Roave\BetterReflection\BetterReflection())->classReflector()->reflect('Bazz');
var_dump($refB->hasMethod('b')); // false

$refA = new \Roave\BetterReflection\Reflection\Adapter\ReflectionClass($refB);
var_dump($refA->hasMethod('b')); // false
```

kukulich:
Fix in https://github.com/Roave/BetterReflection/pull/401

Ocramius:
Handled in #401

PR #401, opened 2018-02-04
Method from trait renamed with alias should be available by the alias 

kukulich:
Fix for https://github.com/Roave/BetterReflection/issues/398.


Ocramius:
This requires backporting to 2.x too

Ocramius:
Backported to `2.0.x` via efc45b50cb52d5eeaacab15741376e981e28738b

Thanks @kukulich!

PR #404, opened 2018-02-15
Account for case insensitivity

theofidry:
The PR is a bit incomplete (missing tests) but it's more for myself to not forget about it :)

The issue is that [PHP is case insensitive](https://3v4l.org/SYNgh) for namespaces, classes and methods. However BetterReflection does not account for that as a result if you have:

```php
use PHAR;
```

Instead of:

```php
use Phar;
```

BetterReflection is going to fail.

### DIFF
--- a/src/SourceLocator/Ast/Locator.php
+++ b/src/SourceLocator/Ast/Locator.php
@@ -86,8 +86,10 @@ class Locator
      */
     private function findInArray(array $reflections, Identifier $identifier) : Reflection
     {
+        $identifierName = strtolower($identifier->getName());
+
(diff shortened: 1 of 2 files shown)


asgrim:
Sorry, this pull request could not be found.

theofidry:
Stop looking at your phone James

theofidry:
Yes sorry, I won't be able to take a look before another 1-2 weeks :/

theofidry:
Tests added. The CI is failure is unrelated (it's failing on master as well)

Ocramius:
:ship: thanks @theofidry!

PR #415, opened 2018-04-16
Improved `ReflectionType` API so that a target reflection class can be found, removed existing named constructor

Ocramius:
Required for https://github.com/Roave/ApiCompare/pull/38

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -4,6 +4,13 @@ This document serves as a reference to upgrade your current
 BetterReflection installation if improvements, deprecations
 or backwards compatibility (BC) breakages occur.
 
+## 4.0.0
+
+### BC breaks
+
+* Method `Roave\BetterReflection\Reflection\ReflectionType::createFromType()` has been removed,
+do use `Roave\BetterReflection\Reflection\ReflectionType::createFromTypeAndReflector()` instead.
+
 ## 3.0.0
 
 ### BC breaks
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -4,17 +4,12 @@ This document serves as a reference to upgrade your current
 BetterReflection installation if improvements, deprecations
 or backwards compatibility (BC) breakages occur.
 
-## 4.0.0
+## 3.0.0
 
 ### BC breaks
 
 * Method `Roave\BetterReflection\Reflection\ReflectionType::createFromType()` has been removed,
 do use `Roave\BetterReflection\Reflection\ReflectionType::createFromTypeAndReflector()` instead.
-
-## 3.0.0
-
-### BC breaks
-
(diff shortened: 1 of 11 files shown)


src/Reflection/Exception/ClassDoesNotExist.php
```diff
@@ -4,8 +4,14 @@
 
 namespace Roave\BetterReflection\Reflection\Exception;
 
+use Roave\BetterReflection\Reflection\Reflection;
 use RuntimeException;
+use function sprintf;
```
roukmoute: Is it really needed?
Tests passes without this.

Ocramius: Gonna likely propose the adoption of `doctrine/coding-standard` here, which enforces all functions to be imported to avoid static analysis issues, as well as runtime function lookups.

Ocramius: Hmm, just seen that there already is an open pull request for that.

roukmoute: Ok thanks :)
It's interesting point of view.

For `to avoid static analysis issues` is it not a problem from static analysis?
And for `function lookups` I don't understand where is the problem with that.
Have you an example? To help me to understand where it could be a problem.

theofidry: if you call a function `foo()` you have no idea if it's a function declared in the same namespace or a global function. PHP falls back on the global scope for functions and constants whereas you never have the ambiguity with classes: you either need to use the FQCN e.g. `\Closure` or import the symbol with a use statement `use Closure`

UPGRADE.md
```diff
@@ -4,6 +4,13 @@ This document serves as a reference to upgrade your current
 BetterReflection installation if improvements, deprecations
 or backwards compatibility (BC) breakages occur.
 
+## 4.0.0
```
kukulich: Is there a reason why not to make it part of version 3?

Ocramius: I just thought `3.0.0` was already released: did this in a real rush, so will amend :+1:

src/Reflection/Exception/ClassDoesNotExist.php
Ocramius: > For `to avoid static analysis issues` is it not a problem from static analysis?

an example I have is this one:

```php
namespace Potato;

define('POTATO', 'POTATO');
```

Is `Potato::POTATO` defined here? Or is `define()` a user-defined function?

You can only be sure of it if it was written this way: 


```php
namespace Potato;

use function define;

define('POTATO', 'POTATO');
```

I hope that clears it up.

asgrim:
Apart from whatever is causing failed build, looks ok at a glance

Ocramius:
@asgrim all green here :+1:

PR #420, opened 2018-05-04
Tidied docs and simply removed weasel words

GeeH:
- Removed weasel words like "simply", "just" and "easily"
- Reformatted to be 120 chars per line.

Pick your way through that 😆


asgrim:
uhh build failures are not related >.< fml

PR #423, opened 2018-05-11
Add Psalm to build and fix issues found

muglug:
This adds Psalm to BetterReflection's build, fixing issues it found (mainly by adding supplementary docblocks).

If you'd rather those docblocks be `assert`s, I'm happy to update.

### DIFF
--- a/src/Reflection/ReflectionClass.php
+++ b/src/Reflection/ReflectionClass.php
@@ -123,6 +123,8 @@ class ReflectionClass implements Reflection, CoreReflector
      * @throws IdentifierNotFound
      * @throws \ReflectionException
      * @throws \InvalidArgumentException
+     *
+     * @psalm-suppress DocblockTypeContradiction
      */
     public static function createFromInstance($instance) : self
     {
@@ -166,6 +168,7 @@ class ReflectionClass implements Reflection, CoreReflector
     public function getShortName() : string
     {
         if (! $this->isAnonymous()) {
+            /** @var \PhpParser\Node\Identifier $this->node->name */
             return $this->node->name->name;
         }
 
@@ -194,6 +197,8 @@ class ReflectionClass implements Reflection, CoreReflector
     /**
      * Get the "namespace" name of the class (e.g. for A\B\Foo, this will
      * return "A\B").
+     *
+     * @psalm-suppress PossiblyNullPropertyFetch
      */
     public function getNamespaceName() : string
     {
@@ -340,7 +345,7 @@ class ReflectionClass implements Reflection, CoreReflector
             array_filter(
                 $this->getMethodsIndexedByName(),
                 function (ReflectionMethod $method) use ($filter) {
-                    return $filter & $method->getModifiers();
+                    return (bool) ($filter & $method->getModifiers());
                 }
             )
         );
@@ -623,7 +628,7 @@ class ReflectionClass implements Reflection, CoreReflector
         return array_filter(
             $this->cachedImmediateProperties,
             function (ReflectionProperty $property) use ($filter) {
-                return $filter & $property->getModifiers();
+                return (bool) ($filter & $property->getModifiers());
             }
         );
     }
@@ -686,7 +691,7 @@ class ReflectionClass implements Reflection, CoreReflector
         return array_filter(
             $this->cachedProperties,
             function (ReflectionProperty $property) use ($filter) {
-                return $filter & $property->getModifiers();
+                return (bool) ($filter & $property->getModifiers());
             }
         );
     }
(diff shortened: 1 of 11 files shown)


src/Reflection/ReflectionClass.php
```diff
@@ -340,7 +345,7 @@ public function getMethods(?int $filter = null) : array
             array_filter(
                 $this->getMethodsIndexedByName(),
                 function (ReflectionMethod $method) use ($filter) {
-                    return $filter & $method->getModifiers();
+                    return (bool) ($filter & $method->getModifiers());
```
kukulich: The closure should have return type hint as well.

src/SourceLocator/Type/MemoizingSourceLocator.php
```diff
@@ -17,7 +17,7 @@
     /** @var SourceLocator */
     private $wrappedSourceLocator;
 
-    /** @var Reflection[] indexed by reflector key and identifier cache key */
+    /** @var (Reflection|null)[] indexed by reflector key and identifier cache key */
```
kukulich: This syntax is not supported by other tools (eg. PHPCS), please don't use it.

src/TypesFinder/ResolveTypes.php
```diff
@@ -30,6 +30,7 @@ public function __invoke(array $stringTypes, Context $context) : array
             $resolvedTypes[] = $this->typeResolver->resolve($stringType, $context);
         }
 
+        /** @var Type[] */
```
kukulich: This doccomment should be on line 27 and contain variable name.

Ocramius:
Note: this is to be taken over by the @Roave folks, as I just asked @muglug to open the PR with his current patch

src/TypesFinder/ResolveTypes.php
muglug: I did one of those for-the-typechecker things where I told it that the array only gets non-nullable objects in the `foreach`

src/Reflection/ReflectionClass.php
```diff
@@ -685,8 +690,8 @@ function (ReflectionClass $trait) use ($filter) {
 
         return array_filter(
             $this->cachedProperties,
-            function (ReflectionProperty $property) use ($filter) {
-                return $filter & $property->getModifiers();
+            function (ReflectionProperty $property) use ($filter) : bool {
```
carusogabriel: This should be founded with CS :(

kukulich: It's difficult. There are no annotations.

prisis:
Why did you choose Psalm and not phpstan?

muglug:
https://github.com/vimeo/psalm/graphs/contributors

asgrim:
@prisis we had `phpstan` before, but it's incompatible with PhpParser 4, `psalm` is a great tool also :+1:

kukulich:
PHPStan will return :) It is disabled only temporarily https://github.com/Roave/BetterReflection/commit/31129a702cb15553b7260aca8df16e1d2a181d2b

muglug:
Also because of https://twitter.com/Ocramius/status/994684609928540161

But now PHPStan supports 4.0 in `dev-master`, this PR may well be extraneous.

Ocramius:
> this PR may well be extraneous.

No, I think this PR is a good fit. This is an extremely clean library (not the best architecture, but we put some effort in it), and it values types over everything else, so having multiple type checkers is beneficial both for the library and for the authors of the type checkers.

src/SourceLocator/Type/ClosureSourceLocator.php
```diff
@@ -156,6 +156,7 @@ public function getClosureNodes() : ?array
         $nodeTraverser->addVisitor($nodeVisitor);
         $nodeTraverser->traverse($ast);
 
+        /** @var array $closureNodes */
```
Ocramius: Should this be `Node[]`?

src/SourceLocator/Type/AnonymousClassObjectSourceLocator.php
```diff
@@ -47,6 +47,8 @@
      *
      * @throws \InvalidArgumentException
      * @throws \ReflectionException
+     *
+     * @psalm-suppress DocblockTypeContradiction
```
Ocramius: Adding custom annotations is a bit risky due to various annotation parsers. @muglug what do you think about these? Is this kind of stuff something that `psalm.xml` can handle at all?

src/Reflection/ReflectionClass.php
Ocramius: We're basically missing a `mixed` return type (in the language) in order to have CS here too...

PR #425, opened 2018-05-29
Better exception messages and assertions for unrecognized class constant fetch expressions

Ocramius:
This basically replaces a null pointer error with a nicer exception with some
minimal context/description around it.

### DIFF
--- a/src/NodeCompiler/CompileNodeToValue.php
+++ b/src/NodeCompiler/CompileNodeToValue.php
@@ -74,7 +74,7 @@ class CompileNodeToValue
                 return true;
             default:
                 if (! defined($firstName)) {
-                    throw Exception\UnableToCompileNode::becauseOfOfNotFoundConstantReference($context, $constNode);
+                    throw Exception\UnableToCompileNode::becauseOfNotFoundConstantReference($context, $constNode);
                 }
 
                 return constant($firstName);
--- a/src/NodeCompiler/CompileNodeToValue.php
+++ b/src/NodeCompiler/CompileNodeToValue.php
@@ -112,7 +112,6 @@ class CompileNodeToValue
             $classInfo = $context->getReflector()->reflect($className);
         }
 
-        /** @var ReflectionClassConstant $reflectionConstant */
         $reflectionConstant = $classInfo->getReflectionConstant($nodeName);
 
         if (! $reflectionConstant instanceof ReflectionClassConstant) {
--- a/src/NodeCompiler/CompileNodeToValue.php
+++ b/src/NodeCompiler/CompileNodeToValue.php
@@ -87,7 +87,7 @@ class CompileNodeToValue
      * @return string|int|float|bool|mixed[]|null
      *
      * @throws IdentifierNotFound
-     * @throws Exception\UnableToCompileNode if a referenced constant could not be located on the expected referenced class
+     * @throws Exception\UnableToCompileNode If a referenced constant could not be located on the expected referenced class.
      */
     private function compileClassConstFetch(Node\Expr\ClassConstFetch $node, CompilerContext $context)
     {
--- a/src/NodeCompiler/CompileNodeToValue.php
+++ b/src/NodeCompiler/CompileNodeToValue.php
@@ -13,10 +13,8 @@ use Roave\BetterReflection\Util\FileHelper;
 use function constant;
 use function defined;
 use function dirname;
-use function get_class;
 use function realpath;
 use function reset;
-use function sprintf;
 
 class CompileNodeToValue
 {
@@ -37,7 +35,7 @@ class CompileNodeToValue
 
         $constExprEvaluator = new ConstExprEvaluator(function (Node\Expr $node) use ($context) {
             if ($node instanceof Node\Expr\ConstFetch) {
-                return $this->compileConstFetch($node);
+                return $this->compileConstFetch($node, $context);
             }
 
             if ($node instanceof Node\Expr\ClassConstFetch) {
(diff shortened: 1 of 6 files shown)


src/NodeCompiler/Exception/UnableToCompileNode.php
```diff
@@ -5,7 +5,56 @@
...
+            self::compilerContextToContextDescription($fetchContext),
+            $constantFetch->getLine()
+        ));
+    }
+
+    private static function compilerContextToContextDescription(CompilerContext $fetchContext) : string
+    {
+        return $fetchContext->hasSelf()
+            ? $fetchContext->getSelf()->getName()
+            : 'unknown context'; // @TODO review feedback here, plox!
```
Ocramius: This is probably a problem

asgrim: If the context is in a function?

src/NodeCompiler/CompileNodeToValue.php
```diff
@@ -52,7 +50,7 @@ public function __invoke(Node $node, CompilerContext $context)
                 return $this->compileClassConstant($context);
             }
 
-            throw new Exception\UnableToCompileNode('Unable to compile expression: ' . get_class($node));
```
asgrim: This is dreadful and you should feel bad.

Ocramius: IIRC that's your code...

src/NodeCompiler/Exception/UnableToCompileNode.php
```diff
@@ -5,7 +5,56 @@
...
 use LogicException;
+use PhpParser\Node;
+use Roave\BetterReflection\NodeCompiler\CompilerContext;
+use Roave\BetterReflection\Reflection\ReflectionClass;
+use function get_class;
+use function sprintf;
 
 class UnableToCompileNode extends LogicException
 {
+    public static function forUnRecognizedExpressionInContext(Node\Expr $expression, CompilerContext $context) : self
```
BackEndTea: The other two methods are named  `becauseOf<thing>`, but this method doesn't follow that same pattern. 
How about changing the name of this method to follow the same pattern?

Ocramius: Could do, but at the call-site it reads better this way, so I'd prefer better readability from a consumer perspective than from a consistency across multiple named constructors

src/NodeCompiler/Exception/UnableToCompileNode.php
```diff
@@ -5,7 +5,56 @@
...
+        return new self(sprintf(
+            'Could not locate constant %s::%s while trying to evaluate constant expression in %s at line %s',
+            $targetClass->getName(),
+            $constantFetch->name->name,
+            self::compilerContextToContextDescription($fetchContext),
+            $constantFetch->getLine()
+        ));
+    }
+
+    public static function becauseOfOfNotFoundConstantReference(
```
djneo92nl: Is this the best name for this method, it is pretty similar to "becauseOfNotFoundClassConstantReference"

Ocramius: Constant and ClassConstant are different concepts both for this library and for PHP-Parser

djneo92nl: never mind then 👼

Ocramius:
@asgrim before, you'd get a `Throwable` due to a null pointer: now you get a specific exception. Code that caught `Throwable` will therefore continue to run.

asgrim:
@Ocramius that's my point, you'd get an `Error` before, now you get something that extends `Exception`, right? if someone has `catch (\Error $blah)` it may break. I think it's fairly unlikely, so we can probably just ignore.

Ocramius:
Yeah, I don't think that catching that would make sense, plus we really noticed the problem due to reports about this precisely. I will still document the change when tagging though.

asgrim:
Oh build is failing?

asgrim:
Looks like some CS issues to fix, I'll push the fixes  :)

Ocramius:
Thanks!

src/Reflection/ReflectionFunctionAbstract.php
```diff
@@ -557,6 +557,10 @@ public function removeParameter(string $parameterName) : void
         $lowerName = strtolower($parameterName);
 
         foreach ($this->node->params as $key => $paramNode) {
+            if ($paramNode->var instanceof Node\Expr\Error) {
```
asgrim: Not a fan of doing this, but it's the only way I could force `psalm` to not think `$paramNode->var` is ever a `Node\Expr\Error`. I tried using `/** @var Node\Expr\Error $paramNode->var */` but it still complained that sometimes `name` property doesn't exist (if `var` is an `Error` node)... is this the right solution do you think @muglug ?

asgrim:
@Ocramius I would say apart from that static analysis ugly thing, this is ready to go. The patch (especially around Psalm as above) is required to make the build pass in other PRs like #431, #433 ... WDYT, good to merge or should we fix this better?

src/NodeCompiler/Exception/UnableToCompileNode.php
asgrim: `CompilerContext` works for functions - basically, if `$fetchContext->hasSelf()` is `false`, it's a function AFAIK

asgrim: Yes, we're missing which function we're in (if any) but as you say that's only useful for exceptions. Maybe make a new issue to improve on this and add that capability to `CompilerContext`?

Ocramius: I'll open an issue then. Can you meanwhile drop the `@TODO` blocks in this PR?

asgrim:
@Ocramius the `todo` is sorted, just that static analysis workaround thing, LMK WDYT plz

Ocramius:
:shit:

PR #427, opened 2018-05-30
[RFC] Pass ClassIdentifier to StubClassSourceLocator

covex-nn:
When i tested `roave/backward-compatibility-check` tool, i found that it cannot check a class, that extends some other class from `require-dev` deps of composer.json

This job https://travis-ci.com/covex-nn/test-bc-check/builds/74891794, for example, was failed with a wierd message `Provided node "Symfony\Component\Filesystem\Filesystem" is not class, but "interface"`

~~~With this PR that message would be not so confusing~~~

This PR will let `StubClassSourceLocator` know what `Identifier` it must "locate"

* If it will be instanceof `ClassIdentifier`, them it will be `class`
* Otherwise - it will be interface

### DIFF
--- a/src/Identifier/ClassIdentifier.php
+++ b/src/Identifier/ClassIdentifier.php
@@ -1,0 +1,9 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Roave\BetterReflection\Identifier;
+
+class ClassIdentifier extends Identifier
+{
+}
--- a/src/Reflection/ReflectionClass.php
+++ b/src/Reflection/ReflectionClass.php
@@ -21,6 +21,8 @@ use ReflectionClass as CoreReflectionClass;
 use ReflectionProperty as CoreReflectionProperty;
 use Reflector as CoreReflector;
 use Roave\BetterReflection\BetterReflection;
+use Roave\BetterReflection\Identifier\ClassIdentifier;
+use Roave\BetterReflection\Identifier\IdentifierType;
 use Roave\BetterReflection\Reflection\Exception\ClassDoesNotExist;
 use Roave\BetterReflection\Reflection\Exception\NoObjectProvided;
 use Roave\BetterReflection\Reflection\Exception\NotAClassReflection;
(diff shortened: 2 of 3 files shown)


Ocramius:
Hold yer horses!

This is probably because of jttps://github.com/Roave/BackwardCompatibilityCheck/issues/77

src/Reflection/Exception/NotAClassReflection.php
```diff
@@ -5,17 +5,25 @@
...
-
         if ($class->isTrait()) {
             $type = 'trait';
+        } elseif ($class->isInterface()) {
+            $type = 'interface';
+        } else {
+            $type = 'unknown';
+            try {
+                $class->getParentClass();
+            } catch (IdentifierNotFound $e) {
```
Ocramius: Doing this is a bit risky. Is there a way to avoid the operation at all, even if the exception message is a bit worse?

covex-nn:
May be a tool fail to check because of that issue number 77, i do not know. Also i found, that `require-dev` deps are just does not loaded in [LocateDependenciesViaComposer::__invoke](https://github.com/Roave/BackwardCompatibilityCheck/blob/master/src/LocateDependencies/LocateDependenciesViaComposer.php#L44) because of `$installer->setDevMode(false);`

But this issue is about wierd message =)

Ocramius:
> `$installer->setDevMode(false);`

This I did on purpose ;-)

src/Reflection/Exception/NotAClassReflection.php
covex-nn: I run `vendor\roave\backward-compatibility-check\bin\roave-backward-compatibility-check.php` with xdebug (and with `-dmemory_limit=2G`) and realized that you are right. Not found class became interface. So, there won't be any "unknown" and "try-catch" blocks. It is that issue 77. I will close PR

covex-nn:
A second try. A last one, i guess )

@Ocramius in that issue number 77 you said, that `The stubbing mechanism also has no way to determine if Foo and Bar are to be classes or interfaces`. But in `Roave\BetterReflection\Reflection\ReflectionClass::getParentClass` we already know that `$this->node->extends->toString()` is a class, not an interface, not a trait. Just because that class came from extends.

So, i created a new empty class `ClassIdentifier extends Identifier`, that will be passed to `StubClassSourceLocator`. And there it will be possible to determine what it should be `<?php namespace %s{interface %s {}}` or `<?php namespace %s{class %s {}}`

src/Identifier/ClassIdentifier.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Roave\BetterReflection\Identifier;
+
+class ClassIdentifier extends Identifier
```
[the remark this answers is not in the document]
covex-nn: Actualy, i just wanted to show, that there is a `way to determine if Foo and Bar are to be classes or interfaces`, this knowledge can come to `StubClassSourceLocator` from `Roave\BetterReflection\Reflection\ReflectionClass::getParentClass`. I cannot finish this PR, please close it after you read it

PR #435, opened 2018-06-26
Return of the PHPStan

kukulich:


src/NodeCompiler/CompileNodeToValue.php
```diff
@@ -33,6 +33,9 @@ public function __invoke(Node $node, CompilerContext $context)
             return $this($node->expr, $context);
         }
 
+        /** @var Node\Expr $node */
+        $node = $node;
```
Ocramius: Isn't a declaration enough?

src/SourceLocator/Ast/Strategy/NodeToReflection.php
```diff
@@ -36,6 +36,8 @@ public function __invoke(
         }
 
         if ($node instanceof Node\FunctionLike) {
+            /** @var Node\Stmt\ClassMethod|Node\Stmt\Function_|Node\Expr\Closure $node */
+            $node = $node;
```
Ocramius: Same: why is this being fone with an assignment? Also, should the instanceof check verify these 3 types?

src/NodeCompiler/CompileNodeToValue.php
kukulich: Fixed with `@param Node\Stmt\Expression|Node\Expr $node` now.

src/SourceLocator/Ast/Strategy/NodeToReflection.php
kukulich:
```
        if ($node instanceof Node\Stmt\ClassMethod
            || $node instanceof Node\Stmt\Function_
            || $node instanceof Node\Expr\Closure
        ) {
```

You're right, this is definitely better.

PR #439, opened 2018-07-27
Adapter's ReflectionClass::getConstructor() returns null when no constructor exists

lookyman:
As per https://secure.php.net/manual/en/reflectionclass.getconstructor.php. If someone can guide me on how to write a test for this, I'll be happy to do it.

### DIFF
--- a/src/Reflection/Adapter/ReflectionClass.php
+++ b/src/Reflection/Adapter/ReflectionClass.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Roave\BetterReflection\Reflection\Adapter;
 
+use OutOfBoundsException;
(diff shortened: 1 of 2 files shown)


Ocramius:
@lookyman look at the existing adapter tests in https://github.com/Roave/BetterReflection/blob/ca0298b9ba44e54c66311cb47a9e558f7d9e6fcc/test/unit/Reflection/Adapter/ReflectionClassTest.php

What you want is something like:

```php
$class = $this->createMock(ReflectionClass::class);

$class->expects(self::any())->method('getConstructor')->willThrowException(...);
```

Ocramius:
Thanks @lookyman! Preparing a release right now :-)

PR #448, opened 2018-10-10
removed the returnvalue of the inherited parse method

gvanbeck:


Ocramius:
Can you please adapt the version of php-parser that this is targeting in `composer.json`? If there was an upstream BC break, we need to report it too.

gvanbeck:
Yes, here is the composer json:
{
    "name": "nikic/php-parser",
    "description": "A PHP parser written in PHP",
    "keywords": ["php", "parser"],
    "type": "library",
    "license": "BSD-3-Clause",
    "authors": [
        {
            "name": "Nikita Popov"
        }
    ],
    "require": {
        "php": ">=7.0",
        "ext-tokenizer": "*"
    },
    "require-dev": {
        "phpunit/phpunit": "^6.5 || ^7.0"
    },
    "autoload": {
        "psr-4": {
            "PhpParser\\": "lib/PhpParser"
        }
    },
    "bin": ["bin/php-parse"],
    "extra": {
        "branch-alias": {
            "dev-master": "4.1-dev"
        }
    }
}

composer.json
```diff
@@ -1,5 +1,5 @@
 {
-    "name": "roave/better-reflection",
+    "name": "gvanbeck/better-reflection",
```
asgrim: Why are you renaming this...?

src/Reflection/ReflectionClass.php
```diff
@@ -168,8 +168,7 @@ public static function createFromNode(
     public function getShortName() : string
     {
         if (! $this->isAnonymous()) {
-            /** @var Node\Identifier $this->node->name */
-            return $this->node->name->name;
+            return (string)$this->node->name;
```
asgrim: Can you make this a separate PR for this change please? Also what does this change benefit us by?

src/Reflection/ReflectionFunctionAbstract.php
```diff
@@ -359,15 +363,15 @@ public function isGenerator() : bool
      */
     public function getStartLine() : int
     {
-        return $this->node->getStartLine();
+        return $this->node->getAttributes()['startLine'] ?? -1;
```
asgrim: Why is this changing to not use the API given?

src/Reflection/ReflectionProperty.php
```diff
@@ -169,7 +169,11 @@ public function getModifiers() : int
      */
     public function getName() : string
     {
-        return $this->node->props[$this->positionInNode]->name->name;
+        if(is_object($this->node->props[$this->positionInNode]->name)) {
```
asgrim: Again, not sure what the benefit of this is? Could you explain please?

src/SourceLocator/Ast/Parser/MemoizingParser.php
```diff
@@ -30,7 +30,7 @@ public function __construct(Parser $wrappedParser)
     /**
      * {@inheritDoc}
      */
-    public function parse(string $code, ?ErrorHandler $errorHandler = null) : ?array
+    public function parse($code, ?ErrorHandler $errorHandler = null)
```
asgrim: What is the signature of the parent here? Why is it complaining about incompatibility exactly? Was there an upstream change that broke this?

src/Util/GetFirstDocComment.php
```diff
@@ -14,7 +14,8 @@
 {
     public static function forNode(NodeAbstract $node) : string
     {
-        foreach ($node->getComments() as $comment) {
+        $comments = $node->getAttributes()['comments'] ?? [];
```
asgrim: Why changing away from the given API?

composer.json
gvanbeck: Please I was not meaning to do pullrequests. Please decline them all for today

Sorry to bother

Ocramius:
Moved to https://github.com/Roave/BetterReflection/issues/450

Closing as `invalid` here, since the patch has unrelated/invalid changes.

PR #452, opened 2018-10-18
Upgrade to Doctrine CS 5.0

Majkl578:


.gitignore
```diff
@@ -1,3 +1,4 @@
-vendor
-.idea
-composer.lock
+/.phpcs-cache
+/phpcs.xml
+/vendor
+/.idea
```
Ocramius: Conflict due to this

.travis.yml
```diff
@@ -69,7 +70,7 @@ jobs:
       php: 7.1
```
Ocramius: Please run the CS with the newer PHP 7.2 version

Ocramius: We can actually just bump the dependency: no biggie

Issue #440, opened 2018-08-14
Test failures under PHP 7.3


Majkl578:
The following errors & failures are reported under PHP 7.3:
```
There were 10 errors:

1) Roave\BetterReflectionTest\Reflection\ReflectionParameterTest::testGetDocBlockTypeStrings
TypeError: array_filter() expects parameter 1 to be array, array given

/tmp/BetterReflection/src/TypesFinder/PhpDocumentor/NamespaceNodeToReflectionTypeContext.php:75
/tmp/BetterReflection/src/TypesFinder/PhpDocumentor/NamespaceNodeToReflectionTypeContext.php:56
/tmp/BetterReflection/src/TypesFinder/PhpDocumentor/NamespaceNodeToReflectionTypeContext.php:28
/tmp/BetterReflection/src/TypesFinder/FindParameterType.php:48
/tmp/BetterReflection/src/Reflection/ReflectionParameter.php:347
/tmp/BetterReflection/src/Reflection/ReflectionParameter.php:330
/tmp/BetterReflection/test/unit/Reflection/ReflectionParameterTest.php:194

2) Roave\BetterReflectionTest\Reflection\ReflectionParameterTest::testGetDocBlockTypes
TypeError: array_filter() expects parameter 1 to be array, array given

/tmp/BetterReflection/src/TypesFinder/PhpDocumentor/NamespaceNodeToReflectionTypeContext.php:75
/tmp/BetterReflection/src/TypesFinder/PhpDocumentor/NamespaceNodeToReflectionTypeContext.php:56
/tmp/BetterReflection/src/TypesFinder/PhpDocumentor/NamespaceNodeToReflectionTypeContext.php:28
/tmp/BetterReflection/src/TypesFinder/FindParameterType.php:48
/tmp/BetterReflection/src/Reflection/ReflectionParameter.php:347
/tmp/BetterReflection/test/unit/Reflection/ReflectionParameterTest.php:207

3) Roave\BetterReflectionTest\Reflection\ReflectionPropertyTest::testGetDocBlockTypeStrings with data set #0 ('privateProperty', array('int', 'float', '\stdClass'))
TypeError: array_filter() expects parameter 1 to be array, array given

/tmp/BetterReflection/src/TypesFinder/PhpDocumentor/NamespaceNodeToReflectionTypeContext.php:75
/tmp/BetterReflection/src/TypesFinder/PhpDocumentor/NamespaceNodeToReflectionTypeContext.php:56
[...]

PR #454, opened 2018-11-14
Allow installing with php 7.3

hanneskod:
TravisCI now supports php 7.3 RC, but testing projects that require BetterReflection against 7.3 is not possible with the current version requirement in composer.json. If you would like to consider a change it would be much appreciated.


pxgamer:
I guess PHP 7.3 will also need to be added to the Travis configuration so that the unit tests can be checked? 

See: #440 🤔

PR #464, opened 2019-04-22
AppVeyor is dead, long live Travis

kukulich:


Ocramius:
Interesting: build failing (probably) because GIT is checking out everything with the wrong CRLF flags

kukulich:
@Ocramius It should be fixed by https://github.com/Roave/BetterReflection/pull/464/commits/5b56d96be018ed6873159163dde0c3b64635f4a5#diff-fc723d30b02a4cca7a534518111c1a66R1

It works in my other repository. However it looks it does not work in PR :( The `.gitattributes` change has to be in master first probably.

Ocramius:
Ok, let's try this then. If needed, I'll revert.

Ocramius:
Looking good in `master` 👍

PR #466, opened 2019-04-24
Correct PHPUnit MockObject class names as per PHPUnit 8+

kukulich:
Broken since https://github.com/Roave/BetterReflection/commit/b9a4c135809d1295325e3e819b18414f42ac793e


Ocramius:
I wonder if we can add some minimal type-checking in the test suite?

kukulich:
I've tried to enable psalm and phpstan for unit tests but both report too much errors currently (psalm 384 , phpstan 484)

Ocramius:
Aye, I just went through it in https://github.com/Roave/BackwardCompatibilityCheck/pull/122, and it requires lots of work. Merging as-is here 👍

PR #470, opened 2019-05-08
PhpInternalSourceLocator is not based on internal reflection anymore

kukulich:


theofidry:
Will give it a shot on PHP-Scoper tonight or tomorrow

theofidry:
Works good on my side, thanks a lot @kukulich. For reference, here is the PR: https://github.com/humbug/php-scoper/pull/318

src/SourceLocator/SourceStubber/ReflectionSourceStubber.php
```diff
@@ -38,8 +38,12 @@
...
 use function array_key_exists;
+use function class_exists;
 use function explode;
+use function function_exists;
 use function in_array;
+use function interface_exists;
+use function trait_exists;
 
 /**
  * It generates a stub source from a given reflection instance.
```
theofidry: needs to be updated

bin/generate-phpstorm-stubs-map.php
```diff
@@ -0,0 +1,144 @@
...
+    foreach (new DirectoryIterator($directoryInfo->getPathName()) as $fileInfo) {
+        if ($fileInfo->isDot()) {
+            continue;
+        }
+
+        if (! preg_match('/\.php$/', $fileInfo->getBasename())) {
+            continue;
+        }
+
+        $ast = $phpParser->parse(file_get_contents($fileInfo->getPathname()));
```
theofidry: it is unlikely to fail but shouldn't its potential error be checked nonetheless?

bin/generate-phpstorm-stubs-map.php
```diff
@@ -0,0 +1,144 @@
...
+    ksort($typeFiles);
+    foreach ($typeFiles as $name => $file) {
+        $output .= sprintf('        \'%s\' => &$files[%s],%s', $name, $fileNumbers[$file], "\n");
+    }
+
+    $output .= "    ],\n";
+}
+$output .= "];\n";
+
+file_put_contents(__DIR__ . '/../src/SourceLocator/SourceStubber/PhpStormStubsMap.php', $output);
```
theofidry: Same here: the write might fail

src/SourceLocator/SourceStubber/PhpStormStubsMap.php
```diff
@@ -0,0 +1,4910 @@
+<?php
```
theofidry: Up to the Roave team but might be worth adding a `Makefile` with a rule to auto-generate that file?

src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
```diff
@@ -45,179 +42,163 @@ final class PhpStormStubsSourceStubber implements SourceStubber
...
 
-        $this->nodeTraverser->addVisitor($nodeVisitor);
+        return new StubData($this->createStub($this->functionNodes[$functionName]), $this->getExtensionFromFilePath($filePath));
+    }
 
-        foreach ($this->getExtensionStubsFiles($extensionName) as $filePath) {
-            FileChecker::assertReadableFile($filePath);
+    private function parseFile(string $filePath) : void
+    {
+        $ast = $this->phpParser->parse(file_get_contents($this->getAbsoluteFilePath($filePath)));
```
theofidry: the read here might fail

src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
```diff
@@ -45,179 +42,163 @@ final class PhpStormStubsSourceStubber implements SourceStubber
...
+    private function createStub(Node $node) : string
+    {
         return "<?php\n\n" . $this->prettyPrinter->prettyPrint([$node]) . "\n";
     }
 
-    private function getClassNodeVisitor(CoreReflectionClass $classReflection) : NodeVisitorAbstract
+    private function createCachingVisitor() : NodeVisitorAbstract
     {
-        return new class($classReflection->getName()) extends NodeVisitorAbstract
+        return new class() extends NodeVisitorAbstract
```
theofidry: One might say it's not BetterReflection but this piece is a deal-breaker currently for Box. Indeed since it is an anonymous class, this object cannot be serialized which is causing issues when leveraging parallelisation in Box. Would it be possible to move it to a dedicated class? If the API stability is a concern, it could be annotated with a `@internal` tag

src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
kukulich: Fixed.

bin/generate-phpstorm-stubs-map.php
kukulich: The script is only for internal use to generate new map when JetBrains/phpstorm-stubs are updated...

theofidry: true, can be left alone then I guess :)

src/SourceLocator/SourceStubber/PhpStormStubsMap.php
theofidry: Thinking again about it: if it's not too slow to generate maybe a Composer hook on install/update could be enough

Ocramius: This may take a while, especially if the stubs reference each other somewhere (even in something as simple as `Exception`)

src/SourceLocator/SourceStubber/PhpStormStubsMap.php
```diff
@@ -0,0 +1,4910 @@
...
+    'zend/zend.php',
+    'zend/zend_f.php',
+    'zip/zip.php',
+    'zlib/zlib.php',
+    'zmq/zmq.php',
+];
+
+return [
+    [
+        'AMQPBasicProperties' => &$files[28],
```
Ocramius: Uhhh, what is going on here? Why is this by-ref?

Ocramius: Indeed: this should be a constant expression, as that would allow some degree of OpCache optimisations. Memory optimisation shouldn't be a big concern for such a small definition file

src/SourceLocator/SourceStubber/PhpStormStubsMap.php
```diff
@@ -0,0 +1,4910 @@
+<?php
+
+declare(strict_types=1);
+
+$files = [
```
Ocramius: Would it be OK to make this a static class with a constant? I don't like having global variables declared under an `src/` path, so this entire structure should be the result of a `var_export()`, and pushed into a `private` class constant

bin/generate-phpstorm-stubs-map.php
```diff
@@ -0,0 +1,144 @@
...
+
+declare(strict_types=1);
+
+use PhpParser\Node;
+use PhpParser\NodeTraverser;
+use PhpParser\NodeVisitor\NameResolver;
+use PhpParser\NodeVisitorAbstract;
+use PhpParser\ParserFactory;
+
+require __DIR__ . '/../vendor/autoload.php';
```
Ocramius: Please wrap the entire file in an IIFE:

```php
(function () {
    // ...
})()
```

bin/generate-phpstorm-stubs-map.php
```diff
@@ -0,0 +1,144 @@
+<?php
+
+declare(strict_types=1);
+
+use PhpParser\Node;
```
Ocramius: This file should still live within a namespace

bin/generate-phpstorm-stubs-map.php
```diff
@@ -0,0 +1,144 @@
...
+
+        if (! preg_match('/\.php$/', $fileInfo->getBasename())) {
+            continue;
+        }
+
+        $ast = $phpParser->parse(file_get_contents($fileInfo->getPathname()));
+
+        $nodeTraverser->traverse($ast);
+
+        /** @psalm-suppress UndefinedMethod */
```
Ocramius: Seems like `$fileVisitor` has the wrong type at this point?

bin/generate-phpstorm-stubs-map.php
```diff
@@ -0,0 +1,144 @@
...
+        foreach ($fileVisitor->getClassNames() as $className) {
+            $map['classes'][$className] = $fileInfo->getPathname();
+        }
+
+        /** @psalm-suppress UndefinedMethod */
```
Ocramius: Seems like `$fileVisitor` has the wrong type at this point?

bin/generate-phpstorm-stubs-map.php
```diff
@@ -0,0 +1,144 @@
...
+
+$output = "<?php\n\ndeclare(strict_types=1);\n\n\$files = [\n";
+foreach ($uniqueFiles as $fileName => $tmp) {
+    $relativeFileName = str_replace('\\', '/', substr($fileName, strlen($phpStormStubsDirectory) + 1));
+
+    $output .= sprintf('    \'%s\',%s', $relativeFileName, "\n");
+}
+$output .= "];\n\n";
+
+$output .= "return [\n";
```
Ocramius: This entire section should be a single `var_export()`

bin/generate-phpstorm-stubs-map.php
Ocramius: An assertion here is indeed required :+1:

src/SourceLocator/SourceStubber/PhpStormStubsMap.php
Ocramius: The reasoning being that I'd rather go through the composer autoloader here, rather than dynamic paths

src/SourceLocator/SourceStubber/PhpStormStubsMap.php
Ocramius: Hmm, @theofidry may be right here, as we're unlikely going to be releasing faster than `jetbrains/phpstorm-stubs`. Adding a post-install hook is super-annoying though...

src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
Ocramius: Uhhh, that is `private` state of this particular object: why is this breaking Box?

src/SourceLocator/SourceStubber/PhpStormStubsMap.php
kukulich: The script should be used only when we upgrade `jetbrains/phpstorm-stubs` in https://github.com/Roave/BetterReflection/blob/master/composer.json#L8

So I don't understand why we should need composer hook.

bin/generate-phpstorm-stubs-map.php
kukulich: `$fileVisitor` is anonymous class with custom methods :)

src/SourceLocator/SourceStubber/PhpStormStubsMap.php
Ocramius: Ah, the stubs are locked to a specific version. I thought this was a version range selector

bin/generate-phpstorm-stubs-map.php
Ocramius: Ah, indeed :+1:

bin/generate-phpstorm-stubs-map.php
kukulich: The `var_export` export is ugly :) If I change it to static class with constants, it should then be compatible with CS so it's easier to genereted by myself.

Ocramius: It may as well be ugly, but it works and it will prevent any escaping issues :P

Let's please not write anything around `\\` and stuff like that, and just use `var_export()`

theofidry:
Box executes something along the lines of:

Parallel_map(scope, $filesAndContents);

That function will spawn workers in which “scope” will be called. That
scope function is a static anonymous function which includes the scoper
(itself including BetterReflection). As a result this function is gonna be
serialized & unserialized but anonymous functions cannot be.

Ocramius:
@theofidry that's a problem that may occur anywhere at any point in time for any of the API of BetterReflection: I don't think that should be discussed in this PR, but I'd also suggest that this should be spawned as an issue in Box itself first.

theofidry:
@Ocramius sure, ideally a more long term plan would be to add support for
anonymous class serialization in amphp/parallel the same way the
serialization for anonymous closures has been added.

However right now there is no issue except for this piece of newly
introduced code which can easily be changed.

src/SourceLocator/SourceStubber/PhpStormStubsMap.php
kukulich: Done.

src/SourceLocator/SourceStubber/PhpStormStubsMap.php
kukulich: Done.

bin/generate-phpstorm-stubs-map.php
kukulich: Done.

bin/generate-phpstorm-stubs-map.php
kukulich: Done.

bin/generate-phpstorm-stubs-map.php
kukulich: Done. And fixed formatting automatically via `phpcbf` :)

bin/generate-phpstorm-stubs-map.php
kukulich: Done.

Ocramius:
@theofidry the issue is that anonymous symbols (classes, closures) may be introduced at any time, even in patch releases.

Unless explicitly stated, services are not serializable, so there is an architectural issue there, which should not be solved in here.

bin/generate-phpstorm-stubs-map.php
```diff
@@ -0,0 +1,178 @@
...
+
+        $mapFile = __DIR__ . '/../src/SourceLocator/SourceStubber/PhpStormStubsMap.php';
+
+        $bytesWritten = @file_put_contents($mapFile, $output);
+        if ($bytesWritten === false) {
+            throw new Exception(sprintf('File "%s" is not writeable.', $mapFile));
+        }
+
+        exit('Done');
+    } catch (Throwable $e) {
```
Ocramius: This can be dropped

bin/generate-phpstorm-stubs-map.php
```diff
@@ -0,0 +1,178 @@
...
+        $exportedFunctions = var_export($mapWithRelativeFilePaths['functions'], true);
+
+        $output = <<<"PHP"
+<?php
+
+declare(strict_types=1);
+
+namespace Roave\BetterReflection\SourceLocator\SourceStubber;
+
+final class PhpStormStubsMap
```
Ocramius: Needs to be `@internal`

test/unit/SourceLocator/SourceStubber/PhpStormStubsSourceStubberTest.php
```diff
@@ -154,6 +147,11 @@ private function assertSameClassAttributes(CoreReflectionClass $original, Reflec
         $this->assertSameInterfaces($original, $stubbed);
 
         foreach ($original->getMethods() as $method) {
+            // Needs fix in JetBrains/phpstorm-stubs
```
Ocramius: Can we make an issue out of this?

bin/generate-phpstorm-stubs-map.php
kukulich: Done.

bin/generate-phpstorm-stubs-map.php
kukulich: Done.

test/unit/SourceLocator/SourceStubber/PhpStormStubsSourceStubberTest.php
kukulich: https://github.com/JetBrains/phpstorm-stubs/pull/573
https://github.com/JetBrains/phpstorm-stubs/pull/574
https://github.com/JetBrains/phpstorm-stubs/pull/583
https://github.com/JetBrains/phpstorm-stubs/pull/587
https://github.com/JetBrains/phpstorm-stubs/pull/582
https://github.com/JetBrains/phpstorm-stubs/pull/584
https://github.com/JetBrains/phpstorm-stubs/pull/585

https://github.com/php/php-src/pull/4139

And I'm not done yet :)

Ocramius: Ouch xD

Ocramius:
Merging here.

@theofidry let's discuss a possible resolution for serialisation in a separate issue/PR

Ocramius:
Thanks @kukulich!

PR #471, opened 2019-05-13
Added ReflectionConstant

kukulich:
Based on https://github.com/Roave/BetterReflection/pull/470


kukulich:
@theofidry Can you try this please? :)

theofidry:
I will :) May need to wait tonight or tomorrow evening

kukulich:
Rebase done :)

phpunit.xml.dist
```diff
@@ -8,6 +8,9 @@
     <filter>
         <whitelist processUncoveredFilesFromWhitelist="true">
             <directory suffix=".php">./src</directory>
+            <exclude>
+              <file>./src/SourceLocator/SourceStubber/PhpStormStubsMap.php</file>
```
Ocramius: Is it worth excluding this at all?

phpstan.neon
```diff
@@ -9,5 +9,6 @@ parameters:
         # Some parent constructors are explicitly to be ignored
         - '#does not call parent constructor#'
         - '#Access to an undefined property PhpParser\\Node\\Param::\$isOptional#'
+        - '#Access to an undefined property PhpParser\\Node\\Expr::\$value#'
```
Ocramius: Seems invalid: is a constant an `Expr`?

phpstan.neon
```diff
@@ -9,5 +9,6 @@ parameters:
...
         # Impossible to define type hint for anonymous class
-        - '#Call to an undefined method PhpParser\\NodeVisitorAbstract::(clearNodes|getClassNodes|getFunctionNodes)\(\)#'
+        - '#Call to an undefined method PhpParser\\NodeVisitorAbstract::(clearNodes|getClassNodes|getFunctionNodes|getConstantNodes|getNode|setConstantName)\(\)#'
```
Ocramius: Can we limit these exclusions to just the stub map generator

src/Reflection/StringCast/ReflectionConstantStringCast.php
```diff
@@ -0,0 +1,47 @@
...
+namespace Roave\BetterReflection\Reflection\StringCast;
+
+use Roave\BetterReflection\Reflection\ReflectionConstant;
+use function gettype;
+use function sprintf;
+
+/**
+ * @internal
+ */
+final class ReflectionConstantStringCast
```
Ocramius: Reading this class doesn't make it clear what it is to be used for: can you add a sentence explaining that it is supporting/utility?

src/Reflection/ReflectionConstant.php
```diff
@@ -0,0 +1,283 @@
...
+        $constant->reflector          = $reflector;
+        $constant->node               = $node;
+        $constant->locatedSource      = $locatedSource;
+        $constant->declaringNamespace = $namespace;
+        $constant->positionInNode     = $positionInNode;
+
+        return $constant;
+    }
+
+    private static function createFromFuncCallNode(
```
Ocramius: S/func/function

Ocramius: Is this for `declare()` calls?

src/SourceLocator/Ast/FindReflectionsInTree.php
```diff
@@ -96,6 +97,37 @@ public function enterNode(Node $node)
...
+
+                        $this->reflections[] = $reflection;
+                    }
+
+                    return null;
+                }
+
+                if ($node instanceof Node\Expr\FuncCall
+                    && $node->name instanceof Node\Name
+                    && $node->name->toLowerString() === 'define'
```
Ocramius: Do we operate on a FQN here? Risky otherwise, as `define()` can be declared in userland

src/Reflector/ConstantReflector.php
```diff
@@ -0,0 +1,64 @@
...
+            throw Exception\IdentifierNotFound::fromIdentifier($identifier);
+        }
+
+        return $constantInfo;
+    }
+
+    /**
+     * Get all the constants available in the scope specified by the SourceLocator.
+     *
+     * @return ReflectionConstant[]
```
Ocramius: Can we restrict on key type here?

src/SourceLocator/Ast/Strategy/NodeToReflection.php
```diff
@@ -47,6 +49,17 @@ public function __invoke(
             );
         }
 
+        if ($node instanceof Node\Stmt\Const_) {
+            return ReflectionConstant::createFromNode($reflector, $node, $locatedSource, $namespace, $positionInNode);
+        }
+
+        if ($node instanceof Node\Expr\FuncCall
+            && $node->name instanceof Node\Name
+            && $node->name->toLowerString() === 'define'
```
Ocramius: Same as above: can't remember if the name is the FQN

src/Reflection/ReflectionConstant.php
```diff
@@ -0,0 +1,283 @@
...
+
+    /**
+     * Get the "short" name of the constant (e.g. for A\B\FOO, this will return
+     * "FOO").
+     */
+    public function getShortName() : string
+    {
+        if ($this->node instanceof Node\Expr\FuncCall) {
+            /** @psalm-suppress UndefinedPropertyFetch */
+            return $this->node->args[0]->value->value;
```
Ocramius: Do we know that `args[0]` is a constant? This should likely be a constant expression evaluation instead (which may fail)

src/SourceLocator/SourceStubber/ReflectionSourceStubber.php
```diff
@@ -111,11 +115,47 @@ public function generateFunctionStub(string $functionName) : ?StubData
...
-        return $this->createStubData($this->generateStubInNamespace($functionNode, $functionReflection->getNamespaceName()), $functionReflection);
+        return $this->createStubData($this->generateStubInNamespace($functionNode->getNode(), $functionReflection->getNamespaceName()), $extensionName);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function generateConstantStub(string $constantName) : ?StubData
+    {
+        // Not supported because of resource as value
```
Ocramius: Sigh @ this programming language

src/Reflection/ReflectionConstant.php
kukulich: Yes, it's for `define` calls.

Ocramius: Can we rename this to make that obvious?

src/SourceLocator/Ast/FindReflectionsInTree.php
kukulich: Changed the condition to ` && $node->name instanceof Node\Name\FullyQualified`

src/Reflection/ReflectionConstant.php
kukulich: Renamed to `createFromConstKeyword` and `createFromDefineFunctionCall`

phpunit.xml.dist
kukulich: Removed. It's not reported as "no coverage" anymore.

src/SourceLocator/Ast/Strategy/NodeToReflection.php
kukulich: Changed the condition to `&& $node->name instanceof Node\Name\FullyQualified`

src/Reflection/ReflectionConstant.php
theofidry: Bails out on:

```
<?php

namespace Acme;

const FOO_CONST = 'BAR';
```

because `$this->node->args[0]->value` is an instance of `PhpParser\Node\Expr\ConstFetch` which doesn't have a `$value` property.

kukulich: Strange, this should not match condition on line 116

theofidry: Ran PHP-Scoper with:

```php
$newNode = $this->node->args[0]->value;

if ($newNode instanceof Node\Scalar\String_) {
    // With:
    // define('BAR_CONST', foo());
    // $newNode->value will be 'BAR_CONST'
    //
    // However with:
    // define('Acme\BAR_CONST', foo());
    // then $newNode->value will be 'Acme\BAR_CONST'
    // and the constant declared will be Acme\BAR_CONST as in the namespace
    // Acme. Hence it's necessary to check if the value is not a namespaced string value.
    return (new Node\Name($newNode->value))->getFirst();
}

if ($newNode instanceof Node\Expr\ConstFetch) {
    // Happens with:
    // const FOO = 'BAR';
    return $newNode->name->getFirst();
}

throw new \InvalidArgumentException('Unknown node type: '.\get_class($newNode));
```

It passed the unit tests, checking for more

src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
```diff
@@ -107,6 +115,29 @@ public function generateFunctionStub(string $functionName) : ?StubData
...
+            return null;
+        }
+
+        $filePath = PhpStormStubsMap::CONSTANTS[$constantName];
+
+        if (! array_key_exists($constantName, $this->constantNodes)) {
+            $this->parseFile($filePath);
+        }
+
+        return new StubData($this->createStub($this->constantNodes[$constantName]), $this->getExtensionFromFilePath($filePath));
```
theofidry: Hm this is failing with `STDIN` and I suppose `STDOUT` and `STDERR` as well. This might be due to them being resources (e.g. `fopen('php://stderr', 'w')`)  as opposed to constant values

kukulich: Yes, see line 132. It´s not possible to support them

src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
```diff
@@ -128,11 +159,16 @@ private function parseFile(string $filePath) : void
         foreach ($this->cachingVisitor->getFunctionNodes() as $functionName => $functionNode) {
             $this->functionNodes[$functionName] = $functionNode;
         }
+
+        /** @psalm-suppress UndefinedMethod */
+        foreach ($this->cachingVisitor->getConstantNodes() as $constantName => $constantNode) {
```
theofidry: this wouldn't result in an `UndefinedMethod` if the underlying anonymous class was a dedicated class (and same for the FunctionReflector

src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
```diff
@@ -160,6 +199,38 @@ public function enterNode(Node $node) : ?int
...
+
+                if ($node instanceof Node\Expr\FuncCall
+                    && $node->name instanceof Node\Name
+                    && $node->name->toLowerString() === 'define'
+                ) {
+                    /** @psalm-suppress UndefinedPropertyFetch */
+                    $constantName = $node->args[0]->value->value;
+
+                    // Not supported because of resource as value
+                    if (in_array($constantName, ['STDIN', 'STDOUT', 'STDERR'], true)) {
```
theofidry: it's fine to not support it but as of now it's bailing out with a `TypeError` on https://github.com/Roave/BetterReflection/pull/471/files#r284000358 due to this

src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
theofidry: Just saw that. Is there a way to bail out with at least a more explicit message than a cryptic `TypeError` in this case?

kukulich: I should ignore these constants in the generate-map script. It should solve everything in much simpler way.

src/SourceLocator/Ast/FindReflectionsInTree.php
kukulich: Hmm, I cannot use `Node\Name\FullyQualified` because `define` call in namespace is resolved as `NamespaceName\define`. PHP itself will fallback to global `define` so I have to ignore FQN.

phpstan.neon
kukulich: Removed in favour of inline `@var` annotations.

phpstan.neon
kukulich: Done.

src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
kukulich: Should be fixed.

src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
kukulich: Should be fixed.

src/Reflection/ReflectionConstant.php
kukulich: I've made some improvements.

src/Reflector/ConstantReflector.php
kukulich: Changed to `array<int, ReflectionConstant>`

src/Reflection/StringCast/ReflectionConstantStringCast.php
kukulich: Added.

kukulich:
I hope it's ready for another review :)

src/SourceLocator/Ast/FindReflectionsInTree.php
Ocramius: This is still an issue IMO: we could do a lookup for `NamespaceName\define` and fall back to `\define` if it is not existing

src/SourceLocator/Ast/Strategy/NodeToReflection.php
Ocramius: Similar to https://github.com/Roave/BetterReflection/pull/471#discussion_r284239629 - need to still use `FullyQualified` and fallback if function lookup fails

kukulich: Would it be enough to solve it in one place? This class is internal.

Ocramius:
Just a note: consider not squashing while amidst a PR, because I have no idea of what was adapted since the last review

src/Reflection/ReflectionConstant.php
```diff
@@ -0,0 +1,308 @@
...
+
+        return $constant;
+    }
+
+    private static function createFromDefineFunctionCall(
+        Reflector $reflector,
+        Node\Expr\FuncCall $node,
+        LocatedSource $locatedSource
+    ) : self {
+        if (! ConstantNodeChecker::isValidDefineFunctionCall($node)) {
```
Ocramius: Would be better to make this an assertion, such as `assertValidDefineFunctionCall()`, which can then throw with the given AST node. Currently, the exception has no context, which will make things hard to debug

src/Reflection/Exception/InvalidNode.php
```diff
@@ -0,0 +1,15 @@
...
+
+namespace Roave\BetterReflection\Reflection\Exception;
+
+use RuntimeException;
+
+class InvalidNode extends RuntimeException
+{
+    public static function create() : self
+    {
+        return new self('Invalid node');
```
Ocramius: This exception is indeed too simplistic. I would suggest building a pretty-printer and dumping the code in here

src/Reflection/ReflectionConstant.php
```diff
@@ -0,0 +1,308 @@
...
+    private function __construct()
+    {
+    }
+
+    /**
+     * Create a ReflectionConstant by name, using default reflectors etc.
+     *
+     * @throws IdentifierNotFound
+     */
+    public static function createFromName(string $constantName) : self
```
Ocramius: Wondering if we need this for newer API, @asgrim?

src/Reflection/ReflectionConstant.php
```diff
@@ -0,0 +1,308 @@
...
+     *
+     * @return mixed
+     */
+    public function getValue()
+    {
+        if ($this->valueWasCached !== false) {
+            return $this->value;
+        }
+
+        /** @psalm-suppress PossiblyNullArrayOffset */
```
Ocramius: This is a leap of faith being done: would it be better for the utility that inspects constant declarations to already return the parameters?

src/SourceLocator/SourceStubber/AggregateSourceStubber.php
```diff
@@ -47,4 +47,20 @@ public function generateFunctionStub(string $functionName) : ?StubData
 
         return null;
     }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function generateConstantStub(string $constantName) : ?StubData
+    {
+        foreach ($this->sourceStubbers as $sourceStubber) {
```
Ocramius: Not important, but this stuff ia usually an `array_reduce()` with a `??` in its body

src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
```diff
@@ -107,6 +117,29 @@ public function generateFunctionStub(string $functionName) : ?StubData
         return new StubData($this->createStub($this->functionNodes[$functionName]), $this->getExtensionFromFilePath($filePath));
     }
 
+    /**
+     * {@inheritDoc}
+     */
+    public function generateConstantStub(string $constantName) : ?StubData
+    {
+        // https://github.com/JetBrains/phpstorm-stubs/pull/591
+        if (in_array($constantName, ['TRUE', 'FALSE', 'NULL'])) {
```
Ocramius: `strict` should be used here

src/SourceLocator/SourceStubber/ReflectionSourceStubber.php
```diff
@@ -111,11 +115,47 @@ public function generateFunctionStub(string $functionName) : ?StubData
...
+        foreach ($constants as $constantExtensionName => $extensionConstants) {
+            if (array_key_exists($constantName, $extensionConstants)) {
+                $found         = true;
+                $value         = $extensionConstants[$constantName];
+                $extensionName = $constantExtensionName !== 'user' ? $constantExtensionName : null;
+                break;
+            }
+        }
+
+        if (! $found) {
```
Ocramius: This pattern is quite terrible to maintain/debug: let's please use a `private` method or such instead, if possible.

src/Reflection/ReflectionConstant.php
kukulich: It cannot be `null` because if the node is `Node\Stmt\Const_` then the `ReflectionConstant` is created by `createFromConstKeyword` where `positionInNode` is not nullable.

src/SourceLocator/SourceStubber/ReflectionSourceStubber.php
kukulich: Done.

src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php

src/SourceLocator/SourceStubber/AggregateSourceStubber.php
kukulich: Done.

src/Reflection/Exception/InvalidNode.php
kukulich: Done.

src/Reflection/ReflectionConstant.php
kukulich: Done.

src/SourceLocator/Ast/Strategy/NodeToReflection.php
kukulich: https://github.com/Roave/BetterReflection/pull/471/commits/68cd029675a3397660dd726960bdb5c71a7aec31

I used a hack to solve circular dependency. I would be very happy for better suggestion how to solve it here.

src/BetterReflection.php
```diff
@@ -84,7 +94,9 @@ public function phpParser() : Parser
     public function astLocator() : AstLocator
     {
         return $this->astLocator
-            ?? $this->astLocator = new AstLocator($this->phpParser());
+            ?? $this->astLocator = new AstLocator($this->phpParser(), function () : FunctionReflector {
+                return $this->functionReflector();
```
theofidry: doesn't that mean you're using a different function reflector each time? Looks a bit weird to me and it also makes it way harder to instantiate the different Reflectors without `BetterReflection` (e.g. in PHP-Scoper since I'm configuring the locators differently

kukulich: Yes, I don't like it either. I used this hack as temporary solution to solve circular dependency. I would be very happy for better suggestion how to solve it.

Ocramius: This is only inside the `BetterReflection` instance, which is acceptable to me.

src/SourceLocator/Ast/Strategy/NodeToReflection.php
Ocramius: Works for me for now :+1:

Ocramius:
Thanks @kukulich!

src/BetterReflection.php
theofidry: Err yeah it's just that I don't see a nice way to do it for me:

```php
f (null === $this->reflector) {
            $phpParser = $this->getParser();
            $astLocator = new Locator($phpParser);

            $sourceLocator = new MemoizingSourceLocator(
                new PhpInternalSourceLocator(
                    $astLocator,
                    new AggregateSourceStubber(
                        new PhpStormStubsSourceStubber($phpParser),
                        new ReflectionSourceStubber()
                    )
                )
            );

            $classReflector = new ClassReflector($sourceLocator);

            $functionReflector = new FunctionReflector($sourceLocator, $classReflector);

            return new Reflector(
                $classReflector,
                $functionReflector,
                new ConstantReflector($sourceLocator, $classReflector)
            );
        }

        return $this->reflector;
```

Btw the locator is marked as `@internal` but I don't really see a convenient way to do without for now...

muglug:
@theofidry this broke https://github.com/humbug/php-scoper/blob/1c054e20a4e37ac99b7d1389a54065a73f90a1b2/src/Container.php#L71

theofidry:
@muglug since it's an internal component it's ok in terms of BC. It should be fixed soonish when I merge the PR in PHP-Scoper

muglug:
Nah, because you’re referencing it (with only one argument) in the link above. Psalm can warn about using internal classes, btw...

muglug:
It broke a Psalm build last night is what I’m saying (and likely many others)

theofidry:
Yes I'm referring it with only one argument because it was like that in 3.4 and changed in 3.5. It should be a BC break but it's not because it's tag as `@internal`.

I'm aware it was internal, it's just I cannot do without for PHP-Scoper :)

Also despite the fix being easy, it seems a few more changes within BetterReflection are not scoping friendly, i.e. my build is still failing after fixing the above. I'll try to fix this issue in https://github.com/humbug/php-scoper/pull/324 instead of https://github.com/humbug/php-scoper/pull/318 as the later has a few more changes that can make finding the fix a bit more tedious.

muglug:
Yeah, I fixed locally by pinning the composer version of BetterReflection in composer.json

Issue #444, opened 2018-09-03
`self::class`, `parent::class` and `static::class` are not correctly resolved into their corresponding strings


Ocramius:
Example:

```php
<?php

declare(strict_types=1);

use Roave\BetterReflection\BetterReflection;
use Roave\BetterReflection\Reflector\ClassReflector;
use Roave\BetterReflection\SourceLocator\Type\StringSourceLocator;

require_once __DIR__ . '/vendor/autoload.php';

$foo = (new ClassReflector(
    new StringSourceLocator(
        <<<'PHP'
<?php

class Baz {
}

class Foo extends Baz {
    public $bar = self::class;
    public $baz = static::class;
    public $taz = parent::class;
}
PHP
        ,
        (new BetterReflection())->astLocator()
    )
))->reflect('Foo');

var_dump($foo->getDefaultProperties());
```

The output will be something like:

```
array(3) {
  'bar' =>
  string(4) "self"
  'baz' =>
  string(6) "static"
  'taz' =>
  string(6) "parent"
}
```

The expected output should be:

```
array(3) {
  'bar' =>
  string(4) "Foo"
  'baz' =>
  string(6) "Foo"
  'taz' =>
  string(6) "Baz"
}
```

PR #472, opened 2019-05-14
 Fixed resolving self::, static:: and parent:: as property or constant values

kukulich:
Fixes https://github.com/Roave/BetterReflection/issues/444

### DIFF
--- a/src/NodeCompiler/CompileNodeToValue.php
+++ b/src/NodeCompiler/CompileNodeToValue.php
@@ -105,7 +105,7 @@ class CompileNodeToValue
         $classInfo = null;
 
         if ($className === 'self' || $className === 'static') {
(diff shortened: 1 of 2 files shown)


src/NodeCompiler/CompileNodeToValue.php
```diff
@@ -98,14 +98,27 @@ private function compileClassConstFetch(Node\Expr\ClassConstFetch $node, Compile
         $className = $node->class->toString();
 
         if ($nodeName === 'class') {
```
Ocramius: Can we isolate this entire block into a `private` method? I'd like to avoid the added CCLOC of `compileClassConstFetch`

kukulich: Done

Ocramius:
Thanks @kukulich!

PR #474, opened 2019-05-19
Run tests on PHP 7.4

kukulich:


.travis.yml
```diff
@@ -76,6 +76,16 @@ jobs:
         - echo "memory_limit=1G" >> /c/tools/php/php.ini
       script: vendor/bin/phpunit
 
+    - stage: Test
+      php: 7.4snapshot
```
Ocramius: We should probably still allow failures on 7.4

kukulich: Done.

PR #479, opened 2019-06-05
Fix behaviour of ReflectionClass::export/ReflectionObject::export to match behaviour of core classes

asgrim:
Replaces #397 from @muglug - I've rebased onto latest `master`, and implemented the proposal.

Note: introduces a BC break, so definitely targeting 4.0.0 for this.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -3,6 +3,22 @@
 This document serves as a reference to upgrade your current BetterReflection installation if improvements, deprecations
 or backwards compatibility (BC) breakages occur.
 
+## 4.0.0
+
+### BC Breaks
+
+* The following classes no longer implement `\Reflector`. If you need something that implements `\Reflector`, you should
+  wrap the BetterReflection in the appropriate `\Roave\BetterReflection\Reflection\Adapter\*` class.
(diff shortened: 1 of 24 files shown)


asgrim:
@Ocramius sure, if we want to go from 3.5 to 4.0 :+1: I'm fine with that

Ocramius:
Right, from now on `master` is only dedicated to `4.0.0`. @asgrim @kukulich if you have patches with BC Breaks, don't be scared to break it ;-)

Ocramius:
Fun times! Merging this resulted in following (something in dependencies):

```
There was 1 error:
1) Roave\BetterReflectionTest\Reflection\Adapter\ReflectionMethodTest::testAdapterMethods with data set #8 ('getDocComment', null, false, array())
TypeError: Return value of Mock_ReflectionMethod_15775ca8::getDocComment() must be of the type string, boolean returned
C:\Users\travis\build\Roave\BetterReflection\src\Reflection\Adapter\ReflectionMethod.php:109
C:\Users\travis\build\Roave\BetterReflection\test\unit\Reflection\Adapter\ReflectionMethodTest.php:130
```

PR #482, opened 2019-06-11
Removed wrong classes and functions from PHPStorm stubs map

kukulich:


theofidry:
Will need to port that to https://github.com/JetBrains/phpstorm-stubs/pull/614

asgrim:
@theofidry do we want to block this until upstream is merged?

theofidry:
No it's alright, I've referenced it in the PR :)

asgrim:
Alrighty, thanks!

asgrim:
Thanks @kukulich !

Ocramius:
Note: this is still targeting `4.0.0`. Do you need a fix in the current stable release, @kukulich?

kukulich:
It could be in 3.5.1 but it's probably not important. I don't think somebody would use class name `___PHPSTORM_HELPERS\\PS_UNRESERVE_PREFIX_static` :)

PR #498, opened 2019-10-02
Allow phpdocumentor/type-resolver version ^1.0.0

paul-thebaud:
This PR updates `composer.json` to allow usage of `phpdocumentor/type-resolver` with version `^1.0.0`.
It also updates `composer.lock` for those new dependencies (but in my opinion, `composer.lock` should not be committed for a package).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -8,7 +8,7 @@
         "jetbrains/phpstorm-stubs": "2019.2",
         "nikic/php-parser": "^4.2.1",
         "phpdocumentor/reflection-docblock": "^4.3.2",
-        "phpdocumentor/type-resolver": "^0.4.0 || ^1.0.0",
+        "phpdocumentor/type-resolver": "^1.0.0",
         "roave/signature": "^1.0"
     },
     "authors": [
--- a/composer.json
+++ b/composer.json
@@ -8,7 +8,7 @@
         "jetbrains/phpstorm-stubs": "2019.2",
         "nikic/php-parser": "^4.2.1",
         "phpdocumentor/reflection-docblock": "^4.3.2",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -8,7 +8,7 @@
         "jetbrains/phpstorm-stubs": "2019.2",
         "nikic/php-parser": "^4.2.1",
         "phpdocumentor/reflection-docblock": "^4.3.2",
-        "phpdocumentor/type-resolver": "^0.4.0",
+        "phpdocumentor/type-resolver": "^0.4.0 || ^1.0.0",
```
Ocramius: Please just bump it: let's not keep the old one around.

paul-thebaud: Ok. This is done in my last commit: dd1173567d2d4b850c9ff529d88170aaaf7480b6.
Thanks for the quick answer. I let you mark the code review as resolved if you think it is :)

Ocramius:
CI failed on something unrelated.

Ocramius:
Thanks @paul-thebaud!

kunicmarko20:
Any chance we can get a release with this fix or should I just use master?

Ocramius:
Hmm, I think I can do a round of releases on projects I maintain this week - will see if I can get it done, since I'm on the road.

kunicmarko20:
I am on master, no rush really. Thank you!

fezfez:
@Ocramius : any chance to have release for this ? Thanks in advance !

Ocramius:
@asgrim besides #538, I'd be OK with releasing the new major: WDYT?

kukulich:
It may be good to solve https://github.com/Roave/BetterReflection/issues/519 before new major.

PR #500, opened 2019-10-28
Correct a coding standard issue

duncan3dc:
While working on an issue Travis was shouting about coding standards, so I ran `phpcbf` and here's a PR


Ocramius:
Heya, I'm mostly waiting to be able to release doctrine/coding-standard, which would fix this and require the spacing

Ocramius:
Handled in #514, which applied the newer `doctrine/coding-standard`. Thanks anyway, @duncan3dc! :+1:

PR #501, opened 2019-10-28
Replaced the string cast with a getName() call

duncan3dc:
The `__toString()` method has been deprecated since 7.1, it's started to raise an error in 7.4 so this PR switches to use `getName()` instead.

`getName()` operates as this code expects (eg returns the type name without the nullable indicator `?`), and in PHP 8.0 `__toString()` will start to return the nullable indicator.

I don't think it's possible for `CoreReflectionType` to be passed to this method anymore, but the stubs suggest it is and I've not dug further into that yet, so this PR attempts to retain support for that scenario as well

### DIFF
--- a/src/SourceLocator/SourceStubber/ReflectionSourceStubber.php
+++ b/src/SourceLocator/SourceStubber/ReflectionSourceStubber.php
@@ -33,6 +33,7 @@ use ReflectionClassConstant;
 use ReflectionFunction as CoreReflectionFunction;
 use ReflectionFunctionAbstract as CoreReflectionFunctionAbstract;
 use ReflectionMethod as CoreReflectionMethod;
+use ReflectionNamedType as CoreReflectionNamedType;
 use ReflectionParameter;
 use ReflectionProperty as CoreReflectionProperty;
 use ReflectionType as CoreReflectionType;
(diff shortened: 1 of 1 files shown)


src/SourceLocator/SourceStubber/ReflectionSourceStubber.php
```diff
@@ -475,7 +476,7 @@ private function parameterDefaultValue(
      */
     private function formatType(CoreReflectionType $type) : NodeAbstract
     {
-        $name     = (string) $type;
+        $name     = $type instanceof CoreReflectionNamedType ? $type->getName() : (string) $type;
```
Ocramius: Is the second branch of this conditional ever reached? Mutation tests should indicate if that isn't happening

duncan3dc: I don't believe so, I'll take a look though and see if I can confirm and remove it

asgrim: That's a little way off yet, maybe just create an issue for now?

Ocramius: Yeah, I was mostly looking at what `ReflectionType` instances may pop up. An assertion would be best here, for now.

Ocramius:
Fixed in #522, closing here, marking as part of the milestone anyway, since the detailed explanation is excellent reference. Thanks @duncan3dc!

Ocramius:
Re-targeted to `4.0.0`: this was merged to a development branch only.

PR #505, opened 2019-12-15
fix anonymous class getParentClass

### LINKED ISSUE #504: Bug : Calling getParentClass on an Anonymous class returns self
Anonymous class with a parent class returns self when calling getParentClass which makes getParentClassNames go into an infinite recursion.

sidux:
Resolves #504


src/Reflection/ReflectionClass.php
```diff
@@ -775,8 +775,12 @@ public function getParentClass() : ?ReflectionClass
         }
 
         // @TODO use actual `ClassReflector` or `FunctionReflector`?
-        /** @var self $parent */
-        $parent = $this->reflector->reflect($this->node->extends->toString());
+        if ($this->isAnonymous()) {
+            $parent = self::createFromName($this->node->extends->toString());
```
Ocramius: This seems incorrect: what happens if you try to get the `ReflectionProperty` for `$someProperty` (defined in `ClassForHinting`) in your test? If I get it correctly, it won't be defined with this code.

Ocramius (changes requested):

src/Reflection/ReflectionClass.php
Ocramius: To be more precise, `assertEquals($parent, $reflector->reflector(ParentName::class))`

sidux: I improved the test to check for the property and it seems to work, I also added a `\PhpParser\NodeVisitor\NameResolver` to the `AnonymousClassObjectSourceLocator` to resolve the parent class name.

test/unit/Reflection/ReflectionClassTest.php
```diff
@@ -680,6 +681,28 @@ public function testIsAnonymousWithAnonymousClassNoNamespace() : void
...
+            new class extends ClassForHinting {
+                public function a() : void
+                {
+                }
+            }
+        );
+
+        $parent = $anonymousClassInfo->getParentClass();
+        self::assertSame(ClassForHinting::class, $parent->getName());
+        self::assertSame((string) $parent, (string) $reflector->reflect(ClassForHinting::class));
```
Ocramius: @asgrim can you check this as well? I'm unsure about the test behavior, think this should be a `self::assertEquals()?

Ocramius (approved):

src/Reflection/ReflectionClass.php
Ocramius: Checked this again: it won't use the same reflector configuration when calling `self::createFromName()`. 

The tests work because the defaults of `(new BetterReflection())` use the `ComposerAutoloader` (IIRC), but I was indeed right to smell something here :D

Ideally, `$this->reflector` should always to happen, and `self::createFromName()` should never be used.

I think the changes in `AnonymousClassObjectSourceLocator` are sufficient to make this patch work, or at least they are heading in the right direction: changes in `ReflectionClass` (here) shouldn't be needed.

Ocramius (changes requested):
As mentioned in detail in comment above, there is a design issue with this patch: using `ReflectionClass::createFromName()` leads to a completely new `BetterReflection` instance, and that means that the reflected class is completely disjoint from the current reflector. This would be highlighted when using a custom AST locator, such as an AST locator operating on a string.

@asgrim I think `ReflectionClass::createFromName()` (and similar methods) need to be deprecated

src/Reflection/ReflectionClass.php
asgrim: If I'm understanding what's going on correctly, it *can't* use the same reflector, since the reflector used when doing `\Roave\BetterReflection\Reflection\ReflectionObject::createFromInstance` will ONLY use the located source for `\Roave\BetterReflection\SourceLocator\Type\AnonymousClassObjectSourceLocator`:

https://github.com/Roave/BetterReflection/blob/311c6c98de57807ac186f55d0c3dfa96971fa8f0/src/Reflection/ReflectionObject.php#L66-L70

I think this means the parent class isn't in scope, so any class located using that instance of the reflector is going to be the anonymous one (hence why, if you remove the change here, it will return the wrong value:

```
Expected :'Roave\BetterReflectionTest\Fixture\ClassForHinting'
Actual   :'class@anonymous /home/james/workspace/better-reflection/test/unit/Reflection/ReflectionClassTest.php(694)'
```

Possible ideas for fixing, let's discuss before implementing though (I think the right place to fix this is in `\Roave\BetterReflection\Reflection\ReflectionObject::createFromInstance`?)

 * Instead of ONLY using `\Roave\BetterReflection\SourceLocator\Type\AnonymousClassObjectSourceLocator`, wrap it up with an aggregate source locator of `\Roave\BetterReflection\BetterReflection::sourceLocator()` combined with `AnonymousClassObjectSourceLocator`
 * Add a second parameter to `createFromInstance` to inject an existing source locator
 * Is there some other way of getting an existing (or rather "the" existing source locator) in...?

Ocramius:
Removed milestone: patch needs more work.

Ocramius:
@sidux did you have time to check the feedback on this?

src/Reflection/ReflectionObject.php
```diff
@@ -64,10 +65,17 @@ public static function createFromInstance($object) : ReflectionClass
...
 
         if (strpos($className, ReflectionClass::ANONYMOUS_CLASS_NAME_PREFIX) === 0) {
-            $reflector = new ClassReflector(new AnonymousClassObjectSourceLocator(
-                $object,
-                (new BetterReflection())->phpParser()
-            ));
+            $reflector = new ClassReflector(
+                new AggregateSourceLocator(
+                    [
+                        (new BetterReflection())->sourceLocator(),
```
Ocramius, line 71: Note: the problem with this diff is that we have multiple `(new BetterReflection())` instances here, which means that these source locators are operating on disjunct sources.

Ocramius: Te better clarify scope, use a `StringSourceLocator` and reproduce the problem within a single string containing an anonymous class: you will see a crash (can't find the parent class)

test/unit/Reflection/ReflectionClassTest.php
```diff
@@ -680,6 +681,27 @@ public function testIsAnonymousWithAnonymousClassNoNamespace() : void
         self::assertStringEndsWith('Fixture/AnonymousClassNoNamespace.php(3)', $classInfo->getName());
     }
 
+    public function testIsAnonymousWithParentClass() : void
+    {
+        $reflector = new ClassReflector(
```
Ocramius, line 686: Is this unused on purpose, or is the test incomplete?

Do you need any support to get this written?

Maybe just a reproducer and drop it to us to get the bug fixed?

sidux: @Ocramius sorry, i was a bit busy last weeks, i will commit a test with your last suggestion about using `StringSourceLocator`, i will try to get this fixed by tomorrow, if am not able, i will close this PR and add the test code to the issue.

src/SourceLocator/Type/AnonymousClassObjectSourceLocator.php
```diff
@@ -136,6 +137,7 @@ public function getAnonymousClassNode() : ?Class_
 
         $nodeTraverser = new NodeTraverser();
         $nodeTraverser->addVisitor($nodeVisitor);
+        $nodeTraverser->addVisitor(new NameResolver());
```
Ocramius, line 140: Can this change be tested in isolation? The added test case works with `ReflectionClass`

test/unit/Reflection/ReflectionClassTest.php
```diff
@@ -746,6 +747,27 @@ public function testIsAnonymousWithAnonymousClassNoNamespace() : void
         self::assertStringEndsWith('Fixture/AnonymousClassNoNamespace.php(3)', $classInfo->getName());
     }
 
+    public function testIsAnonymousWithParentClass() : void
```
Ocramius, line 750: Should this test be moved to `ReflectionObjectTest`?

test/unit/Reflection/ReflectionClassTest.php
```diff
@@ -746,6 +747,27 @@ public function testIsAnonymousWithAnonymousClassNoNamespace() : void
...
+            namespace Foo { class FooClass {}}
+            namespace Bar {
+                use Foo\FooClass;
+                new class extends FooClass {};
+            }', $this->astLocator),
+        );
+        $parent    = $reflector->getAllClasses()[1]->getParentClass();
+        self::assertSame('Foo\FooClass', $parent->getName());
+
+        $anonymousClassInfo = ReflectionObject::createFromInstance(
```
Ocramius, line 763: This should probably be a new/separate test method

Ocramius (changes requested):
Only few more changes requested, but overall looks very good! 

I can pick up the remaining changes, if you need any help.

asgrim:

kukulich:
This should be fixed in https://github.com/Roave/BetterReflection/pull/750 with https://github.com/kukulich/BetterReflection/commit/569ea7dcb8664ebe031833dbefed03d04ba10a2f

kukulich:
Yes, this is fixed in #750, there's also test: https://github.com/kukulich/BetterReflection/commit/569ea7dcb8664ebe031833dbefed03d04ba10a2f#diff-e0dd711174c42542396ea4773a4b7db9ba6025250e7d50950c152a6d43b5fe91R786

PR #506, opened 2019-12-18
Fix CS

ondrejmirtes:
Hi, I'm in the process of [creating my own PHP 7.1-compatible fork](https://github.com/ondrejmirtes/BetterReflection) (hope you're fine with that! It will land in PHPStan pretty soon I expect). I stumbled upon this build issue (it fails here on master), so I'm sending a fix :)

Thank you!

### DIFF
--- a/test/unit/SourceLocator/Type/MemoizingSourceLocatorTest.php
+++ b/test/unit/SourceLocator/Type/MemoizingSourceLocatorTest.php
@@ -205,7 +205,7 @@ class MemoizingSourceLocatorTest extends TestCase
                 Reflector $reflector,
                 Identifier $identifier
             ) use (
-                & $fetchedSymbolsCount
+                &$fetchedSymbolsCount
             ) : ?Reflection {
                 $identifierId = spl_object_hash($identifier);
                 $reflectorId  = spl_object_hash($reflector);


Ocramius:
Do you really need PHP 7.1 support, considering that it's EOL'd and out of security support? On this end, we'll likely keep compat only with the current minor, since PHP is an elusive target.

test/unit/SourceLocator/Type/MemoizingSourceLocatorTest.php
```diff
@@ -205,7 +205,7 @@ private function assertMemoization(
                 Reflector $reflector,
                 Identifier $identifier
             ) use (
-                & $fetchedSymbolsCount
+                &$fetchedSymbolsCount
```
Ocramius: This is potentially caused by `doctrine/coding-standard:6.0.0` depending on PSR-12 rules that were changed in upstream phpcs. Upgrading to `doctrine/coding-standard:7.0.2` should fix these, in theory.

ondrejmirtes:
> Do you really need PHP 7.1 support, considering that it's EOL'd and out of security support? 

I'm thinking pragmatically. There's nothing in PHP 7.2 that I'd take advantage of. It also breaks my heart when someone `composer require`s PHPStan and it installs PHPStan 0.9.2 (ancient version by now) because it's the last one that supports PHP 7.0. The user does not even know they're not on the latest version.

When it's cheap for me to support PHP 7.1, there isn't a reason not to. Funny thing - BetterReflection would still fully support PHP 7.1 if it weren't for the constraint in composer.json. The only thing I had to do to make the build pass was to downgrade PHPUnit to 7.5.18 :)

Also, thanks to PHAR-only distribution, I now have the advantage of a build step. I could theoretically develop PHPStan for PHP 7.4 only, and transform the sources to also support PHP 7.1 every time before publishing.

test/unit/SourceLocator/Type/MemoizingSourceLocatorTest.php
ondrejmirtes: @kukulich told me that he'll be updating BetterRefleciton to `doctrine/coding-standard:7.x` so I'll leave it up to him :)

Ocramius:
> When it's cheap for me to support PHP 7.1, there isn't a reason not to. 

We dropped previous versions because it was not cheap for us to maintain them :woman_shrugging:

ondrejmirtes:
Yeah, but this does not apply now because of small jumps in 7.2 and 7.3. I also wanted `void` and nullable typehints so I dropped 7.0.

PR #514, opened 2020-01-10
Update doctrine/coding-standard from 6.0.0 to 7.0.2

Lctrs:
Closes #506. 
Closes #507. 
Closes #510.

### DIFF
--- a/test/unit/Reflector/ClassReflectorTest.php
+++ b/test/unit/Reflector/ClassReflectorTest.php
@@ -12,6 +12,7 @@ use Roave\BetterReflection\Reflector\Exception\IdentifierNotFound;
 use Roave\BetterReflection\SourceLocator\Type\SingleFileSourceLocator;
 use Roave\BetterReflection\SourceLocator\Type\StringSourceLocator;
 use Roave\BetterReflectionTest\BetterReflectionSingleton;
+use function assert;
 
 /**
  * @covers \Roave\BetterReflection\Reflector\ClassReflector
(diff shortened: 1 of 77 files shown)


test/unit/Reflector/ClassReflectorTest.php
```diff
@@ -32,12 +33,12 @@ public function testReflectProxiesToSourceLocator() : void
     {
         $reflection = $this->createMock(ReflectionClass::class);
 
-        /** @var StringSourceLocator|MockObject $sourceLocator */
         $sourceLocator = $this
             ->getMockBuilder(StringSourceLocator::class)
             ->disableOriginalConstructor()
             ->setMethods(['locateIdentifier'])
             ->getMock();
+        assert($sourceLocator instanceof StringSourceLocator || $sourceLocator instanceof MockObject);
```
Ocramius: Should be `&&` for these transformations

test/unit/SourceLocator/Type/AbstractSourceLocatorTest.php
```diff
@@ -22,28 +23,28 @@ class AbstractSourceLocatorTest extends TestCase
 {
     public function testLocateIdentifierCallsFindReflection() : void
     {
-        /** @var Reflector|MockObject $mockReflector */
         $mockReflector = $this->createMock(Reflector::class);
+        assert($mockReflector instanceof Reflector || $mockReflector instanceof MockObject);
```
Ocramius: Since newer phpunit versions, we can drop these, since direct `$this->createMock()` calls (without builder) can be correctly inferred AFAIK

Ocramius:
Build https://travis-ci.org/Roave/BetterReflection/jobs/635186981 failed due to too old phpunit version: we can bump the minimum supported version safely

Lctrs:
:crossed_fingers:

Ocramius:
```
Fatal error:  Allowed memory size of 1073741824 bytes exhausted (tried to allocate 266240 bytes) in phar:///home/travis/build/Roave/BetterReflection/vendor/phpstan/phpstan/phpstan/vendor/ondrejmirtes/better-reflection/src/Reflector/Exception/IdentifierNotFound.php on line 24
```

LULZ

CS is good though, so I would merge this as-is, and bring the new problem with me as a separate issue.

PR #516, opened 2020-01-10
Run PHPStan without memory limit

Lctrs:


Ocramius:
This fixes the phpstan analysis, as it seems 👍

Lctrs:
Yeah. I suspect PHPStan to parse all JetBrains' stubs, which are rather quite large.

PR #517, opened 2020-01-10
Bump psalm to 3.8.2

Lctrs:


Lctrs:
It will need a good triage between false positives/real errors.

Lctrs:
@Ocramius Some advices on this one would be really appreciated :)

Ocramius:
I think I'm OK with this as-is for now, will review the baseline anyway though

psalm-baseline.xml
```diff
@@ -0,0 +1,145 @@
...
+    <InvalidScalarArgument occurrences="2"/>
+  </file>
+  <file src="src/SourceLocator/Type/DirectoriesSourceLocator.php">
+    <DocblockTypeContradiction occurrences="1">
+      <code>is_string($directory)</code>
+    </DocblockTypeContradiction>
+  </file>
+  <file src="src/SourceLocator/Type/FileIteratorSourceLocator.php">
+    <DocblockTypeContradiction occurrences="1">
+      <code>$fileInfo instanceof SplFileInfo</code>
```
Ocramius: `@param Iterator|SplFileInfo[] $fileInfoIterator` is wrong here: this is an `@param Iterator<SplFileInfo> $fileInfoIterator` instead

Ocramius:
We should probably open a new issue for the comments above. Reaching 100% type coverage in psalm will be damn hard, but worth the effort.

If you can rebase this one, and open a new issue, we should be good to go here for now

Lctrs:
@Ocramius Done :)

PR #518, opened 2020-01-10
Run CI only once on PR from another branch of the repo

Lctrs:
This effectively restricts CI on push to branch or tags only.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,8 @@ os: linux
 dist: xenial
 language: php
 
+if: type != push OR branch = master OR branch =~ /^v?\d+\.\d+(\.\d+)?(-\S*)?$/
+
 before_script:
   - rm composer.lock
   - travis_retry composer update --prefer-dist $DEPENDENCIES


.travis.yml
```diff
@@ -2,6 +2,8 @@ os: linux
 dist: xenial
 language: php
 
+if: type != push OR branch = master OR branch =~ /^v?\d+\.\d+(\.\d+)?(-\S*)?$/
```
Ocramius: Not sure if this optimisation is worth it, considering that the repo doesn't get frequent/daily contributions

Lctrs: I got annoyed by Dependabot running CI twice on every opened PR/rebase too much maybe.

And well, it's also good for the planet™

Ocramius: Makes sense, but you shall be pinged if this doesn't work as expected :-P

Lctrs: And I'll answer the call :sunglasses:

Ocramius: Thanks! :heart:

PR #521, opened 2020-01-11
Travis improvements

kukulich:


.travis.yml
```diff
@@ -201,33 +249,33 @@ jobs:
       script: test/demo/check-demo.sh
 
     - stage: Check Coding Standard
-      php: 7.3
+      php: 7.4
```
Lctrs: Coding standard should be checked against lowest supported PHP version (so `7.2`) to not get unsupported feature from higher version.

kukulich: It should be configured to not use features from newer versions.

Lctrs: With last changes : :+1:

PR #523, opened 2020-01-11
ReflectionClass::getConstructor() should work event if the constructor isn't the first method in the class

kukulich:
It looks @ondrejmirtes is too lazy to send PR 😛


Ocramius:
Re-targeted to 4.0.0: this was merged to a development branch only.

PR #524, opened 2020-01-11
Speedup code coverage generation by using xdebug filter

Lctrs:


kukulich:
Why not use pcov?

Lctrs:
Could be an option too. But due to difference in the coverage generation, diff could be huge. The decision is not on me.

Issue #525, opened 2020-01-11
Achieve 100% type coverage with Psalm


Lctrs:
See https://github.com/Roave/BetterReflection/pull/517 for a great explanation of current issues made by the almighty @Ocramius.

orklah:
I tried to look at this.

There are two main sources of mixed type in the library.
- composer.json and installed.json are two big chunks of decoded json that are passed around as mixed[]. It could be handled with this:
`array{name: string, autoload: array{classmap: array<int, string>, files: array<int, string>, psr-4: array<int, string>, psr-0: array<int, string>}}`.
It's a big chunk of doc that have to be repeated 17 times in MakeLocatorFor* Classes but when it's done, it rises the coverage from 98.9751 to 99.2148

- The second is CompileNodeToValue::__invoke, It's a big complicated and recursive method who return mixed. I don't know enough to know what's really going on here, but if this mixed type here could be removed (by actually detailing every possible type), It could then be propagated easily to the rest of the code and rise almost to 100%...

kukulich:
I think that `CompileNodeToValue::__invoke` returns simple types (or array of simple types)

orklah:
> I think that `CompileNodeToValue::__invoke` returns simple types (or array of simple types)

Great! I will update this and see how it goes!

orklah:
Note: https://github.com/vimeo/psalm/pull/2737 This is a new config for the next Psalm release that will allow reading Php-parser's @property notations. When activated, it should bump the type coverage a bit

Ocramius:
Nice! 💪

BTW, what happens if we run with `totallyTyped="true"`?

orklah:
> BTW, what happens if we run with `totallyTyped="true"`?

35 errors / 99.6937% (with patch from PR that was fixed another way)
Some fixed in https://github.com/nikic/PHP-Parser/pull/647
Some due to black magic and closures
Some due to Reflection returning unpredictable values

It could probably still be narrowed but it's tedious work :p

Ocramius:
With 35 errors left, there is a lot of hope \o/

We just can't achieve 100% if we rely on any core API that exposes `mixed`, heh :D

Ocramius:
Closing here: we're already at a good point :)

PR #526, opened 2020-01-12
Improve Psalm coverage for decoded datas from json files

orklah:
As discussed here  #525 I described some types to improve Psalm coverage from 98.9751% to 99.2148%

This is not very elegant but I don't see a better way without refactoring the code just to pass smaller chunks of the whole file.


Ocramius:
Good improvement! Can you also update the baseline accordingly?

orklah:
> Good improvement! Can you also update the baseline accordingly?

Thanks! Nothing changes in the baseline when I launch the command. I think it's because I didn't add nor remove errors, just type coverage

Issue #511, opened 2020-01-03
ReflectionNamedType adapter is missing


ondrejmirtes:
In PHP 7.x series, ReflectionNamedType is always returned from methods like `ReflectionParameter::getType()` or `ReflectionMethod::getReturnType()`.

BetterReflection in the Adapter namespace does not behave like this - since this namespace is supposed to be used en lieu of the native reflection, this break these assumptions.

Proof of concept that works for me: https://github.com/ondrejmirtes/BetterReflection/commit/092cb1053230854f4603a70f6315785cb39e72bf

ondrejmirtes:
If I reword this problem - there's no way to get the name of the type without using `ReflectionType::__toString()` that was deprecated in PHP 7.4.

PR #528, opened 2020-01-12
Added ReflectionNamedType adapter

kukulich:
Fixes https://github.com/Roave/BetterReflection/issues/511

I hope it's backward compatible.

### DIFF
--- a/src/Reflection/Adapter/ReflectionNamedType.php
+++ b/src/Reflection/Adapter/ReflectionNamedType.php
@@ -1,0 +1,25 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Roave\BetterReflection\Reflection\Adapter;
+
+use Roave\BetterReflection\Reflection\ReflectionType as BetterReflectionType;
+
+class ReflectionNamedType extends ReflectionType
+{
+    /** @var BetterReflectionType */
+    private $betterReflectionType;
+
+    public function __construct(BetterReflectionType $betterReflectionType)
+    {
+        parent::__construct($betterReflectionType);
+
+        $this->betterReflectionType = $betterReflectionType;
+    }
+
+    public function getName()
+    {
+        return $this->betterReflectionType->getName();
+    }
+}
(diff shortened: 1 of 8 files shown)


src/Reflection/Adapter/ReflectionType.php
```diff
@@ -10,7 +10,7 @@
 class ReflectionType extends CoreReflectionType
 {
     /** @var BetterReflectionType */
-    private $betterReflectionType;
+    protected $betterReflectionType;
```
Ocramius: Avoidable?

src/Reflection/Adapter/ReflectionNamedType.php
```diff
@@ -0,0 +1,13 @@
...
+
+declare(strict_types=1);
+
+namespace Roave\BetterReflection\Reflection\Adapter;
+
+class ReflectionNamedType extends ReflectionType
+{
+    public function getName()
+    {
+        return $this->betterReflectionType->getName();
```
Ocramius: Maybe we should implement `getName` on the parent type, then let inheritance do its thing here?

kukulich: We don't need the `ReflectionNamedType` at all after that change... Hmm?

Ocramius: True that, but it would still be there for symmetry with the core API

kukulich: But there's no `getName()` method in `ReflectionType` in the core API :) How much symmetric should we be? :)

So maybe it would be better to just fix the `protected`?

Ocramius: I'm fine with protected if it's `@internal` then 👍

kukulich: So anything I should do now? :)

Ocramius: Either we mark the `protected` field as `@internal`, or we declare `getName()` on `ReflectionType`. I think the first approach is cleaner, to avoid polluting the API

kukulich: I've made a version where everything is private. Is it ok or should I change it to `@internal`?

Ocramius: All good as-is :+1:

PR #529, opened 2020-01-13
Fix dependencies installation

Lctrs:
I suspect that the dependencies has been updated with a higher PHP version than `7.2.0`, our minimal supported PHP version.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -941,35 +941,34 @@
         },
         {
             "name": "ocramius/package-versions",
-            "version": "1.5.1",
+            "version": "1.4.2",
             "source": {
                 "type": "git",
                 "url": "https://github.com/Ocramius/PackageVersions.git",
-                "reference": "1d32342b8c1eb27353c8887c366147b4c2da673c"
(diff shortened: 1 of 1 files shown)


kukulich:
Is it useful for us to bump PHP to `^7.4` ? I can do it and add types to properties but I already know that @ondrejmirtes will have to revert all my changes in his fork to support even PHP 7.1.

Ocramius:
Well, over here, development needs to proceed, and we'll need to start working on 8.x support as next feature. PHP 7.2~7.3 support will be dropped much before those go EOL: if support for older versions if needed, it will be in older minor releases

kukulich:
So can we release 4.0.0 with `^7.2` and bump PHP to `^7.4` then?

Ocramius:
Seems acceptable

ondrejmirtes:
I've wanted to write a tool that would convert 7.4 code to 7.1 anyway :D So I can write PHPStan in phpstan-src in PHP 7.4 only, but still support PHP 7.1 in the PHAR.

Ocramius:
If you are working on that, see https://github.com/endel/galapagos

Issue #530, opened 2020-01-14
ReflectionClass::getTraits don't return the same type as native ReflectionClass::getTraits


orklah:
Hi

While working on #525, I documented ReflectionClass::getTraits as returning `list<ReflectionClass>`.

This is deduced by looking at the data source `$this->node->stmts` which is an `array<int, Node\Stmt>` (or even a `list<Node\Stmt>`, not sure yet).

However, the doc of [ReflectionClass::getTraits](https://www.php.net/manual/en/reflectionclass.gettraits.php) (and psalm stubs) describe the return type as:
> an array with trait names in keys and instances of trait's ReflectionClass in values

> `array<string,ReflectionClass>`

I'm not sure what to do about that...

I wanted to add a failing test in a PR but the only place where I've seen a call to getTraits is in a dataProvider that checks that the return is an empty array.

Ocramius:
I guess that's under-tested now... I'd leave it as-is until someone adds testing around `getTraits`

orklah:
The issue is reversed in ReflectionClass::getReflectionConstants. The native object will return a numerically indexed array where the library returns a string indexed array.

Here's an example of native code: https://3v4l.org/t05Dv

kukulich:
I think everything is ok, see https://github.com/Roave/BetterReflection/pull/593

ReflectionClass adapter returns the right types.

orklah:
Yep, these 3 cases were fixed by @Ocramius in https://github.com/Roave/BetterReflection/commit/beca868be431778a5c5da3ee0aa457763d7f20b4

The @psalm-suppress could have been removed at that occasion.

PR #531, opened 2020-01-14
add psalm types

orklah:
This is a draft PR with the changes I made for #525 to see the rest of the CI and to discuss about some changes

The increase in type coverage is disappointing because a lot of mixed types still comes from outside.

Still, the changes prepare the way for when 
https://github.com/nikic/PHP-Parser/pull/647
and
https://github.com/vimeo/psalm/pull/2612

will be merged.

### DIFF
--- a/src/NodeCompiler/CompileNodeToValue.php
+++ b/src/NodeCompiler/CompileNodeToValue.php
@@ -24,7 +24,8 @@ class CompileNodeToValue
      *
      * @param Node\Stmt\Expression|Node\Expr $node Node has to be processed by the PhpParser\NodeVisitor\NameResolver
      *
-     * @return scalar|scalar[]|null
+     * @return bool|int|float|string|array|null
+     * @psalm-return scalar|array<scalar>|null
      *
      * @throws Exception\UnableToCompileNode
      */
@@ -60,7 +61,8 @@ class CompileNodeToValue
     /**
      * Compile constant expressions
      *
-     * @return scalar|scalar[]|null
+     * @return bool|int|float|string|array|null
+     * @psalm-return scalar|array<scalar>|null
      *
      * @throws Exception\UnableToCompileNode
      */
@@ -86,7 +88,8 @@ class CompileNodeToValue
     /**
      * Compile class constants
      *
-     * @return scalar|scalar[]|null
+     * @return bool|int|float|string|array|null
+     * @psalm-return scalar|array<scalar>|null
      *
      * @throws IdentifierNotFound
      * @throws Exception\UnableToCompileNode If a referenced constant could not be located on the expected referenced class.
--- a/src/NodeCompiler/CompileNodeToValue.php
+++ b/src/NodeCompiler/CompileNodeToValue.php
@@ -24,7 +24,7 @@ class CompileNodeToValue
      *
      * @param Node\Stmt\Expression|Node\Expr $node Node has to be processed by the PhpParser\NodeVisitor\NameResolver
      *
-     * @return mixed
+     * @return scalar|scalar[]|null
      *
      * @throws Exception\UnableToCompileNode
      */
@@ -60,7 +60,7 @@ class CompileNodeToValue
     /**
      * Compile constant expressions
      *
-     * @return bool|mixed|null
+     * @return scalar|scalar[]|null
      *
      * @throws Exception\UnableToCompileNode
      */
@@ -86,7 +86,7 @@ class CompileNodeToValue
     /**
      * Compile class constants
      *
-     * @return string|int|float|bool|mixed[]|null
+     * @return scalar|scalar[]|null
      *
      * @throws IdentifierNotFound
      * @throws Exception\UnableToCompileNode If a referenced constant could not be located on the expected referenced class.
--- a/src/Reflection/ReflectionClass.php
+++ b/src/Reflection/ReflectionClass.php
@@ -68,16 +68,16 @@ class ReflectionClass implements Reflection
     /** @var ClassLikeNode */
     private $node;
 
-    /** @var ReflectionClassConstant[]|null indexed by name, when present */
+    /** @var array<string, ReflectionClassConstant>|null indexed by name, when present */
     private $cachedReflectionConstants;
 
-    /** @var ReflectionProperty[]|null */
+    /** @var array<string, ReflectionProperty>|null */
     private $cachedImmediateProperties;
 
     /** @var ReflectionProperty[]|null */
     private $cachedProperties;
 
-    /** @var ReflectionMethod[]|null */
+    /** @var array<string, ReflectionMethod>|null */
     private $cachedMethods;
 
     private function __construct()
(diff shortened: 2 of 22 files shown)


src/Reflection/ReflectionProperty.php
```diff
@@ -448,6 +450,8 @@ private function assertClassExist(string $className) : void
      * @throws NoObjectProvided
      * @throws NotAnObject
      * @throws ObjectNotInstanceOfClass
+     * 
+     * @psalm-assert object $object
```
orklah: I think this is a good addition, but it adds an error on psalm because it complains that when this method is called, the variable can only be an object and this is redundant.

I think we should suppress the error and keep the assert.

src/Reflection/StringCast/ReflectionClassConstantStringCast.php
```diff
@@ -16,13 +19,14 @@ final class ReflectionClassConstantStringCast
     public static function toString(ReflectionClassConstant $constantReflection) : string
     {
         $value = $constantReflection->getValue();
+        assert($value === null || is_scalar($value));
```
orklah: A constant can technically return an array. This makes sure that it doesn't happen

src/Reflection/ReflectionClass.php
```diff
@@ -874,7 +874,7 @@ public function isInterface() : bool
      * Get the traits used, if any are defined. If this class does not have any
      * defined traits, this will return an empty array.
      *
-     * @return ReflectionClass[]
+     * @return list<ReflectionClass>
```
orklah: This is discussed here: #530 It raises an error with Psalm that I think is legitimate and I don't know how to handle.

src/Reflection/ReflectionClass.php
```diff
@@ -519,7 +519,7 @@ static function (ReflectionClassConstant $constant) : string {
      * Get an associative array of the defined constants in this class,
      * with keys as constant names and values as {@see ReflectionClassConstant} objects.
      *
-     * @return ReflectionClassConstant[] indexed by name
+     * @return array<string, ReflectionClassConstant> indexed by name
```
orklah: This raise a new error on Psalm too because their internal stubs expects this to be an array<int, ReflectionCl
assConstant>. Not sure who's right on this

src/NodeCompiler/CompileNodeToValue.php
```diff
@@ -60,7 +60,7 @@ public function __invoke(Node $node, CompilerContext $context)
     /**
      * Compile constant expressions
      *
-     * @return bool|mixed|null
+     * @return scalar|scalar[]|null
```
Ocramius: I think we should use `@psalm-` -prefixed annotations for psalm-specific types, to avoid confusing existing tools like @phpstan or @jetbrains

src/Reflection/ReflectionClass.php
```diff
@@ -68,16 +68,16 @@ class ReflectionClass implements Reflection
...
 
-    /** @var ReflectionClassConstant[]|null indexed by name, when present */
+    /** @var array<string, ReflectionClassConstant>|null indexed by name, when present */
     private $cachedReflectionConstants;
 
-    /** @var ReflectionProperty[]|null */
+    /** @var array<string, ReflectionProperty>|null */
     private $cachedImmediateProperties;
 
     /** @var ReflectionProperty[]|null */
```
Ocramius: What's the type of this array's keys?

src/SourceLocator/Ast/FindReflectionsInTree.php
```diff
@@ -32,9 +32,13 @@ final class FindReflectionsInTree
     /** @var FunctionReflector */
     private $functionReflector;
 
-    /** @var Closure */
+    /** @var Closure(): FunctionReflector */
     private $functionReflectorGetter;
 
+    /**
+     * @param AstConversionStrategy $astConversionStrategy
```
Ocramius: Shouldn't be needed

src/SourceLocator/Ast/Locator.php
```diff
@@ -27,6 +28,10 @@ class Locator
     /** @var Parser */
     private $parser;
 
+    /**
+     * @param Parser $parser
```
Ocramius: Shouldn't be needed

orklah:
> Overall excellent work, but we _need_ (unavoidable, sorry :-( ) to use `@psalm-`-prefixed annotations for any types that aren't understood by the wider ecosystem, sorry :S

Will do 👍

src/Reflection/ReflectionClass.php
orklah: I added it. I couldn't deduce it at first so I skipped it.

src/Reflection/ReflectionClass.php
Ocramius: I think the stub may need updating here

Ocramius:
Read through review comments.

Note that the `list` type was added to psalm recently

src/Reflection/ReflectionClass.php
orklah: https://3v4l.org/t05Dv I took this example directly from php.net. The index of array is numerical.
I'll add the case to #530 as it is roughly the same issue

Ocramius: Yeh, the question is whether it is a list or not: I think it is a list in core too

orklah: > Yeh, the question is whether it is a list or not: I think it is a list in core too

I think you got me wrong. The library returns `array<string, ReflectionClassConstant>` right now but native Object return `array<int, ReflectionClassConstant>` (and more probably a `list<ReflectionClassConstant>` but this is not the issue).

I'll silence those errors in psalm to allow the merge, but I think those are legitimate issues.

Ocramius: Oh, careful! We have adapters and then there is this stuff: the adapters are supposed to be compliant with internal definitions, but these classes don't need to be.

orklah: > Oh, careful! We have adapters and then there is this stuff: the adapters are supposed to be compliant with internal definitions, but these classes don't need to be.

oh, I get it. But Psalm sees that this class implements Reflection and warn about the type of the interface being different than declared

orklah:
> Note that the `list` type was added to psalm recently

Yeah but PHPStorm won't read it, so I added @psalm- where possible while keeping @return

I just removed the draft tag. I suppressed or fixed the remaining psalm errors

src/Reflection/ReflectionProperty.php
Ocramius: I think the assertion could be refined further to `@psalm-assert T`, where `T` is the type for this `ReflectionProperty`'s owner, but that's for another time/discussion.

Indeed, psalm assertions are annoying when put in test scope, because they remove obvious/dumb scenarios :D

Ocramius:
> > Note that the `list` type was added to psalm recently
> 
> Yeah but PHPStorm won't read it, so I added @psalm- where possible while keeping @return
> 
> I just removed the draft tag. I suppressed or fixed the remaining psalm errors

This was to note that the stubs may not yet use `list`, since it is a recent addition ;-)

Ocramius:
Thanks @orklah!

orklah:
@Ocramius Thanks for merging :)

I must have broke Psalm on CI though. I added a windows path on xml file. It worked fine locally but it breaks the CI. Can you fix it or should I create a PR?

Ocramius:
Please do send a separate patch when you have time: I'm not actively working on the package these weeks, besides following contributions

PR #532, opened 2020-01-15
fix windows path to linux + phpcbf

orklah:
Fix somes issues introduced in #531 (Just wait for CI to complete before merge, I'll check that Psalm is fixed)


Ocramius:
Thanks for the quick reaction, @orklah

PR #533, opened 2020-01-20
Add compatibility with Composer v2 installed.json files

naderman:
The installed.json is now a map with one entry for packages containing the same former array of packages.


test/unit/SourceLocator/Type/Composer/Factory/MakeLocatorForComposerInstalledJsonTest.php
```diff
@@ -114,6 +115,24 @@ public function expectedLocators() : array
...
+                                $projectComposerV2 . '/vendor/a/b/src/ab_PSR-4_Sources',
+                            ],
+                        ]),
+                        $astLocator
+                    ),
+                    new PsrAutoloaderLocator(
+                        Psr0Mapping::fromArrayMappings([]),
+                        $astLocator
+                    ),
+                    new DirectoriesSourceLocator([], $astLocator),
```
naderman: It's not clear to me why either of the above two loaders exist, particularly the empty Psr0 one. But if anything this seems like a bug unrelated to my change, so I just added them here to the expected result.

Ocramius: This comes from the `"files"` and `"classmap"` configurations, but since you don't have any, it's empty by design.

src/SourceLocator/Type/Composer/Factory/MakeLocatorForComposerJsonAndInstalledJson.php
```diff
@@ -50,17 +50,19 @@ public function __invoke(string $installationPath, Locator $astLocator) : Source
...
             throw FailedToParseJson::inFile($composerJsonPath);
         }
 
-        if (! is_array($installed)) {
+        if (! is_array($installedJson)) {
             throw FailedToParseJson::inFile($installedJsonPath);
         }
 
+        /** @var list<array{name: string, autoload: array{classmap: array<int, string>, files: array<int, string>, psr-4: array<string, array<int, string>>, psr-0: array<string, array<int, string>>}}>|null $installed*/
+        $installed = $installedJson['packages'] ?? $installedJson;
```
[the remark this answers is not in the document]
naderman: Yes that's all that happened. The file is now a map of values, so we can add elements to it in the future without BC breaks. I think the goal was not to add even more metadata files, and rather accept that 2.0 is a BC break anyway and reformat this in a way allowing for future modification without BC breaks. cc @seldaek

test/unit/SourceLocator/Type/Composer/Factory/MakeLocatorForComposerInstalledJsonTest.php
naderman: Would have assumed it would not generate those locators at all then, but okay ;-) Not like I really read more of this code anyway :-D

Ocramius: An empty locator is no-op anyway: be it there or not makes no big difference :+1:

PR #537, opened 2020-02-01
various improvements for psalm

orklah:
I fixed some errors, updated baseline.

It improved type coverage a bit.

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<files psalm-version="3.8.2@90d6b73fd8062432030ef39b7b6694b3902daa31">
+<files psalm-version="3.8.3@389af1bfc739bfdff3f9e3dc7bd6499aee51a831">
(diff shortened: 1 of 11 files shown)


orklah:
oh. I missed the fact that node could be null in ReflectionFunctionAbstract because it can be unitialized.

I'll fix that

orklah:
Mmh... when trying to add null to the list of types for node, Psalm highlight 18 other lines where it's not handled.

I could fix that for Psalm by adding assert($this->node !== null); on each method using it, but I'm not sure it's the best way to go... Any thoughts?

asgrim:
@orklah perhaps a private method that asserts the node is set, and update $this->node references to point to that method?

orklah:
> 
> 
> Nice! Looks good, waiting for build 👍

Please wait. I'm trying to reconcile getNode and getInitializedNode, they both do the same thing. I'm just running tests locally

orklah:
It's done, it should pass CI.

Thanks for the advice!

orklah:
fixed some CS issues and rebased

PR #538, opened 2020-02-05
Bump PHPStan to 0.12.9

ondrejmirtes:
BTW do you know that Travis build is set up so that it ignores all failures on PHP 7.4? So if PHPStan or Psalm fail, the build still passes...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,10 +8,6 @@ before_script:
   - rm composer.lock
   - travis_retry composer update --prefer-dist $DEPENDENCIES
 
-matrix:
-  allow_failures:
-    - php: 7.4
-
 jobs:
   include:
     - stage: Test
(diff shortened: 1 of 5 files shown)


Ocramius:
> BTW do you know that Travis build is set up so that it ignores all failures on PHP 7.4?

Haven't noticed before, urgh :S

Ocramius:
Hmm, fails indeed: can we drop the `.travis.yml` exclusion to make this visible?

ondrejmirtes:
I removed the allowed_failures and ran PHPCBF. Some CS failures still remain...

Ocramius:
@ondrejmirtes I think the remaining issues are fixed in `master`: can you try a rebase here?

Ocramius:
I think @orklah bumped to `0.12.11` already in #548: closing here. Thanks everyone! :tada:

PR #540, opened 2020-02-13
Upgrade reflection-docblock to version ^5.0.0 and upgrade deps

paul-thebaud:
Hi,

This PR upgrade `phpdocumentor/reflection-docblock` to version `5.0.0`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -7,7 +7,7 @@
         "ext-json": "*",
         "jetbrains/phpstorm-stubs": "2019.3",
         "nikic/php-parser": "^4.3.0",
-        "phpdocumentor/reflection-docblock": "^4.3.4",
+        "phpdocumentor/reflection-docblock": "^5.0.0",
         "phpdocumentor/type-resolver": "^1.0.0",
         "roave/signature": "^1.0"
     },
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -7,7 +7,7 @@
         "ext-json": "*",
         "jetbrains/phpstorm-stubs": "2019.3",
         "nikic/php-parser": "^4.3.0",
-        "phpdocumentor/reflection-docblock": "^4.3.4",
+        "phpdocumentor/reflection-docblock": "^5.0.0",
```
asgrim: @Ocramius since there are no BC breaks that seemingly affect us (I have not checked personally but the build will tell all, I'm sure...) should we support `^4.3.4|^5.0` do you think?

Ocramius: Nay, let's bump and ship a new minor, please. Supporting both is not worth it (from a potential bugs perspective)

asgrim: Aight 👍

asgrim:
@paul-thebaud thanks for the update!

Ocramius:
`master` contains a bunch of BC breaks: need to re-target

PR #541, opened 2020-02-17
Provide path to PHP Storm stubs directory

michaeljoelphillips:
Allows for the `stubsDirectory` to be optionally specified during the construction of `PhpStormStubsSourceStubber`, so as to avoid calls to `is_dir` in stream contexts where `is_dir` is not implemented by a stream wrapper.

This resolves a use case where BR is packaged in a phar, but the stubs directory (i.e. `phar:///tmp/app.phar/vendor/jetbrains/phpstorm-stubs`) cannot be located due to `is_dir` returning `false` behind the phar stream wrapper.  Because the stubs directory path is known, the path can be provided to the constructor instead of having to be resolved at runtime.

### DIFF
--- a/src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
+++ b/src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
@@ -80,8 +80,11 @@ final class PhpStormStubsSourceStubber implements SourceStubber
 
     public static function fromStubsDirectory(Parser $phpParser, string $stubsDirectory)
     {
-        $instance = new self($phpParser);
+        if (is_dir($stubsDirectory) === false) {
+            throw CouldNotFindPhpStormStubs::create();
+        }
 
+        $instance = new self($phpParser);
         $instance->stubsDirectory = $stubsDirectory;
 
         return $instance;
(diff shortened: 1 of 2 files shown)


src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
```diff
@@ -66,10 +66,11 @@ final class PhpStormStubsSourceStubber implements SourceStubber
     /** @var array<string, Node\Const_|Node\Expr\FuncCall> */
     private $constantNodes = [];
 
-    public function __construct(Parser $phpParser)
+    public function __construct(Parser $phpParser, ?string $stubsDirectory = null)
```
Ocramius: I would suggest adding a named constructor, instead of changing this constructor. Test additions required.

michaeljoelphillips: Named constructor added.  Let me know if the test additions are sufficient.

test/unit/SourceLocator/SourceStubber/PhpStormStubsSourceStubberTest.php
```diff
@@ -502,4 +504,24 @@ public function testNoStubForUnknownConstant() : void
...
+    }
+
+    public function testStubberThrowsExceptionWhenInvalidStubsDirectoryIsSpecified() : void
+    {
+        $parser = BetterReflectionSingleton::instance()->phpParser();
+        $stubber = PhpStormStubsSourceStubber::fromStubsDirectory($parser, '/tmp/invalid_directory');
+
+        self::expectException(InvalidFileLocation::class);
+
+        $stubber->generateFunctionStub('array_filter');
```
Ocramius: Hmm, looking at this, we should throw an exception during the `fromStubsDirectory` call, not here..

michaeljoelphillips:
Closing, please see #550 for more info.

PR #543, opened 2020-02-18
CS fixes

kukulich:


kukulich:
Please don't merge now - it looks I have to fix also psalm and phpstan errors.

kukulich:
Ok, CS should be fixed. PHPStan does not report anything now and Psalm errors should be unrelated.

Ocramius:
Thanks!

PR #544, opened 2020-02-18
bump psalm, fix errors, add usePhpDocPropertiesWithoutMagicCall

orklah:
Bumping psalm on last version to use usePhpDocPropertiesWithoutMagicCall.

This allows to drop some ignored errors and gain some type coverage.

Also fixed some errors

### DIFF
--- a/src/Reflection/ReflectionProperty.php
+++ b/src/Reflection/ReflectionProperty.php
@@ -338,7 +338,7 @@ class ReflectionProperty
 
         $instance = $this->assertObject($object);
 
-        return Closure::bind(function ($instance, string $propertyName) {
+        return Closure::bind(function (object $instance, string $propertyName) {
             return $instance->{$propertyName};
         }, $instance, $declaringClassName)->__invoke($instance, $this->getName());
     }
--- a/composer.json
+++ b/composer.json
@@ -38,7 +38,7 @@
         "doctrine/coding-standard": "^7.0.2",
         "phpstan/phpstan": "^0.12.8",
         "phpunit/phpunit": "^8.5.2",
-        "vimeo/psalm": "^3.8.3"
+        "vimeo/psalm": "3.9.2"
     },
     "autoload": {
         "psr-4": {
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "e4ca5654db48fa2b4a0b13e9cecbf196",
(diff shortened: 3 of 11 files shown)


orklah:
This PR doesn't fix all current Psalm issues, but after seeing https://github.com/Roave/BetterReflection/pull/543#issuecomment-587856388, I prefer creating it now to avoid doing this work twice

src/Reflection/ReflectionProperty.php
```diff
@@ -317,7 +317,7 @@ public function __clone()
     /**
      * @param object|null $object
      *
-     * @return mixed
+     * @return Closure(object, string): mixed
```
Ocramius: This isn't returning a callable

orklah: Indeed. Must have overlooked __invoke call at the end.

orklah:
This is now error free on Psalm 3.9.2

FYI, I added those errors in baseline:
```
ERROR: InvalidReturnType - src\Reflection\Adapter\ReflectionClass.php:228:21 - The declared return type 'list<ReflectionProperty>' for Roave\BetterReflection\Reflection\Adapter\ReflectionClass::getProperties is incorrect, got 'array
<string, Roave\BetterReflection\Reflection\Adapter\ReflectionProperty>'
    public function getProperties($filter = null)


ERROR: InvalidReturnType - src\Reflection\Adapter\ReflectionObject.php:202:21 - The declared return type 'list<ReflectionProperty>' for Roave\BetterReflection\Reflection\Adapter\ReflectionObject::getProperties is incorrect, got 'arr
ay<string, Roave\BetterReflection\Reflection\Adapter\ReflectionProperty>'
    public function getProperties($filter = null)


ERROR: PossiblyNullPropertyAssignmentValue - src\Reflection\ReflectionFunctionAbstract.php:512:35 -  with non-nullable declared type 'array<array-key, PhpParser\Node\Stmt>' cannot be assigned nullable type 'array<array-key, PhpParse
r\Node\Stmt>|null'
        $this->getNode()->stmts = $functionNode->getStmts();


ERROR: PossiblyNullPropertyAssignmentValue - src\Reflection\ReflectionFunctionAbstract.php:524:35 -  with non-nullable declared type 'array<array-key, PhpParser\Node\Stmt>' cannot be assigned nullable type 'array<array-key, PhpParse
r\Node\Stmt>|null'
        $this->getNode()->stmts = $this->loadStaticParser()->parse('<?php ' . $newBody);
```

First and second one are due to a more precise definition of Reflection library by Psalm. Error type are not compatible between native library and BR (cf #530)

Third and fourth are due to an assignment in Php-Parser from Php-Parser and one side of the assignment could contain null. This is probably due to an error in phpdoc on Php-Parser side.

Ocramius:
Thanks @orklah!

PR #545, opened 2020-02-18
fix CS

orklah:
Sorry, forgot to run phpcbf on previous PR...


Ocramius:
No problem 👍

Sorry for the low cooperation: am only checking this from the phone

PR #546, opened 2020-02-19
Fixed .travis.yml

### LINKED ISSUE #542: broken CI
I don't see how, but it seems that the ci changed lately and it skips a lot of things now.

I couldn't find a commit that could have this effect.

For example, before, when opening Travis, there were a lot of lines, now there is a single execution. Also, the CI time went from 40+ minutes to 2 minutes.

kukulich:
Fixes #542

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,10 +8,6 @@ before_script:
   - rm composer.lock
   - travis_retry composer update --prefer-dist $DEPENDENCIES
 
-matrix:
-  allow_failures:
-    - php: 7.4
-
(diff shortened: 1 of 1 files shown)


kukulich:
It's a little strange that these bugs don't occur in my other repositories but I hope everything should be fixed.

Ocramius:
@kukulich we had a bunch of these as we did some chores on travis-ci. At this point, I think migrating away from it (at some point) is a wise decision: it has been the weakest part of the whole chain.

PR #547, opened 2020-02-19
fix SA

orklah:
This should solve the two errors, one on Psalm and one on PHPStan that fails on CI.

Weirdly, I didn't have the Psalm one yesterday on my machine... (and the error on PHPStan was introduced by me by error :( )


orklah:
Build is failing because we use phpbench ^0.16.9 but we locked symfony/console to 5.0.4 dans phpbench only support it from 0.17

PR #570, opened 2020-05-06
Enhancement: Allow fetching constant name from exception

### LINKED ISSUE #569: Add accessors to UnableToCompileNode exception
For a project @mstaab and I are working on it would be useful if we could fetch the name of a constant when an `Roave\BetterReflection\NodeCompiler\Exception\UnableToCompileNode` exception is thrown:

https://github.com/Roave/BetterReflection/blob/0095a8df72ff99e1b785418e97688ff259108694/src/NodeCompiler/Exception/UnableToCompileNode.php#L44-L54

Currently we use a regular expression to analyze the exception message, retrieve the name of the constant, then define the constant, and retry the operation. 

Before I open a pull request, I wanted to inquire if you were open for 

- adding accessors to `UnableToCompileNode`
- extracting dedicated exceptions extending `UnableToCompileNode` (for backwards compatibility) and adding accessors there

What do you think?

localheinz:
This PR

* [x] allows fetching the constant name from an `UnableToCompileNode` exception

Fixes #569.

### DIFF
--- a/src/NodeCompiler/Exception/UnableToCompileNode.php
+++ b/src/NodeCompiler/Exception/UnableToCompileNode.php
@@ -15,6 +15,14 @@ use function sprintf;
 
 class UnableToCompileNode extends LogicException
 {
+    /** @var string|null */
+    private $constantName;
+
+    public function constantName() : ?string
+    {
+        return $this->constantName;
+    }
+
     public static function forUnRecognizedExpressionInContext(Node\Expr $expression, CompilerContext $context) : self
     {
         return new self(sprintf(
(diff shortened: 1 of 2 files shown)


src/NodeCompiler/Exception/UnableToCompileNode.php
```diff
@@ -15,6 +15,12 @@
 
 class UnableToCompileNode extends LogicException
 {
+    /**
+     * @psalm-readonly
+     * @var string
```
Ocramius, line 20: maybe `string|null`? an empty string seems misleading, and I know it was done mostly because of static analysis 😅

src/NodeCompiler/Exception/UnableToCompileNode.php
```diff
@@ -15,6 +15,12 @@
 
 class UnableToCompileNode extends LogicException
 {
+    /**
+     * @psalm-readonly
+     * @var string|null
+     */
+    public $constantName;
```
Ocramius, line 22: Seems to fail with:

```
ERROR: InaccessibleProperty - src/NodeCompiler/Exception/UnableToCompileNode.php:63:9 - Roave\BetterReflection\NodeCompiler\Exception\UnableToCompileNode::$constantName is marked readonly (see https://psalm.dev/054)

        $exception->constantName = $constantName;
```

I think it should be set in the ctor...

localheinz: Perhaps an accessor is better, then?

`UnableToCompileNode` is not marked as `@internal`, and overriding the constructor might be a backward-compatibility issue (unless I add `$constantName` as an additional parameter.

What do you think?

Ocramius:
Thanks @localheinz!

localheinz:
Thank you, @Ocramius!

PR #572, opened 2020-05-06
Align adapter types to ext reflection collection-ish method types

Ocramius:
Fixes #568 

Potentially a BC break, but I would rather say that here `BetterReflection` is not respecting the types inherited from `ext-reflection`.

@asgrim can I please have your opinion on this? Should I release it under `4.x`, or push `5.0.0`? I'm inclined to call it a `4.x` problem (since we don't control the type)

### DIFF
--- a/src/Reflection/Adapter/ReflectionClass.php
+++ b/src/Reflection/Adapter/ReflectionClass.php
@@ -13,11 +13,12 @@ use Roave\BetterReflection\Reflection\ReflectionClassConstant as BetterReflectio
 use Roave\BetterReflection\Reflection\ReflectionMethod as BetterReflectionMethod;
 use Roave\BetterReflection\Reflection\ReflectionObject as BetterReflectionObject;
 use Roave\BetterReflection\Reflection\ReflectionProperty as BetterReflectionProperty;
-use Webmozart\Assert\Assert;
 use function array_combine;
 use function array_map;
 use function array_values;
+use function assert;
 use function func_num_args;
+use function is_array;
 use function is_object;
 use function is_string;
 use function sprintf;
@@ -333,8 +334,8 @@ class ReflectionClass extends CoreReflectionClass
             }, $traits)
         );
 
-        Assert::isArray(
-            $traitsByName,
+        assert(
+            is_array($traitsByName),
             sprintf(
                 'Could not create an array<trait-string, ReflectionClass> for class "%s"',
                 $this->betterReflectionClass->getName()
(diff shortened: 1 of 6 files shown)


src/Reflection/Adapter/ReflectionClass.php
```diff
@@ -227,14 +230,9 @@ public function getProperty($name)
...
     {
-        $properties = $this->betterReflectionClass->getProperties();
-
-        $wrappedProperties = [];
-        foreach ($properties as $key => $property) {
-            $wrappedProperties[$key] = new ReflectionProperty($property);
-        }
-
-        return $wrappedProperties;
+        return array_values(array_map(static function (BetterReflectionProperty $property): ReflectionProperty {
```
Ocramius, line 233: Note: if you are wondering about tests, this code is already covered by tests. The change is purely type-driven, so no test additions should really be needed.

src/Reflection/Adapter/ReflectionClass.php
```diff
@@ -12,8 +12,11 @@
 use Roave\BetterReflection\Reflection\ReflectionClassConstant as BetterReflectionClassConstant;
 use Roave\BetterReflection\Reflection\ReflectionMethod as BetterReflectionMethod;
 use Roave\BetterReflection\Reflection\ReflectionObject as BetterReflectionObject;
+use Roave\BetterReflection\Reflection\ReflectionProperty as BetterReflectionProperty;
+use Webmozart\Assert\Assert;
```
Ocramius, line 16: Oooof, we don't have the `webmozart/assert` dependency: will get rid of this one.

asgrim:
> @asgrim can I please have your opinion on this? Should I release it under `4.x`, or push `5.0.0`? I'm inclined to call it a `4.x` problem (since we don't control the type)

Yeah, I'd view this as an implementation bug. No idea on whether many people are actually using the adapters, I'd guess not many (since they don't have much real advantage over built-in reflection I guess, beside static reflection), so I figure it probably won't affect (m)any anyway.

src/Reflection/Adapter/ReflectionClass.php
asgrim: Actually, we don't have either assertion library included as a `require` in `composer.json`, wonder if it's worth adding?

Ocramius: probably not for this specific change

asgrim: OK :+1:

Ocramius:
Applied patch to get rid of `Webmozart\\` in `src/` (found another one lurking): will merge once green :+1:

asgrim:
LGTM :shipit:

ondrejmirtes:
> No idea on whether many people are actually using the adapters

I'm a heavy user of the adapters so you'll definitely hear from me if you screw something up ❤️ 👍

ondrejmirtes:
(I hope you see the friendliness in the message, I used a similar tone @Ocramius often uses :))

Ocramius:
heh, sorry in advance if this breaks something. in theory, if you relied on the `ext-reflection` API, you should be fine 👍

Issue #573, opened 2020-05-11
AutoloadSourceLocator::locateClassByName can crash with "Class MemoizingParser not found"


ondrejmirtes:
Steps to reproduce:

1) Instantiate AutoloadSourceLocator with own Parser instance (second constructor argument).
2) Ask it to locate some class reflection.
3) `locateClassByName` gets executed with the stream wrapper magic. `class_exists` leads to executing AutoloadSourceLocator constructor again, but this time, `$betterReflection->phpParser()` gets called.
4) `new MemoizingParser` fails because the class is not known and we're in a special mode with unregistered `file` wrapper so file loading does not work.

Proof of concept fix: https://github.com/ondrejmirtes/BetterReflection/commit/d7e44eeb962165ef489c498319a5ba46e3c51c0b

Ocramius:
Is this reproducible in a test case? Does it require something like `@runInSeparateProcess`?

ondrejmirtes:
Tell me where to put this test case and I'll try it :)

Ocramius:
Anywhere is fine: we can move it afterwards :+1:

ondrejmirtes:
Reproduced: https://github.com/Roave/BetterReflection/pull/575

PR #575, opened 2020-05-12
Reproduce #573

ondrejmirtes:
This reproduces #573. But something's fishy in PHPUnit.

* If I run `vendor/bin/phpunit test/unit/SourceLocator/AutoloadSourceLocatorWithoutParserTest.php`, the test fails.
* If I uncomment the `class_exists()` line, the test passes.
* If I run `vendor/bin/phpunit`, all the tests pass. I'd suspect that `@runInSeparateProcess` doesn't work and something loads the MemoizingParser class before, but:
* If I uncomment `class_exists` and change the assert on the last line to some nonsense and run `vendor/bin/phpunit`, the whole test suite still passes. Looks like PHPUnit is skipping the test and doesn't run it at all when the whole suite is run.

### DIFF
--- a/test/unit/Foo.php
+++ b/test/unit/Foo.php
@@ -1,0 +1,9 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Roave\BetterReflectionTest;
+
+class Foo
+{
+}
--- a/test/unit/SourceLocator/AutoloadSourceLocatorWithoutParserTest.php
+++ b/test/unit/SourceLocator/AutoloadSourceLocatorWithoutParserTest.php
@@ -1,0 +1,40 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Roave\BetterReflectionTest\SourceLocator;
+
+use PhpParser\Lexer\Emulative;
+use PhpParser\ParserFactory;
+use PHPUnit\Framework\TestCase;
+use Roave\BetterReflection\Reflector\ClassReflector;
+use Roave\BetterReflection\Reflector\FunctionReflector;
+use Roave\BetterReflection\SourceLocator\Ast\Locator as AstLocator;
+use Roave\BetterReflection\SourceLocator\Type\AutoloadSourceLocator;
+use Roave\BetterReflectionTest\Foo;
+
+class AutoloadSourceLocatorWithoutParserTest extends TestCase
+{
+    /**
+     * @runInSeparateProcess
+     */
+    public function testWithoutParser() : void
+    {
+    	//class_exists(\Roave\BetterReflection\SourceLocator\Ast\Parser\MemoizingParser::class);
+        $parser            = (new ParserFactory())->create(ParserFactory::PREFER_PHP7, new Emulative([
+            'usedAttributes' => ['comments', 'startLine', 'endLine', 'startFilePos', 'endFilePos'],
+        ]));
+        $functionReflector = null;
+        $astLocator        = new AstLocator($parser, static function () use (&$functionReflector) : FunctionReflector {
+            return $functionReflector;
+        });
+        $sourceLocator     = new AutoloadSourceLocator(
+            $astLocator,
+            $parser
+        );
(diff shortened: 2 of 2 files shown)


Ocramius:
I think I have a potential fix in mind: giving this a try locally.

ondrejmirtes:
This definitely works https://github.com/ondrejmirtes/BetterReflection/commit/d7e44eeb962165ef489c498319a5ba46e3c51c0b

Ocramius:
That fix seems wrong, since it would make any other `AutoloadSourceLocator` share the same php parser...

Ocramius:
@ondrejmirtes is this only reproducible with `MemoizingParser`, or with any class? Was looking at our tests, and we already kinda verify things:

```php
    public function testClassLoads() : void
    {
        $reflector = new ClassReflector(new AutoloadSourceLocator($this->astLocator));

        self::assertFalse(class_exists(ExampleClass::class, false));
        $classInfo = $reflector->reflect(ExampleClass::class);
        self::assertFalse(class_exists(ExampleClass::class, false));

        self::assertSame('ExampleClass', $classInfo->getShortName());
    }

    public function testClassLoadsWorksWithExistingClass() : void
    {
        $reflector = new ClassReflector(new AutoloadSourceLocator($this->astLocator));

        // Ensure class is loaded first
        new ClassForHinting();
        self::assertTrue(class_exists(ClassForHinting::class, false));

        $classInfo = $reflector->reflect(ClassForHinting::class);

        self::assertSame('ClassForHinting', $classInfo->getShortName());
    }
```

ondrejmirtes:
You misunderstand the nature of the bug - the problem is not when reflecting the `MemoizingParser` class, the problem is when reflecting any class while the `MemoizingParser` class is not yet loaded by the PHP process, because AutoloadSourceLocator will try to instantiate it while the class autoloading is basically disabled.

The problem is that with the current state of things, the PHP-Parser instance I pass into `new AutoloadSourceLocator` isn't the one the locator will use for the stream wrapper instance.

My fix does the same thing the AutoloadSourceLocator already does for AstLocator.

I think it behaves fine because multiple AutoloadSourceLocator instances will not overlap for the small moment the `class_exists` handling is underway.

ondrejmirtes:
If you uncomment `class_exists(\Roave\BetterReflection\SourceLocator\Ast\Parser\MemoizingParser::class);` at the beginning of my reproducing test, it will pass because it `class_exists` will trigger autoloading this class.

Ocramius:
> the problem is when reflecting any class while the `MemoizingParser` class is not yet loaded by the PHP process, because `AutoloadSourceLocator` will try to instantiate it while the class autoloading is basically disabled.

Now that makes more sense, thanks!

Ocramius:
What probably needs to happen is:

 1. add a new autoloader
 2. when the autoloader is triggered with the **expected** class name, replace the stream wrapper (otherwise ignore)
 3. restore autoloader

That would ensure that the stream wrapper is only used for the one specific class that we're looking for.

ondrejmirtes:
Sounds good! It'd also solve that any 3rd party autoloader can currently break for the same reason AutoloadSourceLocator now can fail when calling `new MemoizingParser`.

But I still need my own PHP-Parser instance to be used so that it uses the same cache etc. :)

Ocramius:
> But I still need my own PHP-Parser instance to be used so that it uses the same cache etc. :)

I just checked, it doesn't seem like those dependencies are ever used, once the `AutoloadSourceLocator` is used as stream wrapper, so we should be good there!

ondrejmirtes:
What do you mean by dependencies? Which way is it now that you checked? I need this thing for $phpParser as well:

$validLocator = $astLocator ?? self::$currentAstLocator ?? $betterReflection->astLocator();

Ocramius:
Yes, but those aren't used when the class is used as a stream wrapper: https://github.com/Roave/BetterReflection/pull/604

Issue #574, opened 2020-05-11
AutoloadSourceLocator::locateClassByName - does not work with files loaded from PHAR


ondrejmirtes:
You can imagine how long it took me to figure this out :)

Steps to reproduce:

1) Have an autoloader that contains paths to a PHAR, so they start with `phar://`.
2) Use AutoloadSourceLocator with the magic stream wrappers thing.
3) When locating a class, it's possible that `include` is triggered and that `include` is actually executed because it's handled by a `phar` wrapper, not `file` wrapper.
4) It'll probably end up in some fatal error because we're in a special "not actually loading files" mode, the autoloading cannot be triggered recursively again.

So for example:

* BR is asked for reflection of class `A`.
* BR tries to find its parent `B` that's inside a PHAR.
* `include phar://B.php` is actually executed.
* The class `B` extends class `C` also from inside the PHAR.
* Fatal error - class `C` not found although under normal circumstances, it'd be loaded because the autoloader knows about it.

I'm also gonna try to reproduce it although it might require to commit some small PHAR file to `tests`.

Verified proof of concept fix: https://github.com/ondrejmirtes/BetterReflection/commit/e909d2a273e23c96847cbeeb2c1da8846d71a946

ondrejmirtes:
Reproduced: https://github.com/Roave/BetterReflection/pull/576

PR #576, opened 2020-05-12
Reproduce #574

ondrejmirtes:
Reproduces #574.

So, this one was a bit tougher. Without the try-catch and die, it completely breaks PHPUnit because of its error handler I suppose. At least this way, I'm able to get the message I'm experiencing in practice: `Class 'App\B' not found`

What's happening is that:

1) I ask BR for class `App\A`.
2) Clas `App\A` in the PHAR extends something outside of the PHAR.
3) Because `file` stream wrapper is overriden, the call fails and we end up with this message.

The solution to this is to also override `phar://` protocol and not just `file://`. It might be a good idea to make the list of protocols configurable through constructor because I guess this can happen with other protocols if people use them to autoload classes...

What's interesting is that unlike #575 and even if there's `@runInSeparateProcess`, this test runs even when I run the full test suite.


Ocramius:
I'm not sure that (if phars are involved), the `AutoloadSourceLocator` is usable at all: kinda impressed that it still manages to find stuff in phars :O

ondrejmirtes:
With this commit it works :) https://github.com/ondrejmirtes/BetterReflection/commit/e909d2a273e23c96847cbeeb2c1da8846d71a946

As I switched `PHPStan\\` namespace to the static reflection engine yesterday, I was experiencing these failures: https://github.com/phpstan/phpstan/runs/664051706?check_suite_focus=true (because a test case extended RuleTestCase inside `phpstan.phar` which extends PHPUnit's TestCase outside the PHAR). So I know that the fix works.

ondrejmirtes:
If an autoloaded class from inside the PHAR doesn't extend something from outside the PHAR, it works but those classes are actually loaded (because phar:// isn't overriden).

asgrim:
Fix in https://github.com/ondrejmirtes/BetterReflection/commit/e909d2a273e23c96847cbeeb2c1da8846d71a946 seems reasonable! :+1: would that make these added tests pass @ondrejmirtes ?

ondrejmirtes:
Yes, although the code might be nicer with foreaches over an array of protocols :)

asgrim:
Closing this in favour of #582 as your tests have been incorporated there already - thanks @ondrejmirtes !

PR #581, opened 2020-05-22
Fixed many traits bugs

### LINKED ISSUE #577: Trait method aliases broken
Hi,
consider this example: https://3v4l.org/a2hQf

Two observations from that:

1) Aliased trait method fetched on another trait should have the aliased name, not the original name.
2) Class that uses a trait with an aliased method should also know the aliased method.

Unfortunately, BetterReflection breaks both of those assumptions. Here's a reproducing example: https://gist.github.com/ondrejmirtes/a8481b8e03880aebd4e0bae5c6d87e86

kukulich:
Fixes #577

### DIFF
--- a/src/Reflection/ReflectionClass.php
+++ b/src/Reflection/ReflectionClass.php
@@ -38,6 +38,7 @@ use Roave\BetterReflection\Util\GetFirstDocComment;
 use Traversable;
 use function array_combine;
 use function array_filter;
+use function array_key_exists;
 use function array_map;
 use function array_merge;
 use function array_reverse;
@@ -91,6 +92,12 @@ class ReflectionClass implements Reflection
      * @psalm-var ?array<string, ReflectionMethod>
      */
     private $cachedMethods;
+
+    /** @var array<string, string>|null */
+    private $cachedTraitAliases;
+
+    /** @var array<string, string>|null */
+    private $cachedTraitPrecedences;
 
     private function __construct()
     {
@@ -252,18 +259,15 @@ class ReflectionClass implements Reflection
             ),
             ...array_map(
                 function (ReflectionClass $trait) : array {
-                    return array_map(function (ReflectionMethod $method) use ($trait) : ReflectionMethod {
-                        $methodAst = $method->getAst();
-                        assert($methodAst instanceof ClassMethod);
-
-                        return ReflectionMethod::createFromNode(
-                            $this->reflector,
-                            $methodAst,
-                            $this->declaringNamespace,
-                            $trait,
-                            $this
-                        );
-                    }, $trait->getMethods());
+                    return array_merge(
+                        [],
+                        ...array_map(
+                            function (ReflectionMethod $method) : array {
+                                return $this->createMethodsFromTrait($method);
+                            },
+                            $trait->getMethods()
+                        )
+                    );
                 },
                 $this->getTraits()
             ),
@@ -277,6 +281,55 @@ class ReflectionClass implements Reflection
                 ))
             )
         );
+    }
+
+    /**
+     * @return ReflectionMethod[]
+     */
+    private function createMethodsFromTrait(ReflectionMethod $method) : array
+    {
+        $traitAliases     = $this->getTraitAliases();
+        $traitPrecedences = $this->getTraitPrecedences();
+
+        $methodAst = $method->getAst();
+        assert($methodAst instanceof ClassMethod);
+
+        $methodHash = $this->methodHash($method->getDeclaringClass()->getName(), $method->getName());
+
+        $aliases = [];
+        foreach ($traitAliases as $aliasMethodName => $traitAliasDefinition) {
+            if ($methodHash !== $traitAliasDefinition) {
+                continue;
+            }
+
+            $aliases[] = ReflectionMethod::createFromNode(
+                $this->reflector,
+                $methodAst,
+                $method->getDeclaringClass()->getDeclaringNamespaceAst(),
+                $method->getDeclaringClass(),
+                $this,
+                $aliasMethodName
+            );
+        }
+
+        if ($aliases !== []) {
+            return $aliases;
+        }
+
+        if (array_key_exists($methodHash, $traitPrecedences)) {
+            return [];
+        }
+
+        $newMethod = ReflectionMethod::createFromNode(
+            $this->reflector,
+            $methodAst,
+            $method->getDeclaringClass()->getDeclaringNamespaceAst(),
+            $method->getDeclaringClass(),
+            $this,
+            $method->getAliasName()
+        );
+
+        return [$newMethod];
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/Reflection/ReflectionClass.php
```diff
@@ -236,7 +244,10 @@ public function getExtensionName() : ?string
      */
     private function getAllMethods() : array
     {
-        return array_merge(
+        $traitAliases     = $this->getTraitAliases();
+        $traitPrecedences = $this->getTraitPrecedences();
+
+        return array_filter(array_merge(
```
Ocramius, line 250: would put the `array_filter()` only around the logic that fetches methods from traits

src/Reflection/ReflectionClass.php
```diff
@@ -251,19 +262,54 @@ function (ClassMethod $methodNode) : ReflectionMethod {
...
-                            $trait,
-                            $this
-                        );
-                    }, $trait->getMethods());
+                function (ReflectionClass $trait) use ($traitAliases, $traitPrecedences) : array {
+                    return array_merge(
+                        [],
+                        ...array_filter(
+                            array_map(
+                                function (ReflectionMethod $method) use ($traitAliases, $traitPrecedences) : ?array {
```
Ocramius, line 270: this logic now contains looping and more complex checks: we need it to be split out into a `private` method

src/Reflection/ReflectionClass.php
```diff
@@ -251,19 +262,54 @@ function (ClassMethod $methodNode) : ReflectionMethod {
...
+                                    $methodAst = $method->getAst();
+                                    assert($methodAst instanceof ClassMethod);
+
+                                    $aliases = [];
+                                    foreach ($traitAliases as $aliasMethodName => $traitAliasDefinition) {
+                                        if (sprintf('%s::%s', $method->getDeclaringClass()->getName(), $method->getName()) !== $traitAliasDefinition) {
```
Ocramius, line 276: the `%s::%s` approach for lookups seems very fragile: I see that it is pre-existing, but indeed unfortunate storage mechanism.

I would stash this out into a `private function methodHash(ReflectionClass $declaringClass, ReflectionMethod $method)` or such

src/Reflection/ReflectionClass.php
```diff
@@ -984,12 +1015,60 @@ static function (ReflectionClass $trait) : string {
...
+     * class Foo
+     * {
+     *     use MyTrait, MyTrait2 {
+     *         MyTrait2::foo insteadof MyTrait1;
+     *     }
+     * }
+     * // This method would return
+     * //   ['MyTrait1::foo' => 'MyTrait2::foo']
+     */
+    public function getTraitPrecedences() : array
```
Ocramius, line 1043: we should probably **not** expose this API. It is complex enough to require complex examples, and I can't see a use-case for it where downstream would care about trait precedence?

src/Reflection/ReflectionClass.php
Ocramius: Seems to be done in a closure below, called `$formatMethodName`: we could use the same in `private` API

src/Reflection/ReflectionClass.php
```diff
@@ -1001,23 +1080,15 @@ public function getTraitAliases() : array
...
 
-                if (! ($adaptation instanceof Node\Stmt\TraitUseAdaptation\Alias)) {
-                    continue;
-                }
-
-                if (! $adaptation->newName) {
-                    continue;
+                if ($adaptation instanceof Node\Stmt\TraitUseAdaptation\Alias && $adaptation->newName) {
+                    $this->cachedTraitAliases[$adaptation->newName->name] = $formatMethodName($usedTrait, $adaptation->method);
+                } elseif ($adaptation instanceof Node\Stmt\TraitUseAdaptation\Precedence && $adaptation->insteadof) {
```
Ocramius, line 1085: let's use a `continue` rather than an `elseif`

src/Reflection/ReflectionClass.php
```diff
@@ -984,12 +1015,60 @@ static function (ReflectionClass $trait) : string {
...
+        return $this->cachedTraitPrecedences;
+    }
+
+    private function parseTraitUsages() : void
+    {
+        if ($this->cachedTraitAliases !== null && $this->cachedTraitPrecedences !== null) {
+            return;
+        }
+
+        $formatMethodName = static function (Node\Name $className, Identifier $methodName) : string {
```
Ocramius, line 1057: suggested above: let's move this to `private static` API.

necessary to mark it with `@psalm-pure`.

test/unit/Reflection/ReflectionClassTest.php
```diff
@@ -941,26 +954,18 @@ public function testMethodsFromTraits() : void
 
         $classInfo = $reflector->reflect('TraitFixtureC');
 
-        self::assertTrue($classInfo->hasMethod('a'));
+        self::assertFalse($classInfo->hasMethod('a'));
```
Ocramius, line 957: was this an actual bug? very difficult to see it, given the change in the fixtures: maybe split them instead, so that old tests aren't affected?

Ocramius: never mind: original test was indeed wrong here!

Ocramius:
Note: review is mostly around implementation details. Implementation is working, although the trait logic makes it very confusing.

hoping that stashing more details in separate `private` API could make it more clear

kukulich:
@Ocramius Everything resolved (I hope). I will squash everything to one commit after second round of your review.

src/Reflection/ReflectionClass.php
```diff
@@ -279,6 +285,55 @@ static function (ReflectionClass $ancestor) : array {
...
+                $aliasMethodName
+            );
+        }
+
+        if ($aliases !== []) {
+            return $aliases;
+        }
+
+        if (array_key_exists($methodHash, $traitPrecedences)) {
+            return null;
```
Ocramius, line 322: Would it make sense to return `[]` here too? Or is there a specific rationale for returning `null|[]`?

kukulich: There were approximately 50 versions of the code before I sent the PR - I've almost killed @ondrejmirtes for finding more and more strange bugs :)

So it was just a residue.

Removed and squashed.

ondrejmirtes: Don't blame me, blame PHP :)

Ocramius: Blame decades of development in a dynlang :P

kukulich: We also reported one PHP bug https://bugs.php.net/bug.php?id=79605 (already fixed in PHP 8)

Ocramius:
Static analysis passed: not waiting for the rest :D

Ocramius:
Thanks @kukulich!

PR #582, opened 2020-05-22
`AutoloadSourceLocator` should support `phar://` paths as well as `file://`

### LINKED ISSUE #574: AutoloadSourceLocator::locateClassByName - does not work with files loaded from PHAR

kukulich:
Fixes #574
Fixes #576 

Credit to @ondrejmirtes :)


src/SourceLocator/Type/AutoloadSourceLocator.php
```diff
@@ -183,14 +183,16 @@ private function locateClassByName(string $className) : ?string
             stream_wrapper_register($protocol, self::class);
         }
 
-        class_exists($className);
+        try {
+            class_exists($className);
+        } finally {
```
ondrejmirtes, line 188: This is needed when the autoloader throws an exception which would ideally have also a test case.

Also, I later found out that autoloaders that use a `require` instead of `include` are problematic and still end in fatal error. It helped:

1) to use `restore_error_handler()` instead of `set_error_handler($previousErrorHandler)` although I'm not sure why.
2) instead of `return true` in the error handler I'm throwing and catching an exception immediately: https://github.com/phpstan/phpstan-src/commit/83bbc8122a7286b86f01288fa85957f34378814c#diff-b84fb81e70db3cdb6b4495c80b475a44 Another advantage of the exception is that only the first of the registered autoloaders is hit and not all of them. The `require` still throws a warning on older PHP versions but it's fixed in 7.4.2 (https://3v4l.org/CkDDJ). I didn't find a better solution.

ondrejmirtes: (This is what happens with the current code and unsuccessful require: https://3v4l.org/Nt7Ya)

asgrim: > use `restore_error_handler()` instead of `set_error_handler($previousErrorHandler)`

Yeah that seems a bit of an oversight, looks like a reasonable fix

> instead of `return true` in the error handler I'm throwing and catching an exception immediately

wow :vomiting_face: but still, it's all a big smashing hack, so understandable. Let's go with the throwing exception mechanism to solve those edge cases :+1:

kukulich: > use restore_error_handler() instead of set_error_handler($previousErrorHandler)

We cannot use `restore_error_handler()` because of `AutoloadSourceLocator::url_stat()`

> instead of return true in the error handler I'm throwing and catching an exception immediately

@ondrejmirtes See "WIP" commit please. Tests are failing. Did I missed something?

kukulich:
This PR contains also #602 because we need to fix unrelated hidden bugs to have green build here.

I think this PR can be merged so we don't have hidden bugs again. `AutoloadSourceLocator` can be improved in another PR to fix #573.

src/SourceLocator/Type/AutoloadSourceLocator.php
kukulich: `restore_error_handler()` added. See comment below.

src/Reflection/ReflectionClass.php
```diff
@@ -1472,7 +1473,7 @@ public function removeMethod(string $methodName) : bool
     public function addMethod(string $methodName) : void
     {
         $this->node->stmts[] = new ClassMethod($methodName);
-        unset($this->cachedMethods);
```
Ocramius, line 1475: Oh wow, this was ugly, and most likely unintended!

src/Reflection/StringCast/ReflectionClassConstantStringCast.php
```diff
@@ -18,14 +17,13 @@ final class ReflectionClassConstantStringCast
...
         $value = $constantReflection->getValue();
-        assert($value === null || is_scalar($value));
 
         return sprintf(
             "Constant [ %s %s %s ] { %s }\n",
             self::visibilityToString($constantReflection),
             gettype($value),
             $constantReflection->getName(),
-            (string) $value
+            is_array($value) ? 'Array' : (string) $value
```
Ocramius, line 26: This is un-tested: we should verify that:

```

class A {
    const B = ['C'];
}

var_dump((new ReflectionClassConstant(A::class, 'B'))->__toString());
```

Produces

```
string(38) "Constant [ public array B ] { Array }
"
```

src/SourceLocator/Type/AutoloadSourceLocator.php
```diff
@@ -163,18 +174,34 @@ private function locateClassByName(string $className) : ?string
...
-        set_error_handler($previousErrorHandler);
+
+        $this->setErrorHandler();
+
+        foreach ($this->streamWrapperProtocols as $protocol) {
+            stream_wrapper_unregister($protocol);
+            stream_wrapper_register($protocol, self::class);
+        }
+
+        try {
```
Ocramius, line 185: We should verify this `try-finally` by adding a throwing autoloader to the stack.

The test would be something like:

```php
$toBeThrown = new SpecificExceptionType();

spl_autoload_register(static function () use ($toBeThrown) {
    throw $toBeThrown;
});

try {
    // ... try to locate symbol
    self::fail('No exception was thrown');
} catch (SpecificExceptionType $caught) {
    self::assertSame($toBeThrown, $caught);
}

// verify error handler and stream wrapper here
```

src/SourceLocator/Type/AutoloadSourceLocator.php
```diff
@@ -163,18 +174,34 @@ private function locateClassByName(string $className) : ?string
...
-        $previousErrorHandler      = set_error_handler(static function (int $errno, string $errstr) : bool {
-            return true;
-        });
-        stream_wrapper_unregister('file');
-        stream_wrapper_register('file', self::class);
-        class_exists($className);
-        stream_wrapper_restore('file');
-        set_error_handler($previousErrorHandler);
+
+        $this->setErrorHandler();
```
Ocramius, line 178: The internal function is `set_error_handler()`, but what we do on our end is silencing the error handler.

We should therefore call this `->silenceErrors()` or such

test/unit/SourceLocator/Type/AutoloadSourceLocatorTest.php
```diff
@@ -302,4 +308,77 @@ public function autoload(string $className) : bool
...
+        try {
+            (new AutoloadSourceLocator($this->astLocator))
+                ->locateIdentifier(
+                    $this->getMockReflector(),
+                    new Identifier('Whatever', new IdentifierType(IdentifierType::IDENTIFIER_CLASS))
+                );
+
+            self::fail();
+        } finally {
+            spl_autoload_unregister($brokenAutoloader);
```
Ocramius, line 381: We need to verify that the previous stream wrapper and error handler are restored at this point

src/SourceLocator/Type/AutoloadSourceLocator.php
Ocramius: Errata: test already in place, just needs a couple assertions more :+1:

test/unit/SourceLocator/Type/AutoloadSourceLocatorTest.php
```diff
@@ -302,4 +308,77 @@ public function autoload(string $className) : bool
 
         return true;
     }
+
+    /**
+     * @runInSeparateProcess
+     */
+    public function testPharWorks() : void
```
Ocramius, line 315: `testWillLocateSourcesInPharPath`

test/unit/SourceLocator/Type/AutoloadSourceLocatorTest.php
```diff
@@ -302,4 +308,77 @@ public function autoload(string $className) : bool
...
+        });
+
+        $a = new AutoloadableClassInPhar();
+        $this->assertInstanceOf(AutoloadableClassInPhar::class, $a);
+    }
+
+    /**
+     * @runInSeparateProcess
+     */
+    public function testPharDoesNotWork() : void
```
Ocramius, line 333: Wording here needs adjustment - is it verifying that the source is not inside the phar?

test/unit/SourceLocator/Type/PhpInternalSourceLocatorTest.php
```diff
@@ -162,7 +163,7 @@ public function testCanFetchInternalLocatedSourceForConstants(string $constantNa
             $this->getMockReflector(),
             new Identifier($constantName, new IdentifierType(IdentifierType::IDENTIFIER_CONSTANT))
         );
-        assert($reflection instanceof ReflectionFunction);
+        assert($reflection instanceof ReflectionConstant);
```
Ocramius, line 166: Good catch!

Ocramius:
Note: I think most of the new comments are to be handled in #602 rather than here, and instead we just rebase them out.

src/SourceLocator/Type/AutoloadSourceLocator.php
kukulich: I don't know how to test "verify error handler and stream wrapper here"

Both don't have "get" function.

For error handler I can do something like this:

```
        $previousErrorHandler = set_error_handler(function () {
            return true;
        });
        restore_error_handler();
```

However it's very ugly.

Ocramius: The code for the "previous error handler" is exactly like that.

For the stream wrappers, maybe `stream_get_wrappers()`?

kukulich: It only returns the list of registered wrappers.

Ocramius: Guess we can't test that detail then :shrug:

Ocramius: Actually: we can! `file_get_contents(__FILE__)` should work fine if the stream wrapper is gone (it won't if we don't restore it)

kukulich: Added. However it looks that PHPUnit is broken without stream wrapper restore :)

```
Warning: include(P:\Git\better-reflection\vendor\phpunit\phpunit\src\Framework\Error\Warning.php): failed to open stream: "Roave\BetterReflection\SourceLocator\Type\AutoloadSourceLocator::stream_open" call failed
"Roave\BetterReflection\SourceLocator\Type\AutoloadSourceLocator::stream_open" call failed in P:\Git\better-reflection\vendor\composer\ClassLoader.php on line 444

Warning: include(): Failed opening 'P:\Git\better-reflection\vendor\composer/../phpunit/phpunit/src/Framework/Error/Warning.php' for inclusion (include_path='.;C:\php\pear') in P:\Git\better-reflection\vendor\composer\ClassLoader.php on line 444
```

src/SourceLocator/Type/AutoloadSourceLocator.php
kukulich: Renamed.

test/unit/SourceLocator/Type/AutoloadSourceLocatorTest.php
```diff
@@ -302,4 +307,63 @@ public function autoload(string $className) : bool
...
+
+            self::fail('No exception was thrown');
+        } catch (BrokenAutoloaderException $e) {
+            self::assertSame($e, $toBeThrown);
+        } finally {
+            spl_autoload_unregister($brokenAutoloader);
+        }
+
+        self::assertSame($previousErrorHandler, $getErrorHandler());
+        self::assertTrue(file_exists(__FILE__));
```
Ocramius, line 367: This will still work even with the replaced stream wrapper: what won't work (if we don't de-register it) is reading the file contents with `file_get_contents()` :D

test/unit/SourceLocator/Type/AutoloadSourceLocatorTest.php
kukulich: This test was useless so it's gone.

test/unit/SourceLocator/Type/AutoloadSourceLocatorTest.php
kukulich: Renamed.

test/unit/SourceLocator/Type/AutoloadSourceLocatorTest.php
kukulich: Ok. `file_get_contents` is there :)

Ocramius:
Thanks @kukulich and @ondrejmirtes!

Ocramius:
WRONG TAB+ENTER!!!!! DAMMIT GITHUB UI!

PR #585, opened 2020-05-22
Travis is dead, long live GitHub actions

kukulich:
1. I have to disable tests for some extensions because they are broken - see https://github.com/kukulich/BetterReflection/runs/701451918?check_suite_focus=true

2. PHPUnit on Windows needs more memory, see https://github.com/kukulich/BetterReflection/runs/701502676?check_suite_focus=true


ondrejmirtes:
FYI in order to run Actions here in the PR, there has to be a minimal action commited to master, like https://github.com/phpstan/phpstan-src/commit/f59e15b698102765b21af29ec15f199c0200b09f#diff-02d9c370a663741451423342d5869b21

.github/workflows/build.yml
```diff
@@ -0,0 +1,290 @@
...
+    steps:
+      - name: "Checkout"
+        uses: "actions/checkout@v2.0.0"
+
+      - name: "Install PHP"
+        uses: "shivammathur/setup-php@1.8.2"
+        with:
+          coverage: "none"
+          php-version: "${{ matrix.php-version }}"
+          extensions: mbstring
```
asgrim, line 35: If we require `mbstring`, this should probably be in our own `composer.json` :scream:

test/unit/SourceLocator/SourceStubber/ReflectionSourceStubberTest.php
```diff
@@ -164,7 +164,14 @@ static function (string $symbol) : array {
...
                 $allSymbols,
                 static function (string $symbol) : bool {
-                    return (new CoreReflectionClass($symbol))->isInternal();
+                    $reflection = new CoreReflectionClass($symbol);
+
+                    if (! $reflection->isInternal()) {
+                        return false;
+                    }
+
+                    // Some extensions are broken
```
asgrim, line 173: Might be sensible to create an issue for these to revisit when they're working again, and add the issue link here?

.github/workflows/build.yml
```diff
@@ -0,0 +1,290 @@
+# https://help.github.com/en/categories/automating-your-workflow-with-github-actions
```
asgrim, line 1: Rebase onto `master` when #594 is merged, @ondrejmirtes pointed out that should get the build to run in this PR

.github/workflows/build.yml
kukulich: @asgrim mbstring is only for phpunit. And this requirement will be removed with newer phpunit.

test/unit/SourceLocator/SourceStubber/ReflectionSourceStubberTest.php
kukulich: Yes, I plan to revisit it :)

.github/workflows/build.yml
asgrim: Ahh gotcha, thanks for confirming!

.github/workflows/build.yml
asgrim: #594 is merged, please could you rebase now?

kukulich: Rebased.

kukulich:
Rebased and it looks that all actions work!

kukulich:
And it's fast!

kukulich:
Hmm, but Scrutinizer is still waiting for External Code Coverage... I don't know how to fix it :(

Ocramius:
I think we can sunset scrutinizer: we should rather use mutation testing and  `--min-msi`. We can re-introduce coverage (badges) later, but at the moment scrutinizer is lagging back behind in the ecosystem.

Ocramius:
to make this more clear: we add an `infection/infection` test run and we add a `--min-msi` equivalent to the current MSI

.github/workflows/build.yml
```diff
@@ -1,13 +1,290 @@
...
+          coverage: "none"
+          php-version: "${{ matrix.php-version }}"
+          extensions: mbstring
+
+      - name: "Cache dependencies"
+        uses: "actions/cache@v1.1.2"
+        with:
+          path: "~/.composer/cache"
+          key: "php-${{ matrix.php-version }}-composer-${{ hashFiles('**/composer.json') }}"
+          restore-keys: "php-${{ matrix.php-version }}-composer-"
```
Ocramius, line 42: `composer.json` changes rarely: we probably don't care about this cache key/restore key at all?

.github/workflows/build.yml
```diff
@@ -1,13 +1,290 @@
...
+        with:
+          coverage: "none"
+          php-version: "${{ matrix.php-version }}"
+          extensions: mbstring
+
+      - name: "Cache dependencies"
+        uses: "actions/cache@v1.1.2"
+        with:
+          path: "~/.composer/cache"
+          key: "php-${{ matrix.php-version }}-composer-${{ hashFiles('**/composer.json') }}"
```
Ocramius, line 84: same as above: composer cache doesn't need to be bound to `composer.json`

.github/workflows/build.yml
```diff
@@ -1,13 +1,290 @@
...
+      - name: "Install lowest dependencies"
+        if: ${{ matrix.dependencies == 'lowest' }}
+        run: "composer update --prefer-lowest --prefer-dist --no-interaction --no-progress --no-suggest"
+
+      - name: "Install highest dependencies"
+        if: ${{ matrix.dependencies == 'highest' }}
+        run: "composer update --prefer-dist --no-interaction --no-progress --no-suggest"
+
+      - name: "Tests"
+        run: "php -dmemory_limit=-1 vendor/phpunit/phpunit/phpunit"
```
Ocramius, line 53: why not `php vendor/bin/phpunit`? relying on the package path seems very fragile here

.github/workflows/build.yml
```diff
@@ -1,13 +1,290 @@
...
+      - name: "Install dependencies"
+        run: "composer install"
+
+      - name: "Install PhpBench"
+        run: "composer require --dev --prefer-dist --prefer-stable --no-interaction --no-progress --no-suggest phpbench/phpbench:^0.17.0"
+
+      - name: "PhpBench"
+        run: "vendor/bin/phpbench run --progress=dots --iterations=1"
+
+  code-coverage:
```
Ocramius, line 254: let's replace this with an infection run, and I'll disable scrutinizer 👍

test/unit/SourceLocator/SourceStubber/ReflectionSourceStubberTest.php
Ocramius: yes, but an issue ID is needed in this TODO

test/unit/SourceLocator/Type/PhpInternalSourceLocatorTest.php
```diff
@@ -103,7 +103,12 @@ static function (string $symbol) : array {
                 static function (string $symbol) : bool {
                     $reflection = new CoreReflectionClass($symbol);
 
-                    return $reflection->isInternal();
+                    if (! $reflection->isInternal()) {
+                        return false;
+                    }
+
+                    // Some extensions are broken
```
Ocramius, line 110: similar to the above: we need an issue ID in this comment

ondrejmirtes:
BTW Codecov pairs well with PHPUnit, PCOV and GitHub Actions.

Ocramius:
yeah, just more interesting to set thresholds around mutations than just having thr badge 😅

asgrim:
> yeah, just more interesting to set thresholds around mutations than just having thr badge 😅

Badges 4 lyf

.github/workflows/build.yml
kukulich: This is content of the file on Windows:

```
#!/usr/bin/env sh

dir=$(cd "${0%[/\\]*}" > /dev/null; cd "../phpunit/phpunit" && pwd)

if [ -d /proc/cygdrive ]; then
    case $(which php) in
        $(readlink -n /proc/cygdrive)/*)
            # We are in Cygwin using Windows php, so the path must be translated
            dir=$(cygpath -m "$dir");
            ;;
    esac
fi

"${dir}/phpunit" "$@" 
```

So we really need the PHP file :(

Ocramius: weeeeiiiirrrrd :| 

Ok, understandable :+1:

test/unit/SourceLocator/SourceStubber/ReflectionSourceStubberTest.php
kukulich: Added.

.github/workflows/build.yml
kukulich: Done.

.github/workflows/build.yml
kukulich: Done.

kukulich:
@Ocramius I've removed scrutinizer but I'm not familiar with Infection so it would be better someone else can add it.

This can be merged I think. It will speedup every other PR.

Ocramius:
Reported #599 to add infection: I'll do it, no worries :-)

Ocramius:
Thanks, @kukulich!

PR #586, opened 2020-05-24
Fixed ReflectionMethod::isAbstract() in adapter

kukulich:
See https://3v4l.org/sdhSP

### DIFF
--- a/src/Reflection/Adapter/ReflectionMethod.php
+++ b/src/Reflection/Adapter/ReflectionMethod.php
@@ -273,7 +273,7 @@ class ReflectionMethod extends CoreReflectionMethod
      */
     public function isAbstract()
     {
-        return $this->betterReflectionMethod->isAbstract();
+        return $this->betterReflectionMethod->isAbstract() || $this->betterReflectionMethod->getDeclaringClass()->isInterface();
     }
 
     /**
--- a/test/unit/Reflection/Adapter/ReflectionMethodTest.php
+++ b/test/unit/Reflection/Adapter/ReflectionMethodTest.php
@@ -377,4 +377,24 @@ class ReflectionMethodTest extends TestCase
         $this->expectException(CoreReflectionException::class);
         $reflectionMethodAdapter->invokeArgs();
     }
+
+    public function testIsAbstractWhenDeclaringClassIsInterface() : void
+    {
+        $betterReflectionClass = $this->createMock(BetterReflectionClass::class);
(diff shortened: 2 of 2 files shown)


src/Reflection/Adapter/ReflectionMethod.php
```diff
@@ -273,7 +273,7 @@ public function isProtected()
      */
     public function isAbstract()
     {
-        return $this->betterReflectionMethod->isAbstract();
+        return $this->betterReflectionMethod->isAbstract() || $this->betterReflectionMethod->getDeclaringClass()->isInterface();
```
asgrim, line 276: @kukulich I think this should probable be fixed in `\Roave\BetterReflection\Reflection\ReflectionClass::isAbstract` too, perhaps?

asgrim:
Oops, merged the wrong MR :roll_eyes: @kukulich see comment above :grin:

src/Reflection/Adapter/ReflectionMethod.php
kukulich: It looks the current version is right: https://3v4l.org/EBjjL

Ocramius: wow, PHP is confusing 🤣

asgrim: Oh, I made a mistake, I meant the BR ReflectionMethod @kukulich 🙄 sorry

Ocramius: Reported in #596

kukulich: @asgrim I'm not sure. BR is "better" so it should maybe report only really `abstract` methods...

But there's `ReflectionMethod::getModifiers()` based also on `isAbstract()`...

You tell me how I should fix it :D

ondrejmirtes: I wasn't kidding when I said that PHPStan is an integration test suite for BetterReflection :D https://twitter.com/OndrejMirtes/status/1261621042176548865

Ocramius: > You tell me how I should fix it :D

Oooooooooof

Guess we'll need more tests, but if you have a bug or scenario in mind, report a new issue referencing this thread.

The test suite and code are in good shape enough to "outsource" all of this work to the twitters :-)

PR #593, opened 2020-05-25
Removed outdated @psalm-suppress

kukulich:


orklah:
You should also be able to update psalm's baseline. That should remove two errors that were added here: https://github.com/Roave/BetterReflection/pull/544#issuecomment-587893374 basically for the same reason. That was fixed by Marco too.

kukulich:
@orklah Thanks, updated.

PR #594, opened 2020-05-26
Add hello world github action for #585 to base off

asgrim:
HT to @ondrejmirtes who suggested that for #585 to build in the PR, the `master` branch needs already GH actions enabled. So this adds this.


asgrim:
Build in Travis is green, status update must be delayed :roll_eyes:

PR #595, opened 2020-05-26
Psalm.xml cleanup

orklah:
This PR removes some suppressed errors in psalm.xml.

Those errors seem to not exists anymore, so this will prevent new similar errors to be introduced in future commits.


asgrim:
Travis is green, status not updated :roll_eyes:

PR #597, opened 2020-05-27
ReflectionMethod::isAbstract() should return true for interface methods

### LINKED ISSUE #596: Verify implementation of `ReflectionMethod#isAbstract()`
Ref https://github.com/Roave/BetterReflection/pull/586#discussion_r430921287

BetterReflection does not seem to mark reflection methods as abstract when they are declared in interfaces, but not defined in concrete classes.

kukulich:
Fixes #596


Ocramius:
Not waiting for coverage build here: otherwise :heavy_check_mark:

PR #601, opened 2020-05-27
Methods are case insensitive, and should be mapped and looked up without considering casing

### LINKED ISSUE #588: `ReflectionClass::hasMethod()`, `getMethod()` etc should be case insensitive
Only adapter is case insensitive now but there are strange bugs like https://github.com/Roave/BetterReflection/compare/master...kukulich:case-insensitive-methods (it returns two methods now).

I can prepare PR if you agree.

kukulich:
Fixes #588


Ocramius:
Thanks, @kukulich! Excellent work!

PR #604, opened 2020-05-27
#573 ensure `AutoloadSourceLocator` does not interfere with autoloading of PHP-Parser nor BetterReflection internals

### LINKED ISSUE #573: AutoloadSourceLocator::locateClassByName can crash with "Class MemoizingParser not found"

Ocramius:
Fixes #573
Fixes #575

Removed stream wrapper logic from `AutoloadSourceLocator`: using the new `FileReadTrapStreamWrapper` for that
    
This simplifies the `AutoloadSourceLocator` massively:
    
 * the constructor no longer needs to be zero-arguments (although not changed that, for BC compliance)
 * internal state slimmed down
 * magic around the object is "mostly" gone
 * allows using custom AST locators internally

### DIFF
--- a/src/SourceLocator/Type/AutoloadSourceLocator.php
+++ b/src/SourceLocator/Type/AutoloadSourceLocator.php
@@ -159,6 +159,12 @@ class AutoloadSourceLocator extends AbstractSourceLocator
      * that it cannot find the file, so we squelch the errors by overriding the
      * error handler temporarily.
      *
+     * Note: the following code is designed so that the first hit on an actual
+     *       **file** leads to a path being resolved. No actual autoloading nor
+     *       file reading should happen, and most certainly no other classes
+     *       should exist after execution. The only filesystem access is to
+     *       check whether the file exists.
+     *
      * @throws ReflectionException
      */
     private function locateClassByName(string $className) : ?string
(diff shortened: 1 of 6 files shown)


src/SourceLocator/Type/AutoloadSourceLocator.php
```diff
@@ -249,7 +225,10 @@ private function locateConstantByName(string $constantName) : ?string
 
         $constantFileName = null;
 
-        foreach (get_included_files() as $includedFileName) {
+        // Note: looking at files in reverse order, since newer files are more likely to have
```
Ocramius, line 228: This changes the test run from ~7s per test to ~1s per test related to constant loading :muscle:

Ocramius:
@kukulich can you review this? I'd then be able to release (if it looks good / after adjustments)

kukulich:
LGTM 👍 

I've ask @ondrejmirtes if he can test it in PHPStan too.

Ocramius:
Merging then :+1: 

Thanks for the review!

ondrejmirtes:
Please wait for my testing before releasing this. This is open-source, there's no need to hurry :)

Ocramius:
Yep, not releasing yet :+1: 

I'm still working on mutation tests (which already found few bugs :D )

ondrejmirtes:
@Ocramius Just tested it, works perfectly! 👍

Ocramius:
Will cut a release then 👍

PR #606, opened 2020-05-27
#599 add infection (mutation testing) to test runs

### LINKED ISSUE #599: Add infection with minimum mutation score indicator to test runs
https://github.com/Roave/BetterReflection/pull/585#issuecomment-634478630

Ocramius:
Fixes #599


src/SourceLocator/Type/AutoloadSourceLocator.php
```diff
@@ -229,6 +230,10 @@ private function locateConstantByName(string $constantName) : ?string
         //       defined a constant that is being looked up. Earlier files are possibly related
         //       to libraries/frameworks that we rely upon.
         foreach (array_reverse(get_included_files()) as $includedFileName) {
+            if (! is_readable($includedFileName)) {
```
kukulich, line 233: Should not be better to use existing FileChecker?

Ocramius: Good call!

Ocramius: Actually, `FileChecker` operates with exceptions: not sure how heavy it could be, but it's probably not the hot path...

.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,62 @@
+# https://help.github.com/en/categories/automating-your-workflow-with-github-actions
```
Ocramius, line 1: My eyes were spinning too much, so I split the workflows into multiple files that I could compare with `diff` without going bananas.

Tried using anchors, but they aren't supported by GitHub's server-side parser.

composer.json
```diff
@@ -3,7 +3,7 @@
     "description": "Better Reflection - an improved code reflection API",
     "license": "MIT",
     "require": {
-        "php": ">=7.2.0,<7.5.0",
+        "php": ">=7.3.0,<7.5.0",
```
Ocramius, line 6: PHP 7.2.0 is gone: it isn't supported by `infection/infection`, so I'd rather have it gone, and have this target `4.4.0` of this repo instead.

Ocramius:
@ondrejmirtes I got this interesting error with this new patch:

```
  -- --------------------------------------------------------------------------- 
     Error                                                                      
 -- --------------------------------------------------------------------------- 
     Reached internal errors count limit of 50, exiting...                      
     Internal error: Child process timed out after 60.0 seconds. Try making it  
     longer with parallel.processTimeout setting.                               
 -- ---------------------------------------------------------------------------
```

Do you have a clue on what may be going on? It's running the phar version (`0.12.25`), which didn't change with the patch itself.

ondrejmirtes:
I sent https://github.com/Roave/BetterReflection/pull/609. This might have been a temporary hiccup, but this file should be excluded in the current stable version. Anyway. On my computer and on @kukulich's the analysis doesn't finish even with a lot of RAM and with this file excluded it doesn't have any problems.

Ocramius:
Gonna cut 4.3.0 as soon as I have a moment, will then merge and release with this too 👍

.github/workflows/psalm.yml
```diff
@@ -0,0 +1,60 @@
...
+
+      - name: "Install highest dependencies"
+        if: ${{ matrix.dependencies == 'highest' }}
+        run: "composer update --no-interaction --no-progress --no-suggest"
+
+      - name: "Install locked dependencies"
+        if: ${{ matrix.dependencies == 'locked' }}
+        run: "composer install --no-interaction --no-progress --no-suggest"
+
+      - name: "PHPStan"
```
albe, line 59:
```suggestion
      - name: "psalm"
```

Ocramius:
`4.3.0` is out: this can now be merged \o/

.github/workflows/psalm.yml
Ocramius: Integrated BTW, thanks @albe!

PR #609, opened 2020-05-28
Exclude PhpStormStubsSourceStubber temporarily from the PHPStan analysis

ondrejmirtes:
This solves problem mentioned in https://github.com/Roave/BetterReflection/pull/606#issuecomment-634989856

This is already optimized in dev-master but it will take some time for me to implement rules that we lose by starting using static reflection and release 0.12.26 so it's a good idea to do this meanwhile.

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -4,6 +4,11 @@ parameters:
     paths:
         - src
 
+    excludes_analyse:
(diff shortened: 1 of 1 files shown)


Ocramius:
Reported #620 to revert this once it is time to do so

PR #608, opened 2020-05-28
PHP 7.4 is required

kukulich:
Fixes #571

PR #610, opened 2020-05-28
Fixed `ReflectionClass::isInstance()` adapter compatibility with native reflection

kukulich:
https://3v4l.org/J1jl9

### DIFF
--- a/src/Reflection/Adapter/ReflectionClass.php
+++ b/src/Reflection/Adapter/ReflectionClass.php
@@ -8,6 +8,7 @@ use InvalidArgumentException;
 use OutOfBoundsException;
 use ReflectionClass as CoreReflectionClass;
 use ReflectionException as CoreReflectionException;
+use Roave\BetterReflection\Reflection\Exception\NotAnObject;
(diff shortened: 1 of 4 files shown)


kukulich:
Reported as PHP bug: https://bugs.php.net/bug.php?id=79645

Ocramius:
This throws a `TypeError` in PHP 8: shouldn't we align to that behaviour, maybe? Seems more sensible to me...

kukulich:
> This throws a `TypeError` in PHP 8: shouldn't we align to that behaviour, maybe? Seems more sensible to me...

I would change it in version 4.4.0 that will require PHP 7.4.1. It would be easier: https://github.com/Roave/BetterReflection/pull/608/commits/bcd55b2ba9bd3451fba0f1b499e7ce0856910fab#diff-90219491fdbdad675bcb14b0405dc888R385

kukulich:
And probably there are more methods that should `TypeError` then.

Ocramius:
Let's hold off with PHP 8 compatibility then: it will come anyway 👍

Ocramius:
Thanks @kukulich!

PR #618, opened 2020-05-28
PhpStormStubsSourceStubber fixes for case-insensitive classes/functions/constants

### LINKED ISSUE #611: PhpStormStubsSourceStubber should be case-insensitive for classes and functions
Asking for `SOAPFault` and `SoapFault` should yield the same results.

PoC fix: https://github.com/ondrejmirtes/BetterReflection/commit/3105728871436e1c34c86db23f6dfe711fd9c12d

kukulich:
Fixes #611

### DIFF
--- a/src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
+++ b/src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
@@ -26,7 +26,6 @@ use function count;
 use function defined;
 use function explode;
 use function file_get_contents;
-use function in_array;
 use function is_dir;
 use function is_string;
 use function sprintf;
@@ -68,7 +67,11 @@ final class PhpStormStubsSourceStubber implements SourceStubber
     /** @var array<string, Node\Stmt\Function_> */
     private $functionNodes = [];
 
-    /** @var array<string, Node\Stmt\Const_|Node\Expr\FuncCall> */
+    /**
+     * `null` means "failed lookup" for constant that is not case insensitive
+     *
+     * @var array<string, Node\Stmt\Const_|Node\Expr\FuncCall|null>
+     */
     private $constantNodes = [];
 
     /** @var array<lowercase-string, string> */
@@ -76,6 +79,9 @@ final class PhpStormStubsSourceStubber implements SourceStubber
 
     /** @var array<lowercase-string, string> */
     private $functionMap;
+
+    /** @var array<lowercase-string, string> */
+    private $constantMap;
 
     public function __construct(Parser $phpParser)
     {
@@ -91,6 +97,7 @@ final class PhpStormStubsSourceStubber implements SourceStubber
 
         $this->classMap    = array_change_key_case(PhpStormStubsMap::CLASSES);
         $this->functionMap = array_change_key_case(PhpStormStubsMap::FUNCTIONS);
+        $this->constantMap = array_change_key_case(PhpStormStubsMap::CONSTANTS);
     }
 
     public function generateClassStub(string $className) : ?StubData
@@ -136,22 +143,35 @@ final class PhpStormStubsSourceStubber implements SourceStubber
 
     public function generateConstantStub(string $constantName) : ?StubData
     {
-        // https://github.com/JetBrains/phpstorm-stubs/pull/591
-        if (in_array($constantName, ['TRUE', 'FALSE', 'NULL'], true)) {
-            $constantName = strtolower($constantName);
-        }
+        $lowercaseConstantName = strtolower($constantName);
 
-        if (! array_key_exists($constantName, PhpStormStubsMap::CONSTANTS)) {
+        if (! array_key_exists($lowercaseConstantName, $this->constantMap)) {
             return null;
         }
 
-        $filePath = PhpStormStubsMap::CONSTANTS[$constantName];
-
-        if (! array_key_exists($constantName, $this->constantNodes)) {
-            $this->parseFile($filePath);
+        if (array_key_exists($lowercaseConstantName, $this->constantNodes)
+            && $this->constantNodes[$lowercaseConstantName] === null
+        ) {
+            return null;
         }
 
-        return new StubData($this->createStub($this->constantNodes[$constantName]), $this->getExtensionFromFilePath($filePath));
+        $filePath     = $this->constantMap[$lowercaseConstantName];
+        $constantNode = $this->constantNodes[$constantName] ?? $this->constantNodes[$lowercaseConstantName] ?? null;
+
+        if ($constantNode === null) {
+            $this->parseFile($filePath);
+
+            $constantNode = $this->constantNodes[$constantName] ?? $this->constantNodes[$lowercaseConstantName] ?? null;
+
+            if ($constantNode === null) {
+                // Still `null` - the constant is not case-insensitive. Save `null` so we don't parse the file again for the same $constantName
+                $this->constantNodes[$lowercaseConstantName] = null;
+
+                return null;
+            }
+        }
+
+        return new StubData($this->createStub($constantNode), $this->getExtensionFromFilePath($filePath));
     }
 
     private function parseFile(string $filePath) : void
--- a/src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
+++ b/src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
@@ -18,6 +18,7 @@ use Roave\BetterReflection\SourceLocator\FileChecker;
 use Roave\BetterReflection\SourceLocator\SourceStubber\Exception\CouldNotFindPhpStormStubs;
 use Roave\BetterReflection\Util\ConstantNodeChecker;
 use Traversable;
+use function array_change_key_case;
 use function array_key_exists;
 use function assert;
 use function constant;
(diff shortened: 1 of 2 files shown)


src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
```diff
@@ -81,21 +90,27 @@ public function __construct(Parser $phpParser)
         $this->nodeTraverser = new NodeTraverser();
         $this->nodeTraverser->addVisitor(new NameResolver());
         $this->nodeTraverser->addVisitor($this->cachingVisitor);
+
+        self::$classMap    = array_change_key_case(PhpStormStubsMap::CLASSES);
+        self::$functionMap = array_change_key_case(PhpStormStubsMap::FUNCTIONS);
+        self::$constantMap = array_change_key_case(PhpStormStubsMap::CONSTANTS);
```
Ocramius, line 96: I was looking at this and being confused: what is the casing?

PHP's default values are annoying...

src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
```diff
@@ -107,37 +122,53 @@ public function generateClassStub(string $className) : ?StubData
...
-            $constantName = strtolower($constantName);
+        $lowercaseConstantName = strtolower($constantName);
+
+        if (! array_key_exists($lowercaseConstantName, self::$constantMap)) {
+            return null;
         }
 
-        if (! array_key_exists($constantName, PhpStormStubsMap::CONSTANTS)) {
+        if (array_key_exists($lowercaseConstantName, $this->constantNodes)
+            && $this->constantNodes[$lowercaseConstantName] === null
```
Ocramius, line 149: why cam this be `null`? I see the type was expanded for that...

src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
```diff
@@ -107,37 +122,53 @@ public function generateClassStub(string $className) : ?StubData
...
-        $filePath = PhpStormStubsMap::CONSTANTS[$constantName];
+        $filePath     = self::$constantMap[$lowercaseConstantName];
+        $constantNode = $this->constantNodes[$constantName] ?? $this->constantNodes[$lowercaseConstantName] ?? null;
 
-        if (! array_key_exists($constantName, $this->constantNodes)) {
+        if ($constantNode === null) {
             $this->parseFile($filePath);
+
+            if (! array_key_exists($constantName, $this->constantNodes)
+                && ! array_key_exists($lowercaseConstantName, $this->constantNodes)
```
Ocramius, line 161: would it be better to index `$this->constantNodes` by `lowercase-string` instead of performing a double lookup here?

src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
```diff
@@ -107,37 +122,53 @@ public function generateClassStub(string $className) : ?StubData
...
+            ) {
+                $this->constantNodes[$lowercaseConstantName] = null;
+
+                return null;
+            }
+
+            $constantNode = $this->constantNodes[$constantName] ?? $this->constantNodes[$lowercaseConstantName];
```
Ocramius, line 168: this seems redundant with the above conditionals

src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
kukulich: Constants are not case insensitive by default. They are case insensitive only if you use third parameter of `define`.

So this is optimization (line 165) if do more searches for the same constant that is not case insensitive.

src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
kukulich: Constants are not case insensitive by default. They are case insensitive only if you use third parameter of `define`.

That's why I try both versions.

src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
Ocramius: That still doesn't clarify the `null` scenario 🤔

src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
kukulich: It's not. I have to update `$constantNode` after the `$this->parseFile()` call.

src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
Ocramius: makes sense. wondering if we can make this more explicit by stashing things into a `private` method, or by adding at least a comment...

src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
Ocramius: Yes, but the conditional right above does the same kind of lookup? Unless existence vs `null` are wildly different, at which point we need some more clarity in the data structure (`null` scenario)

src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
kukulich: Eg. `DATE_ATOM` is not case insensitive.

First search for `date_atom` will call `->parseFile()`. It will save `date_atom` as `null` so second search for `date_atom` will not parse file again.

src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
kukulich: Added comment :)

kukulich: I hope it solve everything :D

src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
```diff
@@ -67,9 +67,18 @@ final class PhpStormStubsSourceStubber implements SourceStubber
     /** @var array<string, Node\Stmt\Function_> */
     private $functionNodes = [];
 
-    /** @var array<string, Node\Stmt\Const_|Node\Expr\FuncCall> */
+    /** @var array<string, Node\Stmt\Const_|Node\Expr\FuncCall|null> */
```
Ocramius, line 70: Could you please add a comment describing `null` here?

src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
```diff
@@ -107,37 +122,52 @@ public function generateClassStub(string $className) : ?StubData
...
+        $constantNode = $this->constantNodes[$constantName] ?? $this->constantNodes[$lowercaseConstantName] ?? null;
 
-        if (! array_key_exists($constantName, $this->constantNodes)) {
+        if ($constantNode === null) {
             $this->parseFile($filePath);
+
+            $constantNode = $this->constantNodes[$constantName] ?? $this->constantNodes[$lowercaseConstantName] ?? null;
+
+            if ($constantNode === null) {
+                // Still `null` - the constant is not case-insensitive. Save `null` so we don't parse the file again for the same $constantName
```
Ocramius, line 163: I think clarifying that `null` means "failed lookup" in the data structure is sufficient

src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
kukulich: Done.

src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
```diff
@@ -81,21 +94,27 @@ public function __construct(Parser $phpParser)
         $this->nodeTraverser = new NodeTraverser();
         $this->nodeTraverser->addVisitor(new NameResolver());
         $this->nodeTraverser->addVisitor($this->cachingVisitor);
+
+        self::$classMap    = array_change_key_case(PhpStormStubsMap::CLASSES);
```
Ocramius, line 98: this is done at every instantiation: since that will happen once (usually), we probably should use instance variables anyway, so we simply object lifecycle semantics

test/unit/SourceLocator/SourceStubber/PhpStormStubsSourceStubberTest.php
```diff
@@ -562,4 +563,99 @@ public function testNoStubForUnknownConstant() : void
...
+     * @dataProvider dataCaseInsensitiveFunction
+     */
+    public function testCaseInsensitiveFunction(string $functionName, string $expectedFunctionName) : void
+    {
+        $functionReflection = $this->functionReflector->reflect($functionName);
+
+        $this->assertSame($expectedFunctionName, $functionReflection->getName());
+    }
+
+    public function dataCaseInsensitiveConstant() : array
```
Ocramius, line 615: Is there a scenario around case-sensitive constants (that shouldn't be available if looked up with the wrong casing)?

kukulich: Yes, all tested constants are case insensitive so `IdentifierNotFound` is thrown.

src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php

Ocramius:
Thanks @kukulich!

PR #621, opened 2020-05-28
Annotations cleanup

kukulich:
PHPCS (thanks to Slevomat CS), PHPStan and Psalm can parse standard annotations even with extended syntax so I think we can use less specific `@psalm` annotations.

### DIFF
--- a/src/Reflection/ReflectionClass.php
+++ b/src/Reflection/ReflectionClass.php
@@ -69,28 +69,16 @@ class ReflectionClass implements Reflection
     /** @var ClassLikeNode */
     private $node;
 
-    /**
-     * @var ReflectionClassConstant[]|null indexed by name, when present
-     * @psalm-var ?array<string, ReflectionClassConstant>
-     */
+    /** @var array<string, ReflectionClassConstant>|null indexed by name, when present */
     private $cachedReflectionConstants;
 
-    /**
-     * @var ReflectionProperty[]|null
-     * @psalm-var ?array<string, ReflectionProperty>
-     */
+    /** @var array<string, ReflectionProperty>|null */
     private $cachedImmediateProperties;
 
-    /**
-     * @var ReflectionProperty[]|null
-     * @psalm-var ?array<string, ReflectionProperty>
-     */
+    /** @var array<string, ReflectionProperty>|null */
     private $cachedProperties;
 
-    /**
-     * @var ReflectionMethod[]|null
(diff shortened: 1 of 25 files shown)


src/Reflection/ReflectionClass.php
```diff
@@ -69,28 +69,16 @@ class ReflectionClass implements Reflection
     /** @var ClassLikeNode */
     private $node;
 
-    /**
-     * @var ReflectionClassConstant[]|null indexed by name, when present
-     * @psalm-var ?array<string, ReflectionClassConstant>
-     */
+    /** @var array<string, ReflectionClassConstant>|null indexed by name, when present */
```
Ocramius, line 72: these break IDE support, sadly 😩

kukulich: However there's a lot of `<>` syntax in master in `@var`, `@param` annotations :)

Ocramius: Yeah, true that, but I'll probably end up adding `Foo[]` in multiple locations as I go through the code 😬

kukulich: I've just found https://plugins.jetbrains.com/plugin/12754-phpstan--psalm--generics

Ocramius: Tried that few weeks ago: not sure if it works now 😅

Ocramius: Ok, after some thought put into this: the folks at @jetbrains will never get moving if we aren't the ones pushing it.

I'm fine with sacrificing some auto-completion to push the boundaries of what's done at type-level within our own annotations :+1:

Ocramius:
Thanks, @kukulich!

src/Reflection/ReflectionClass.php
simPod: The plugin works fine for several use cases. Though eg. support for iterables is not ready yet.

Folks at Jetbrains are not planning to add support for these annotations anytime soon AFAIK :/

iluuu1994: [WI-47158](https://youtrack.jetbrains.com/issue/WI-47158) has been the [most upvoted PHPStorm feature request ever](https://youtrack.jetbrains.com/issues?q=project:PhpStorm%20order%20by:%20votes) for a while and there's still no response from @JetBrains...

PR #626, opened 2020-05-28
Find Method & Function calls

### LINKED ISSUE #154: Method/Function - find function calls
Inspired by: https://twitter.com/Ocramius/status/682249467148529666

Given a method / function / closure, e.g.: 

``` php
function myThing()
{
    $i = foo();
    return $i->whatever();
}
```

With a Method/Function reflection on this, it would be neat to be able to analyse the body AST, like:

``` php
$functionInfo = ReflectionFunction::createFromName('myThing');
$functionInfo->containsFunctionCall('foo'); // returns true
$functionInfo->getFunctionCalls('foo'); // returns array of AST nodes, or maybe a new VO wrapping info
$functionInfo->containsMethodCall('i', 'whatever'); // true
$functionInfo->getMethodCalls('i', 'whatever'); // array of whatever
```

DanielBadura:
Trying to get my first easypick issue done :stuck_out_tongue_closed_eyes:  
Should fix #154 

Not finished atm, maybe someone could tell me if im going the right direction? 😄 
And is this still a desired feature?

### DIFF
--- a/src/Reflection/ReflectionFunction.php
+++ b/src/Reflection/ReflectionFunction.php
@@ -8,6 +8,7 @@ use Closure;
 use PhpParser\Node;
 use PhpParser\Node\FunctionLike as FunctionNode;
 use PhpParser\Node\Stmt\Namespace_ as NamespaceNode;
+use PhpParser\NodeTraverser;
 use Roave\BetterReflection\BetterReflection;
 use Roave\BetterReflection\Reflection\Adapter\Exception\NotImplemented;
 use Roave\BetterReflection\Reflection\Exception\FunctionDoesNotExist;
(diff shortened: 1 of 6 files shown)


ondrejmirtes:
This really isn’t possible without type inference. Static analysis engines like PHPStan and Psalm and much more suitable to answer queries like this. Because here you can’t recognize what type a variable is, so you won’t be able to find all method calls on a given class...

Ocramius:
Indeed: I appreciate the effort, but I think we will never get to detect things like this with a reflector:

```php
/**
 * @param 'foo'|'bar' $method
 * @param T1|T2 $object
 */
function call($object, $method) : void 
{
    $object->$method();
}
```

We'd need some sort of PHI representation of the code *after* type inference for this to work, and we're very far from that :)

I think @dantleech has worked on this sort of stuff though: maybe more insight from there?

DanielBadura:
No problem! I thought something like this would popup thats why i pushed the early draft of this PR :)

dantleech:
So you could happily detect static method and function calls, but anything dynamic would, as said, require type inference - and it's probably not worth the effort of partially supporting that, and fully supporting it would be an epic and ongoing investment (would be very cool though).

In Phpactor method references are found by traversing the AST (from the MS TolerantParser), finding method|function|constant accesses and then calling the [WorseReflection](https://github.com/phpactor/worse-reflection) type resolver (:grimacing:) (which in turn uses a static reflection API like BR to resolve structural info) to analyse the frame containing the access and provide the type info for variables as best as it can.

DanielBadura:
So i think just to cover static method calls would not be sufficient as a feature.. So i would suggest to just close this PR and update the underlying issue?

Ocramius:
Sorry for the delay: indeed, closing this and the original issue is the safest approach (for now).

More advanced code analysis may become interesting in future, but indeed we need PHI notation before analyzing code usage.

PR #629, opened 2020-05-28
Fixed ReflectionClass::getReflectionConstant() in adapter for non-existent constant

### LINKED ISSUE #625: Adapter ReflectionClass::getReflectionConstant returns null but should return false
See PoC fix: https://github.com/ondrejmirtes/BetterReflection/commit/07af30d92461a6a4e34194d21ab188b32e9cfcaa

kukulich:
Fixes #625

### DIFF
--- a/src/Reflection/Adapter/ReflectionClass.php
+++ b/src/Reflection/Adapter/ReflectionClass.php
@@ -249,9 +249,12 @@ class ReflectionClass extends CoreReflectionClass
      */
     public function getReflectionConstant($name)
     {
-        return new ReflectionClassConstant(
-            $this->betterReflectionClass->getReflectionConstant($name),
-        );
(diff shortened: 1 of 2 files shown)


src/Reflection/Adapter/ReflectionClass.php
```diff
@@ -249,9 +249,12 @@ public function getConstant($name)
      */
     public function getReflectionConstant($name)
     {
-        return new ReflectionClassConstant(
-            $this->betterReflectionClass->getReflectionConstant($name),
-        );
+        $betterReflectionConstant = $this->betterReflectionClass->getReflectionConstant($name);
+        if ($betterReflectionConstant === null) {
+            return false;
```
Ocramius, line 254: PHP has such a consistent library! 🤣

PR #632, opened 2020-05-31
FileHelper - fix normalizing paths with protocol:// on Windows

ondrejmirtes:
Previously it generated path with `phar:\\...`.

### DIFF
--- a/src/Util/FileHelper.php
+++ b/src/Util/FileHelper.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Roave\BetterReflection\Util;
 
+use function preg_match;
 use function str_replace;
 use const DIRECTORY_SEPARATOR;
 
@@ -14,12 +15,19 @@ class FileHelper
         return str_replace('\\', '/', $path);
     }
 
-    public static function normalizeSystemPath(string $path) : string
+    public static function normalizeSystemPath(string $originalPath) : string
     {
-        $path = self::normalizeWindowsPath($path);
+        $path = self::normalizeWindowsPath($originalPath);
+        preg_match('~^([a-z]+)\\:\\/\\/(.+)~', $path, $matches);
+        $scheme = null;
+        if ($matches !== []) {
+            [, $scheme, $path] = $matches;
+        }
 
-        return DIRECTORY_SEPARATOR === '\\'
-            ? str_replace('/', '\\', $path)
-            : $path;
+        return ($scheme !== null ? $scheme . '://' : '') . (
+            DIRECTORY_SEPARATOR === '\\'
+                ? str_replace('/', '\\', $path)
+                : $path
+        );
     }
 }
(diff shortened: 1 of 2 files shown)


src/Util/FileHelper.php
```diff
@@ -14,12 +15,21 @@ public static function normalizeWindowsPath(string $path) : string
...
-    public static function normalizeSystemPath(string $path) : string
+    public static function normalizeSystemPath(string $originalPath, string $directorySeparator = DIRECTORY_SEPARATOR) : string
     {
-        $path = self::normalizeWindowsPath($path);
+        $originalPath = self::normalizeWindowsPath($originalPath);
+        preg_match('~^([a-z]+)\\:\\/\\/(.+)~', $originalPath, $matches);
+        if ($matches !== []) {
+            [, $scheme, $path] = $matches;
+        } else {
+            $scheme = null;
```
Ocramius, line 25: these should be default values (no `else` needed)

src/Util/FileHelper.php
```diff
@@ -14,12 +15,21 @@ public static function normalizeWindowsPath(string $path) : string
...
+            $path   = $originalPath;
+        }
 
-        return DIRECTORY_SEPARATOR === '\\'
-            ? str_replace('/', '\\', $path)
-            : $path;
+        return ($scheme !== null ? $scheme . '://' : '') . (
+            $directorySeparator === '\\'
```
Ocramius, line 30: I don't think `$directorySeparator` should leak into the API of this helper: this shifts responsibility to the caller, and that's only more trouble.

ondrejmirtes: @Ocramius I did this so that the code is easily testable, no matter the platform. Any suggestion what to do instead?

Ocramius: Hardcoding: we already have CI running against both linux and windows 👍

ondrejmirtes: Alright, pushed some changes

test/unit/Util/FileHelperTest.php
```diff
@@ -26,4 +26,17 @@ public function testSystemWindowsPath() : void
...
+
+    public function testSystemWindowsPathWithProtocol() : void
+    {
+        if (DIRECTORY_SEPARATOR !== '\\') {
+            $this->markTestSkipped('Test runs only on Windows');
+        }
+
+        $path = 'phar://C:/Users/ondrej/phpstan.phar/src/TrinaryLogic.php';
+        self::assertSame(
+            'phar://C:\Users\ondrej\phpstan.phar\src\TrinaryLogic.php',
```
Ocramius, line 38: Interesting that phars still have borked file paths :shrug:

ondrejmirtes: I don't get it - `'phar://C:\Users\ondrej\phpstan.phar\src\TrinaryLogic.php'` is what we want here.

Ocramius: Yes, but effectively this is no longer a filesystem path.

Anyway, all good with this patch.

ondrejmirtes: Thanks :)

PR #634, opened 2020-06-01
Fixed some bugs reported by Infection

kukulich:


test/unit/Reflection/ReflectionParameterTest.php
```diff
@@ -343,13 +343,13 @@ public function testSetType() : void
     {
         $classInfo     = $this->reflector->reflect(Php7ParameterTypeDeclarations::class);
         $methodInfo    = $classInfo->getMethod('foo');
-        $parameterInfo = $methodInfo->getParameter('intParam');
+        $parameterInfo = $methodInfo->getParameter('forSetType');
 
         $parameterInfo->setType('string');
```
Ocramius, line 348: These are solved by the memoizing parser being immutable. alternatively, we need to run them in separate processes.

By using separate parameters/methods/etc we are just moving the mine to some other location, where somebody else will trip it.

kukulich: Makes sense. Added `@runInSeparateProcess`

PR #637, opened 2020-06-01
`ReflectionSourceStubber` should generate types of properties

kukulich:


Ocramius:
Nice improvement! Wasn't expecting stubs to already have types for properties :)

PR #641, opened 2020-06-08
Fix name of CalculateReflectionColumn class

phil-davis:
There is already `CalculateReflectionColumnTest` (spelled correctly) that should provide test coverage for this typo fix of the class name.


asgrim:
@Ocramius BC break here, are we merging for next major?

kukulich:
@asgrim The class is marked as `@internal`

asgrim:
@kukulich so it is, that'll teach me for reviewing too quickly on my phone 🤣

PR #652, opened 2020-06-09
Corrected trait analysis: method ordering, abstract trait methods do not override concrete ones, method aliases now recognized

### LINKED ISSUE #647: ReflectionClass - abstract trait method should not override concrete method from parent class
(Yeah, I want to kill myself because of PHP and traits too.)

Consider this code (https://3v4l.org/GNnML):

```php
abstract class AbstractFoo
{

	public function myFoo(): void
	{
            var_dump('foo');
	}
	
	public function myBar(): void
	{
	    var_dump('bar from class');
	}

}

trait TraitFoo
{

	abstract public function myFoo(): void;
	
	public function myBar(): void
	{
	    var_dump('bar from trait');
	}

}

final class ClassFoo extends AbstractFoo
{

	use TraitFoo;

}

$foo = new ClassFoo();
$foo->myFoo(); // prints "foo"
$foo->myBar(); // prints "bar from trait"
```

PHP reflection matches that:

```php
$reflectionClass = new \ReflectionClass(\ClassFoo::class);
var_dump($reflectionClass->getMethod('myFoo')->getDeclaringClass()->getName()); // AbstractFoo
var_dump($reflectionClass->getMethod('myBar')->getDeclaringClass()->getName()); // ClassFoo
```

Unfortunately BetterReflection thinks that the implementing class of `myFoo` (declaring class from the Adapter point of view) is ClassFoo and that the method is abstract and therefore PHPStan considered it not implemented.

PoC fix: https://github.com/ondrejmirtes/BetterReflection/commit/aac16f0448261958468f0b3be803cb86ddb6d7a3

Original issue: https://github.com/phpstan/phpstan/issues/3406

### LINKED ISSUE #659: Trait method aliases don't work in a simple scenario without insteadof
See this example:

```php
trait Foo {
	public function bar(): void {

	}
}

class SomeClass {
	use Foo {
		bar as baz;
	}
}
```

BetterReflection reports that `SomeClass::baz()` does not exist.

Additionally, `SomeClass::bar()` also exists and it doesn't seem like a PHP bug. The manual says (https://www.php.net/manual/en/language.oop5.traits.php):

> The Aliased_Talker makes use of the as operator to be able to use B's bigTalk implementation under an additional alias talk. 

Runtime proof: https://3v4l.org/kBjVh
Reproduction: https://gist.github.com/ondrejmirtes/a18fc61b29d2319beb792ec5252b1939

_Silly me 2 days ago thought that https://github.com/Roave/BetterReflection/issues/647 was bad enough 🤦_

kukulich:
Fixes #647 
Fixes #659

Also fixes methods order: https://3v4l.org/LiTi4

### DIFF
--- a/src/Reflection/ReflectionClass.php
+++ b/src/Reflection/ReflectionClass.php
@@ -218,15 +218,9 @@ class ReflectionClass implements Reflection
         $methodAst = $method->getAst();
         assert($methodAst instanceof ClassMethod);
 
-        $methodHash = $this->methodHash($method->getDeclaringClass()->getName(), $method->getName());
-
-        $aliases = [];
-        foreach ($traitAliases as $aliasMethodName => $traitAliasDefinition) {
-            if ($methodHash !== $traitAliasDefinition) {
-                continue;
-            }
-
-            $aliases[] = ReflectionMethod::createFromNode(
+        $methodHash   = $this->methodHash($method->getImplementingClass()->getName(), $method->getName());
+        $createMethod = function (?string $aliasMethodName) use ($method, $methodAst) : ReflectionMethod {
+            return ReflectionMethod::createFromNode(
                 $this->reflector,
                 $methodAst,
                 $method->getDeclaringClass()->getDeclaringNamespaceAst(),
@@ -234,26 +228,22 @@ class ReflectionClass implements Reflection
                 $this,
                 $aliasMethodName,
             );
+        };
+
+        $methods = [];
+        foreach ($traitAliases as $aliasMethodName => $traitAliasDefinition) {
+            if ($methodHash !== $traitAliasDefinition) {
+                continue;
+            }
+
+            $methods[] = $createMethod($aliasMethodName);
         }
 
-        if ($aliases !== []) {
-            return $aliases;
+        if (! array_key_exists($methodHash, $traitPrecedences)) {
+            $methods[] = $createMethod($method->getAliasName());
         }
 
-        if (array_key_exists($methodHash, $traitPrecedences)) {
-            return [];
-        }
-
-        $newMethod = ReflectionMethod::createFromNode(
-            $this->reflector,
-            $methodAst,
-            $method->getDeclaringClass()->getDeclaringNamespaceAst(),
-            $method->getDeclaringClass(),
-            $this,
-            $method->getAliasName(),
-        );
-
-        return [$newMethod];
+        return $methods;
     }
 
     /**
--- a/src/Reflection/ReflectionClass.php
+++ b/src/Reflection/ReflectionClass.php
@@ -206,54 +206,6 @@ class ReflectionClass implements Reflection
     }
 
     /**
-     * Construct a flat list of all methods from current class, traits,
-     * parent classes and interfaces in this precise order.
-     *
-     * @return ReflectionMethod[]
-     */
-    private function getAllMethods() : array
-    {
-        return array_merge(
-            [],
-            array_map(
-                function (ClassMethod $methodNode) : ReflectionMethod {
-                    return ReflectionMethod::createFromNode(
-                        $this->reflector,
-                        $methodNode,
-                        $this->declaringNamespace,
-                        $this,
-                        $this,
-                    );
-                },
-                $this->node->getMethods(),
-            ),
-            ...array_map(
-                function (ReflectionClass $trait) : array {
-                    return array_merge(
-                        [],
-                        ...array_map(
-                            function (ReflectionMethod $method) : array {
-                                return $this->createMethodsFromTrait($method);
-                            },
-                            $trait->getMethods(),
-                        ),
-                    );
-                },
-                $this->getTraits(),
-            ),
-            ...array_map(
-                static function (ReflectionClass $ancestor) : array {
-                    return $ancestor->getMethods();
-                },
-                array_values(array_merge(
-                    array_filter([$this->getParentClass()]),
-                    $this->getInterfaces(),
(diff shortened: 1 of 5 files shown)


test/unit/Fixture/MethodsOrder.php
```diff
@@ -9,7 +9,7 @@ public function forth();
 
 class MethodsOrderParent
 {
-    public function third()
+    public function second()
```
Ocramius, line 12: We should probably have a new fixture here, instead of changing the existing one (risky)

kukulich: There's new fixture but this test is broken in master so this is fix - https://3v4l.org/n5MhC

Ocramius: That didn't parse for me, sorry: you mean this test didn't run on `master`? Or is the change introducing a breakage?

kukulich: The expectation in master is wrong.

Ocramius:
@kukulich I see this is a draft: is it still WIP on your end?

ondrejmirtes:
We discussed this privately with @kukulich - there's no assert about the `bar()` method and I think it will prove the current fix is wrong for that case.

ondrejmirtes:
I confirm this now fixes both #647 and #659.

kukulich:
@Ocramius Please just merge it. I don't want to see the code again for some time... :)

test/unit/Reflection/ReflectionClassTest.php
```diff
@@ -977,14 +1041,14 @@ public function testMethodsFromTraits() : void
 
         $classInfo = $reflector->reflect('TraitFixtureC');
 
-        self::assertFalse($classInfo->hasMethod('a'));
+        self::assertTrue($classInfo->hasMethod('a'));
```
kukulich, line 1044: Yes, it was bug.

test/unit/Reflection/ReflectionClassTest.php
```diff
@@ -977,14 +1041,14 @@ public function testMethodsFromTraits() : void
...
         self::assertTrue($classInfo->hasMethod('a_protected'));
 
         $aProtected = $classInfo->getMethod('a_protected');
 
         self::assertSame('TraitFixtureTraitC', $aProtected->getDeclaringClass()->getName());
 
-        self::assertFalse($classInfo->hasMethod('b'));
+        self::assertTrue($classInfo->hasMethod('b'));
```
kukulich, line 1051: Yes, it was bug.

test/unit/Reflection/ReflectionClassTest.php
```diff
@@ -1019,7 +1089,7 @@ public function testMethodsFromTraitsWithConflicts() : void
         self::assertSame('TraitFixtureTraitD1', $foo->getDeclaringClass()->getName());
         self::assertSame('TraitFixtureD', $foo->getImplementingClass()->getName());
 
-        self::assertFalse($classInfo->hasMethod('hoo'));
+        self::assertTrue($classInfo->hasMethod('hoo'));
```
kukulich, line 1092: Yes, it was bug.

kukulich:
@Ocramius This is ready to review too :)

src/Reflection/ReflectionClass.php
```diff
@@ -801,15 +838,19 @@ public function getParentClass() : ?ReflectionClass
             return null;
         }
 
-        $parent = $this->reflector->reflect($this->node->extends->toString());
-        // @TODO use actual `ClassReflector` or `FunctionReflector`?
-        assert($parent instanceof self);
+        if ($this->cachedParentClass === null) {
+            $parent = $this->reflector->reflect($this->node->extends->toString());
+            // @TODO use actual `ClassReflector` or `FunctionReflector`?
```
Ocramius, line 843: I think this TODO is solved for now: we can improve on the type system of reflectors later on.

src/Reflection/ReflectionClass.php
```diff
@@ -241,70 +283,33 @@ function (ReflectionMethod $method) : array {
...
-
-        $newMethod = ReflectionMethod::createFromNode(
-            $this->reflector,
-            $methodAst,
-            $method->getDeclaringClass()->getDeclaringNamespaceAst(),
-            $method->getDeclaringClass(),
-            $this,
-            $method->getAliasName(),
+        return array_merge(
+            [],
```
Ocramius, line 295: BTW, since we are on PHP 7.4+, `array_merge([], ...$others)` is no longer needed: https://3v4l.org/HUtnH

src/Reflection/ReflectionClass.php
```diff
@@ -801,15 +838,19 @@ public function getParentClass() : ?ReflectionClass
             return null;
         }
 
-        $parent = $this->reflector->reflect($this->node->extends->toString());
-        // @TODO use actual `ClassReflector` or `FunctionReflector`?
-        assert($parent instanceof self);
+        if ($this->cachedParentClass === null) {
```
Ocramius, line 841: Caching this is a nice addition :+1:

src/Reflection/ReflectionClass.php
Ocramius: Not relevant for this patch: can be removed later

src/Reflection/ReflectionClass.php
Ocramius: Not relevant for this patch: maybe interesting for slevomat/coding-standard :D

PR #653, opened 2020-06-09
Getting the last doc comment instead of the first one matches PHP reflection

### LINKED ISSUE #648: Getting last doc comment instead of first one would match PHP reflection
See PHP behaviour: https://3v4l.org/RGiqv

PoC fix: https://github.com/ondrejmirtes/BetterReflection/commit/ab41f90af9119abb2ea9223cd22bf913c777de16

Original issue: https://github.com/phpstan/phpstan/issues/3402

kukulich:
Fixes #648 

There should be no BC break because the class is marked as `@internal`.

@ondrejmirtes is too lazy to send the PR :-D


src/Reflection/ReflectionProperty.php
```diff
@@ -233,7 +233,7 @@ public function getImplementingClass() : ReflectionClass
 
     public function getDocComment() : string
     {
-        return GetFirstDocComment::forNode($this->node);
+        return GetLastDocComment::forNode($this->node);
```
Ocramius, line 236: The modifications on the various `Reflection*` API endpoints need test additions here

kukulich: Added tests.

Ocramius:
Thanks @kukulich!

Issue #603, opened 2020-05-27
PhpStormStubsSourceStubber can support filtering identifiers based on given PHP version


kukulich:
Based on `@since` annotations

Eg. https://github.com/JetBrains/phpstorm-stubs/blob/master/FFI/FFI.php#L17

ondrejmirtes:
A few cases are tricky to handle:

* `array_push` has `@since 7.3` but it's also in older versions.
* `setcookie` is present multiple times which causes it to be overriden in a single-dimension array.
* `CompileError` is available since PHP 7.3, but in the stubs there's `ParseError extends CompileError` which is available since PHP 7.0. I lie about this in my fork and make CompileError available since 7.0 too.

kukulich:
Implemented in https://github.com/Roave/BetterReflection/pull/798

PR #660, opened 2020-06-11
Allow functions to be deprecated

### LINKED ISSUE #38: Create ReflectionFunctions/ReflectionClasses from internal definitions

BackEndTea:
Related:

* fixes #38 
* https://github.com/phpstan/phpstan-deprecation-rules/issues/24

### DIFF
--- a/src/Reflection/ReflectionFunctionAbstract.php
+++ b/src/Reflection/ReflectionFunctionAbstract.php
@@ -6,6 +6,7 @@ namespace Roave\BetterReflection\Reflection;
 
 use Closure;
 use LogicException;
+use phpDocumentor\Reflection\DocBlockFactory;
 use phpDocumentor\Reflection\Type;
 use PhpParser\Comment\Doc;
 use PhpParser\Node;
@@ -262,7 +263,16 @@ abstract class ReflectionFunctionAbstract
      */
     public function isDeprecated(): bool
     {
-        return strpos($this->getDocComment(), '@deprecated') !== false;
+        $docComment = $this->getDocComment();
+
+        if ($docComment === '') {
+            return false;
+        }
+
+        $docBlockFactory = DocBlockFactory::createInstance();
+        $docBlock = $docBlockFactory->create($docComment);
+
+        return $docBlock->hasTag('deprecated');
     }
 
     public function isInternal(): bool
--- a/src/Reflection/ReflectionFunctionAbstract.php
+++ b/src/Reflection/ReflectionFunctionAbstract.php
@@ -40,7 +40,6 @@ use function count;
 use function implode;
 use function is_array;
 use function is_string;
-use function strpos;
 use function strtolower;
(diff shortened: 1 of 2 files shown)


ondrejmirtes:
This should probably be done as an enhancement of https://github.com/Roave/BetterReflection/issues/603 to be really good. (We don't want to tell people running on PHP 7.1 about a deprecation on 7.4.)

src/Reflection/ReflectionFunctionAbstract.php
```diff
@@ -257,16 +258,10 @@ public function isClosure() : bool
...
-     * Note - we cannot reflect on internal functions (as there is no PHP source
-     * code we can access. This means, at present, we can only EVER return false
-     * from this function.
-     *
-     * @see https://github.com/Roave/BetterReflection/issues/38
      */
     public function isDeprecated() : bool
     {
-        return false;
+        return strpos($this->getDocComment(), '@deprecated') !== false;
```
Ocramius, line 264: Probably need to verify that `@deprecated` is followed by whitespace here

src/Reflection/ReflectionFunctionAbstract.php
```diff
@@ -257,16 +258,10 @@ public function isClosure() : bool
 
     /**
      * Is this function deprecated?
```
Ocramius, line 260: We can drop this entire docblock

src/Reflection/ReflectionFunctionAbstract.php
Ocramius: Test case: `@deprecatedPolicy` would for example still trip this

jaapio: Since better reflection is already using phpDocumentor/ReflectionDocblock, why not use that to find the deprecation tag?

Ocramius: Would most certainly help 👍

BackEndTea: I moved to the php documentor check. 
Ive kept the logic within the method. I could also extract the logic for checking if a tag exists to another class, but i'm not sure if it is needed. What do you think @Ocramius ?

Ocramius: LGTM 👍

Just needs a  rebase now 😅

Ocramius:
Thanks @BackEndTea!

BackEndTea:
Thanks @Ocramius, @ondrejmirtes & @jaapio

PR #662, opened 2020-06-14
Fix trait methods precedence with more than one-level parent hierarchy

ondrejmirtes:
Original issue: https://github.com/phpstan/phpstan/issues/3469

### DIFF
--- a/test/unit/Fixture/TraitWithAbstractMethod.php
+++ b/test/unit/Fixture/TraitWithAbstractMethod.php
@@ -29,7 +29,12 @@ trait TraitWithBoo
     }
 }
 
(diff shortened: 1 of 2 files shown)


test/unit/Fixture/TraitWithAbstractMethod.php
```diff
@@ -29,7 +29,12 @@ public function boo(): void
     }
 }
 
-class ClassUsingTraitWithAbstractMethod extends AbstractClassImplementingMethodFromTrait
+abstract class AbstractClassCausingIndirectParent extends AbstractClassImplementingMethodFromTrait
```
Ocramius, line 32: I think we should have a new explicit scenario for this

Issue #673, opened 2020-07-01
Adapter\ReflectionClass::$name is empty


ondrejmirtes:
It should contain the class name instead.

Not sure if this has a solution, tried something but PHP doesn't allow it: https://3v4l.org/dmV2m

ondrejmirtes:
Reflection doesn't help here: https://3v4l.org/9cOop

Ocramius:
Fear not, for I have seen the gates of hell, and I hath returned! https://3v4l.org/PBRN1

ondrejmirtes:
Awesome, I need to take a shower now 😝  😃

powtac:
$name somehow protected?

ondrejmirtes:
Looking at https://github.com/JetBrains/phpstorm-stubs/blob/master/Reflection/Reflection.php there are 12 such properties.

I sent a fix for the first one: https://github.com/Roave/BetterReflection/pull/674 If the review passes I can send fixes for the rest too.

Ocramius:
Handled in #674

PR #674, opened 2020-07-02
Fix Adapter\ReflectionClass::$name

ondrejmirtes:
Part of #673.

### DIFF
--- a/src/Reflection/Adapter/ReflectionClass.php
+++ b/src/Reflection/Adapter/ReflectionClass.php
@@ -33,6 +33,8 @@ class ReflectionClass extends CoreReflectionClass
     public function __construct(BetterReflectionClass $betterReflectionClass)
     {
         $this->betterReflectionClass = $betterReflectionClass;
+
+        unset($this->name);
     }
 
     /**
@@ -67,6 +69,22 @@ class ReflectionClass extends CoreReflectionClass
     public function __toString()
     {
         return $this->betterReflectionClass->__toString();
+    }
+
+    /**
+     * @param string $name
+     *
+     * @return mixed
+     *
+     * @phpcsSuppress SlevomatCodingStandard.TypeHints.ParameterTypeHint.MissingNativeTypeHint
+     */
+    public function __get($name)
+    {
+        if ($name === 'name') {
+            return $this->betterReflectionClass->getName();
+        }
+
+        throw new OutOfBoundsException(sprintf('Property %s::$%s does not exist.', self::class, $name));
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/Reflection/Adapter/ReflectionClass.php
```diff
@@ -68,6 +70,22 @@ public function __toString()
...
+     *
+     * @phpcsSuppress SlevomatCodingStandard.TypeHints.ParameterTypeHint.MissingNativeTypeHint
+     */
+    public function __get($name)
+    {
+        if ($name === 'name') {
+            return $this->betterReflectionClass->getName();
+        }
+
+        return $this->{$name};
```
Ocramius, line 86: I think we need to prevent this - makes properties public-ish by default

ondrejmirtes: I haven't realized this. As there are no other public properties we probably don't need to care about that? But I'm not sure how to trigger the same error PHP itself does when stumbling upon an undefined property...

Ocramius: `$this->betterReflectionClass`

ondrejmirtes: Just updated it.

src/Reflection/Adapter/ReflectionClass.php
```diff
@@ -68,6 +70,22 @@ public function __toString()
...
+     *
+     * @phpcsSuppress SlevomatCodingStandard.TypeHints.ParameterTypeHint.MissingNativeTypeHint
+     */
+    public function __get($name)
+    {
+        if ($name === 'name') {
+            return $this->betterReflectionClass->getName();
+        }
+
+        return $this->betterReflectionClass->{$name};
```
Ocramius, line 86: Can we test this path, or maybe throw an exception?

Ocramius: Yep, throwing an exception seems more sensible here: we just need a test case for it :)

ondrejmirtes: I'd be more comfortable with 1:1 behaviour with PHP's reflection class :)

Ocramius: In which case you'd need a nightmare like https://github.com/Ocramius/ProxyManager/blob/master/src/ProxyManager/ProxyGenerator/Util/PublicScopeSimulator.php :P 

Ok, we just need a test raising an error then (and asserting on it)

ondrejmirtes: Oh god. Alright, what should be thrown?

Ocramius: an `OutOfBoundsException`, probably

ondrejmirtes: Updated.

PR #678, opened 2020-07-07
ReflectionParameter: try to get the correct default value for constants

voku:
For example, this function ```function foo(int $case = \CASE_LOWER);```, ReflectionParameter->allowsNull() for ```$case```will currently return ```true```.

---
This change is [Reviewable](https://reviewable.io/reviews/roave/betterreflection/678)

### DIFF
--- a/src/Reflection/ReflectionParameter.php
+++ b/src/Reflection/ReflectionParameter.php
@@ -11,11 +11,13 @@ use LogicException;
 use OutOfBoundsException;
 use phpDocumentor\Reflection\Type;
 use PhpParser\Node;
+use PhpParser\Node\Expr;
 use PhpParser\Node\NullableType;
 use PhpParser\Node\Param as ParamNode;
 use PhpParser\Node\Stmt\Namespace_;
 use Roave\BetterReflection\NodeCompiler\CompileNodeToValue;
 use Roave\BetterReflection\NodeCompiler\CompilerContext;
+use Roave\BetterReflection\NodeCompiler\Exception\UnableToCompileNode;
 use Roave\BetterReflection\Reflection\Exception\Uncloneable;
 use Roave\BetterReflection\Reflection\StringCast\ReflectionParameterStringCast;
 use Roave\BetterReflection\Reflector\ClassReflector;
@@ -158,7 +160,10 @@ class ReflectionParameter
         return $param;
     }
 
-    private function parseDefaultValueNode(): void
+    /**
+     * @throws LogicException
+     */
+    private function parseDefaultValueNode(): ?Expr
     {
         if (! $this->isDefaultValueAvailable()) {
             throw new LogicException('This parameter does not have a default value available');
@@ -168,34 +173,26 @@ class ReflectionParameter
 
         if ($defaultValueNode instanceof Node\Expr\ClassConstFetch) {
             assert($defaultValueNode->class instanceof Node\Name);
+            assert($defaultValueNode->name instanceof Node\Identifier);
+
             $className = $defaultValueNode->class->toString();
 
             if ($className === 'self' || $className === 'static') {
-                assert($defaultValueNode->name instanceof Node\Identifier);
                 $constantName = $defaultValueNode->name->name;
                 $className    = $this->findParentClassDeclaringConstant($constantName);
             }
 
-            $this->isDefaultValueConstant = true;
-            assert($defaultValueNode->name instanceof Node\Identifier);
+            $this->isDefaultValueConstant   = true;
             $this->defaultValueConstantName = $className . '::' . $defaultValueNode->name->name;
-        }
-
-        if (
+        } elseif (
             $defaultValueNode instanceof Node\Expr\ConstFetch
             && ! in_array(strtolower($defaultValueNode->name->parts[0]), ['true', 'false', 'null'], true)
         ) {
             $this->isDefaultValueConstant   = true;
             $this->defaultValueConstantName = $defaultValueNode->name->parts[0];
-            $this->defaultValue             = null;
-
-            return;
         }
 
-        $this->defaultValue = (new CompileNodeToValue())->__invoke(
-            $defaultValueNode,
-            new CompilerContext($this->reflector, $this->getDeclaringClass()),
-        );
+        return $defaultValueNode;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/Reflection/ReflectionParameter.php
```diff
@@ -187,7 +188,7 @@ private function parseDefaultValueNode(): void
         ) {
             $this->isDefaultValueConstant   = true;
             $this->defaultValueConstantName = $defaultValueNode->name->parts[0];
-            $this->defaultValue             = null;
+            $this->defaultValue             = constant($this->defaultValueConstantName);
```
Ocramius, line 191: This is not really acceptable, since `constant()` leads to autoloading here:

https://3v4l.org/BeYeT

voku: @Ocramius ```Node\Expr\ConstFetch``` is only for global constants not for class constants, instead of ```Node\Expr\ClassConstFetch```, or? If this is correct then there is no autoloading here.

Ocramius: Yeah, but then it will crash in case of an undefined constant: https://3v4l.org/ZZnmJ

Looking at our code, we do indeed use `constant()` elsewhere (in `Roave\BetterReflection\NodeCompiler\CompileNodeToValue`)

I'd say that this patch is good if following is true:

 1. tests are added
 2. `CompileNodeToValue` is used in here, instead of `constant()`

Having a `constant()` lookup there is already a bug: let's keep it isolated for now.

voku: @Ocramius ```constant()``` is replaced and a simple test case is added

But now I see ```Roave\BetterReflection\NodeCompiler\Exception\UnableToCompileNode : Could not locate constant "SOME_DEFINED_VALUE" ...``` in the tests because we are not requiring the class and so the ```\define('SOME_DEFINED_VALUE', 1);``` is not executed, and we can't get the value. Should we ignore these exceptions here or can we get the value of custom global constants somehow?

Ocramius: The test that fails is this one: 

```php
    public function testIsDefaultValueConstantAndGetDefaultValueConstantName() : void
    {
        $classInfo = $this->reflector->reflect(Methods::class);
        $method    = $classInfo->getMethod('methodWithUpperCasedDefaults');

        $boolUpper = $method->getParameter('boolUpper');
        self::assertFalse($boolUpper->isDefaultValueConstant());

        $boolLower = $method->getParameter('boolLower');
        self::assertFalse($boolLower->isDefaultValueConstant());

        $nullUpper = $method->getParameter('nullUpper');
        self::assertFalse($nullUpper->isDefaultValueConstant());

        $method       = $classInfo->getMethod('methodWithConstAsDefault');
        $constDefault = $method->getParameter('constDefault');
        self::assertTrue($constDefault->isDefaultValueConstant());
        self::assertSame(Methods::class . '::SOME_CONST', $constDefault->getDefaultValueConstantName());

        $definedDefault = $method->getParameter('definedDefault');
        self::assertTrue($definedDefault->isDefaultValueConstant());
        self::assertSame('SOME_DEFINED_VALUE', $definedDefault->getDefaultValueConstantName());

        $intDefault = $method->getParameter('intDefault');
        self::assertFalse($intDefault->isDefaultValueConstant());

        $this->expectException(LogicException::class);
        $this->expectExceptionMessage('This parameter is not a constant default value, so cannot have a constant name');
        $intDefault->getDefaultValueConstantName();
    }
```

The source is `Methods.php`:

```php
<?php

namespace Roave\BetterReflectionTest\Fixture;

\define('SOME_DEFINED_VALUE', 1);

abstract class Methods
{
    const SOME_CONST = 1;

    public function __construct()
    {
    }

    public function publicMethod()
    {
    }

    private function privateMethod()
    {
    }

    protected function protectedMethod()
    {
    }

    final public function finalPublicMethod()
    {
    }

    abstract public function abstractPublicMethod();

    public static function staticPublicMethod()
    {
    }

    function noVisibility()
    {
    }

    public function __destruct()
    {
    }

    /**
     * @param string $parameter1
     * @param int|float $parameter2
     */
    public function methodWithParameters($parameter1, $parameter2)
    {
    }

    public function methodWithOptionalParameters($parameter, $optionalParameter = null)
    {
    }

    public function methodWithExplicitTypedParameters(
        \stdClass $stdClassParameter,
        ClassForHinting $namespaceClassParameter,
        \Roave\BetterReflectionTest\Fixture\ClassForHinting $fullyQualifiedClassParameter,
        array $arrayParameter,
        callable $callableParameter
    ) {
    }

    public function methodWithVariadic($nonVariadicParameter, ...$variadicParameter)
    {
    }

    public function methodWithReference($nonRefParameter, &$refParameter)
    {
    }

    public function methodWithNonOptionalDefaultValue($firstParameter = 'someValue', $secondParameter)
    {
    }

    public function methodToCheckAllowsNull($allowsNull, \stdClass $hintDisallowNull, \stdClass $hintAllowNull = null)
    {
    }

    public function methodWithConstAsDefault($intDefault = 1, $constDefault = self::SOME_CONST, $definedDefault = SOME_DEFINED_VALUE)
    {
    }

    public function methodWithUpperCasedDefaults($boolUpper = TRUE, $boolLower = false, $nullUpper = NULL)
    {
    }
}
```

I wonder if we can make the default value "lazy", so that we only trip the exception if a read on the constant is happening. The test above is indeed correct: the `Methods.php` file should never be included, and therefore `SOME_DEFINED_VALUE` cannot be found by `constant()`.

src/Reflection/ReflectionParameter.php
```diff
@@ -187,7 +187,10 @@ private function parseDefaultValueNode(): void
         ) {
             $this->isDefaultValueConstant   = true;
             $this->defaultValueConstantName = $defaultValueNode->name->parts[0];
-            $this->defaultValue             = null;
+            $this->defaultValue             = (new CompileNodeToValue())->__invoke(
+                $defaultValueNode,
+                new CompilerContext($this->reflector, null),
```
Ocramius, line 192: I think `$this->getDeclaringClass()` should still be passed in here.

Still, assuming we get rid of `$this->defaultValue`, and instead store the AST node for the constant, we can use `CompileNodeToValue` in a getter, and defer the exception (for now)

voku: @Ocramius I moved ```CompileNodeToValue``` in the getter and added a ```@throw``` comment there

Ocramius: Nice!

src/Reflection/ReflectionParameter.php
```diff
@@ -449,7 +466,10 @@ public function isDefaultValueConstant(): bool
      */
     public function getDefaultValueConstantName(): string
     {
-        $this->parseDefaultValueNode();
+        if ($this->defaultValueAst === null) {
```
Ocramius, line 469: Are these AST caches necessary? I would suggest avoiding them, if they don't bring a massive advantage (the added complexity is noticeable)

voku: I remvoed the AST cache.

PR #700, opened 2020-10-24
Fix compatibility with phpdoc reflection when the docblocks could not be parsed by phpdocumentor

voku:
-> Call to undefined method phpDocumentor\Reflection\DocBlock\Tags\InvalidTag::getType()

ref: https://github.com/phpDocumentor/ReflectionDocBlock/pull/260

---
This change is [Reviewable](https://reviewable.io/reviews/roave/betterreflection/700)

### DIFF
--- a/src/TypesFinder/FindPropertyType.php
+++ b/src/TypesFinder/FindPropertyType.php
@@ -44,13 +44,19 @@ class FindPropertyType
         }
 
         $context = $this->makeContext->__invoke($namespace);
-        /** @var Var_[] $varTags */
-        $varTags = $this->docBlockFactory->create($docComment, $context)->getTagsByName('var');
+
+        $varTags = $this->docBlockFactory
+            ->create($docComment, $context)
+            ->getTagsByName('var');
 
         return array_merge(
             [],
-            ...array_map(function (Var_ $varTag) use ($context) {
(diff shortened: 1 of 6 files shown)


src/TypesFinder/FindPropertyType.php
```diff
@@ -44,8 +45,11 @@ public function __invoke(ReflectionProperty $reflectionProperty, ?Namespace_ $na
         }
 
         $context = $this->makeContext->__invoke($namespace);
-        /** @var Var_[] $varTags */
-        $varTags = $this->docBlockFactory->create($docComment, $context)->getTagsByName('var');
+
+        /** @var Var_[]|InvalidTag[] $varTags */
```
Ocramius: Hmm, van this type be inferred from upstream?

Ocramius (approved):
LGTM - wish phpdocumentor had better documented types here though.

Before we merge, can you check if removing the `@var` forced types leads to equivalent static analysis?

Ocramius (approved):
LGTM 👍

voku:
> LGTM - wish phpdocumentor had better documented types here though.
> 
> Before we merge, can you check if removing the `@var` forced types leads to equivalent static analysis?

`@var` usage is removed and it looks good + some more tests are added

Ocramius:
Thanks, @voku! It's released 👍

PR #705, opened 2020-11-01
bump psalm to 4.1

orklah:
Hi,

This PR propose upgrading Psalm to 4.1 and fixing new issues along the way.

Note1: I couldn't find a way to upgrade psalm without obliterating the composer.lock first so it has been regenerated completely, I did so with composer 2, I don't know if it could cause an issue.

Note2: There is a future BC in this commit because I changed a parameter name that didn't match with the parent. Technically, this package doesn't support PHP8 so it should be fine to do it now.

Note3: I had to set phpstan version to `0.12.25` instead of `^0.12.25` to avoid updating it. There are errors in `0.12.26` that are related to covariance/contravariance with internal interfaces that can't be fixed and can't be added to ignore section of phpstan. I don't know what can be done about that.

### DIFF
--- a/src/Reflection/ReflectionMethod.php
+++ b/src/Reflection/ReflectionMethod.php
@@ -17,6 +17,7 @@ use Roave\BetterReflection\Reflection\Exception\ObjectNotInstanceOfClass;
 use Roave\BetterReflection\Reflection\StringCast\ReflectionMethodStringCast;
 use Roave\BetterReflection\Reflector\Exception\IdentifierNotFound;
 use Roave\BetterReflection\Reflector\Reflector;
+use Webmozart\Assert\Assert;
 
 use function class_exists;
 use function get_class;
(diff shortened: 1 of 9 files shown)


src/Reflection/ReflectionMethod.php
```diff
@@ -352,9 +357,13 @@ private function callStaticMethod(array $args)
      */
     private function callObjectMethod(object $object, array $args)
     {
-        return Closure::bind(function ($object, string $methodName, array $methodArgs) {
+        $closure = Closure::bind(function ($object, string $methodName, array $methodArgs) {
```
Ocramius, line 360: Love me some shitty PHP core API :P

Ocramius (approved):
LGTM 👍

Ocramius:
Hmm, MT failed due to:

```
Error: ] The minimum required MSI percentage should be 86%, but actual is       
         85.859097820578%. Improve your tests!                                  
```

I think we perhaps need `roave/infection-static-analysis-plugin` for this stuff

orklah:
Yeah, seems like the bump of infection changed the result. The same can be seen here: https://github.com/Roave/BetterReflection/pull/698

Maybe we could merge this and I can look at roave/infection-static-analysis-plugin in another PR?

Ocramius:
I would prefer keeping the build green: all automations dependend on it

orklah:
It's done. roave/infection-static-analysis-plugin mechanically augmented the score so the build is green.

Ocramius (approved):
Excellent!

```
Metrics:
         Mutation Score Indicator (MSI): 88%
         Mutation Code Coverage: 93%
         Covered Code MSI: 94%
```

Thanks @orklah!

PR #711, opened 2020-12-11
Allow non standard `vendor-dir` values when constructing a source locator from a `composer.json`

tomasnorre:
This PR will allow non-standard vendor-dirs as the value is read from composer-json, if not set it will default to 'vendor'. 

This relates to question https://github.com/Roave/BackwardCompatibilityCheck/issues/271

I'm still trying to get my head around how to proper add a test for a non-standard vendor-dir, but still reading through the tests, to see how to add them in a proper way. 

So far I'm thinking of copying this (1) or extending the dataprovider.

https://github.com/Roave/BetterReflection/blob/df8a13c8aa/test/unit/SourceLocator/Type/Composer/Factory/MakeLocatorForComposerInstalledJsonTest.php#L43

The tests are still green, but the added feature isn't tested yet.

Feedback is appreciated.

### DIFF
--- a/src/SourceLocator/Type/Composer/Factory/MakeLocatorForComposerJsonAndInstalledJson.php
+++ b/src/SourceLocator/Type/Composer/Factory/MakeLocatorForComposerJsonAndInstalledJson.php
@@ -38,19 +38,27 @@ final class MakeLocatorForComposerJsonAndInstalledJson
             throw InvalidProjectDirectory::atPath($installationPath);
         }
 
-        $composerJsonPath  = $realInstallationPath . '/composer.json';
-        $installedJsonPath = $realInstallationPath . '/vendor/composer/installed.json';
+        $composerJsonPath = $realInstallationPath . '/composer.json';
 
         if (! file_exists($composerJsonPath)) {
             throw MissingComposerJson::inProjectPath($installationPath);
         }
 
+        /**
+         * @psalm-var array{
+         * autoload: array{classmap: array<int, string>, files: array<int, string>, psr-4: array<string, array<int, string>>, psr-0: array<string, array<int, string>>},
+         * config: array{vendor-dir?: string}
+         * }|null $composer
+         */
+        $composer  = json_decode((string) file_get_contents($composerJsonPath), true);
+        $vendorDir = $composer['config']['vendor-dir'] ?? 'vendor';
+
+        $installedJsonPath = $realInstallationPath . '/' . $vendorDir . '/composer/installed.json';
+
         if (! file_exists($installedJsonPath)) {
             throw MissingInstalledJson::inProjectPath($installationPath);
         }
 
-        /** @var array{autoload: array{classmap: array<int, string>, files: array<int, string>, psr-4: array<string, array<int, string>>, psr-0: array<string, array<int, string>>}}|null $composer */
-        $composer = json_decode((string) file_get_contents($composerJsonPath), true);
         /** @var array{packages: list<array>}|list<array>|null $installedJson */
         $installedJson = json_decode((string) file_get_contents($installedJsonPath), true);
 
@@ -67,10 +75,10 @@ final class MakeLocatorForComposerJsonAndInstalledJson
 
         $classMapPaths       = array_merge(
             $this->prefixPaths($this->packageToClassMapPaths($composer), $realInstallationPath . '/'),
-            ...array_map(function (array $package) use ($realInstallationPath): array {
+            ...array_map(function (array $package) use ($realInstallationPath, $vendorDir): array {
                 return $this->prefixPaths(
                     $this->packageToClassMapPaths($package),
-                    $this->packagePrefixPath($realInstallationPath, $package),
+                    $this->packagePrefixPath($realInstallationPath, $package, $vendorDir),
                 );
             }, $installed),
         );
@@ -78,10 +86,10 @@ final class MakeLocatorForComposerJsonAndInstalledJson
         $classMapDirectories = array_filter($classMapPaths, 'is_dir');
         $filePaths           = array_merge(
             $this->prefixPaths($this->packageToFilePaths($composer), $realInstallationPath . '/'),
-            ...array_map(function (array $package) use ($realInstallationPath): array {
+            ...array_map(function (array $package) use ($realInstallationPath, $vendorDir): array {
                 return $this->prefixPaths(
                     $this->packageToFilePaths($package),
-                    $this->packagePrefixPath($realInstallationPath, $package),
+                    $this->packagePrefixPath($realInstallationPath, $package, $vendorDir),
                 );
             }, $installed),
         );
@@ -91,11 +99,12 @@ final class MakeLocatorForComposerJsonAndInstalledJson
                 new PsrAutoloaderLocator(
                     Psr4Mapping::fromArrayMappings(array_merge_recursive(
                         $this->prefixWithInstallationPath($this->packageToPsr4AutoloadNamespaces($composer), $realInstallationPath),
-                        ...array_map(function (array $package) use ($realInstallationPath): array {
+                        ...array_map(function (array $package) use ($realInstallationPath, $vendorDir): array {
                             return $this->prefixWithPackagePath(
                                 $this->packageToPsr4AutoloadNamespaces($package),
                                 $realInstallationPath,
                                 $package,
+                                $vendorDir,
                             );
                         }, $installed),
                     )),
@@ -104,11 +113,12 @@ final class MakeLocatorForComposerJsonAndInstalledJson
                 new PsrAutoloaderLocator(
                     Psr0Mapping::fromArrayMappings(array_merge_recursive(
                         $this->prefixWithInstallationPath($this->packageToPsr0AutoloadNamespaces($composer), $realInstallationPath),
-                        ...array_map(function (array $package) use ($realInstallationPath): array {
+                        ...array_map(function (array $package) use ($realInstallationPath, $vendorDir): array {
                             return $this->prefixWithPackagePath(
                                 $this->packageToPsr0AutoloadNamespaces($package),
                                 $realInstallationPath,
                                 $package,
+                                $vendorDir,
                             );
                         }, $installed),
                     )),
@@ -169,9 +179,9 @@ final class MakeLocatorForComposerJsonAndInstalledJson
     /**
      * @param array{name: string} $package
      */
-    private function packagePrefixPath(string $trimmedInstallationPath, array $package): string
+    private function packagePrefixPath(string $trimmedInstallationPath, array $package, string $vendorDir): string
     {
-        return $trimmedInstallationPath . '/vendor/' . $package['name'] . '/';
+        return $trimmedInstallationPath . '/' . $vendorDir . '/' . $package['name'] . '/';
     }
 
     /**
@@ -180,9 +190,9 @@ final class MakeLocatorForComposerJsonAndInstalledJson
      *
      * @return array<string, array<int, string>>
      */
-    private function prefixWithPackagePath(array $paths, string $trimmedInstallationPath, array $package): array
+    private function prefixWithPackagePath(array $paths, string $trimmedInstallationPath, array $package, string $vendorDir): array
     {
-        $prefix = $this->packagePrefixPath($trimmedInstallationPath, $package);
+        $prefix = $this->packagePrefixPath($trimmedInstallationPath, $package, $vendorDir);
 
         return array_map(function (array $paths) use ($prefix): array {
             return $this->prefixPaths($paths, $prefix);
--- a/test/unit/Assets/ComposerLocators/project-with-custom-vendor-dir-using-composer-v2/composer.json
+++ b/test/unit/Assets/ComposerLocators/project-with-custom-vendor-dir-using-composer-v2/composer.json
@@ -1,0 +1,5 @@
+{
+  "config": {
+    "vendor-dir": "custom-vendor"
+  }
+}
--- a/test/unit/Assets/ComposerLocators/project-with-custom-vendor-dir-using-composer-v2/custom-vendor/a/b/src/ab_PSR-4_Sources/.gitignore
+++ b/test/unit/Assets/ComposerLocators/project-with-custom-vendor-dir-using-composer-v2/custom-vendor/a/b/src/ab_PSR-4_Sources/.gitignore
@@ -1,0 +1,2 @@
+*
+!.gitignore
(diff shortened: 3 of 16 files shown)


src/SourceLocator/Type/Composer/Factory/MakeLocatorForComposerJsonAndInstalledJson.php
```diff
@@ -39,18 +39,21 @@ public function __invoke(string $installationPath, Locator $astLocator): SourceL
         }
 
         $composerJsonPath  = $realInstallationPath . '/composer.json';
-        $installedJsonPath = $realInstallationPath . '/vendor/composer/installed.json';
 
         if (! file_exists($composerJsonPath)) {
             throw MissingComposerJson::inProjectPath($installationPath);
         }
 
+        /** @var array{autoload: array{classmap: array<int, string>, files: array<int, string>, psr-4: array<string, array<int, string>>, psr-0: array<string, array<int, string>>}, config: array{vendor-dir: <string>}}|null $composer */
```
tomasnorre: Would it's be an idea to have the `json_decode()` for the composer mapped into a `ValueObject`? Instead of working with arrays?

src/SourceLocator/Type/Composer/Factory/MakeLocatorForComposerJsonAndInstalledJson.php
```diff
@@ -38,19 +38,22 @@ public function __invoke(string $installationPath, Locator $astLocator): SourceL
...
-        $installedJsonPath = $realInstallationPath . '/vendor/composer/installed.json';
+        $composerJsonPath = $realInstallationPath . '/composer.json';
 
         if (! file_exists($composerJsonPath)) {
             throw MissingComposerJson::inProjectPath($installationPath);
         }
 
+        /** @var array{autoload: array{classmap: array<int, string>, files: array<int, string>, psr-4: array<string, array<int, string>>, psr-0: array<string, array<int, string>>}, config: array{vendor-dir: array<string, array<int, string>>}}|null $composer */
+        $composer  = json_decode((string) file_get_contents($composerJsonPath), true);
+        $vendorDir = $composer['config']['vendor-dir'] ?? 'vendor';
```
asgrim, line 54: Does this throw any warnings if the keys don't exist, I can never remember sorry heh

src/SourceLocator/Type/Composer/Factory/MakeLocatorForComposerJsonAndInstalledJson.php
```diff
@@ -38,19 +38,22 @@ public function __invoke(string $installationPath, Locator $astLocator): SourceL
...
+
+        $installedJsonPath = $realInstallationPath . '/' . $vendorDir . '/composer/installed.json';
```
asgrim, line 56: We probably need test for these cases:

- `composer.json` with a custom `vendor-dir` specified
- `composer.json` with no `vendor-dir` specified (I'm assuming this is already tested, but I honestly don't remember without looking)

asgrim (changes requested):

src/SourceLocator/Type/Composer/Factory/MakeLocatorForComposerJsonAndInstalledJson.php
asgrim: I think at the moment a VO just for this is possibly overkill; if we want to start looking at other `composer.json` keys in `config` in the future, something I'd certainly advocate.

asgrim:
> I'm still trying to get my head around how to proper add a test for a non-standard vendor-dir, but still reading through the tests, to see how to add them in a proper way.

https://github.com/Roave/BetterReflection/blob/4.12.x/test/unit/SourceLocator/Type/Composer/Factory/MakeLocatorForComposerJsonAndInstalledJsonTest.php

So this test, basically has multiple test cases for different project "setups". The skeletons are in https://github.com/Roave/BetterReflection/tree/4.12.x/test/unit/Assets/ComposerLocators - you'd probably have to make a couple of new project structures, something like `project-with-custom-vendor-dir`,`project-with-custom-vendor-dir-composer-v2` and add these as test cases.

src/SourceLocator/Type/Composer/Factory/MakeLocatorForComposerJsonAndInstalledJson.php
asgrim: I checked, yes it is tested already (also, see https://github.com/Roave/BetterReflection/pull/711#issuecomment-743718259 for testing suggestions!)

src/SourceLocator/Type/Composer/Factory/MakeLocatorForComposerJsonAndInstalledJson.php
tomasnorre: To my knowledge no. As the test don't throw any warning and it isn't set in the test, this should be safe.

src/SourceLocator/Type/Composer/Factory/MakeLocatorForComposerJsonAndInstalledJson.php
tomasnorre: The `with no vendor-dir specified` test is already there. I'll add the `with custom vendor-dir` test case too.

tomasnorre:
I have added a test case the in princible works. 

But this function 
https://github.com/Roave/BetterReflection/blob/4.12.x/src/SourceLocator/Type/Composer/Factory/MakeLocatorForComposerJsonAndInstalledJson.php#L174

still have vendor hard-coded, which makes my test fail. From what I see I have no other option than give the `$vendorDir` as a function parameter and have it default to `vendor` like this 

```diff
/**
 * @param array{name: string} $package
 */
-private function packagePrefixPath(string $trimmedInstallationPath, array $package): string
+private function packagePrefixPath(string $trimmedInstallationPath, array $package, string $vendorDir = 'vendor'): string
{
-     return $trimmedInstallationPath . '/vendor/' . $package['name'] . '/';
+    return $trimmedInstallationPath . '/' . $vendorDir . '/' . $package['name'] . '/';
}
```

As I see this it would not be a breaking changes for two reasons. 1 it's a private function and a default value is assigned if not set.
Please correct me if I'm wrong. 

But I'm not sure how you like the approach, but haven't found a better way yet.

src/SourceLocator/Type/Composer/Factory/MakeLocatorForComposerJsonAndInstalledJson.php
```diff
@@ -38,19 +38,22 @@ public function __invoke(string $installationPath, Locator $astLocator): SourceL
...
-        $composerJsonPath  = $realInstallationPath . '/composer.json';
-        $installedJsonPath = $realInstallationPath . '/vendor/composer/installed.json';
+        $composerJsonPath = $realInstallationPath . '/composer.json';
 
         if (! file_exists($composerJsonPath)) {
             throw MissingComposerJson::inProjectPath($installationPath);
         }
 
+        /** @var array{autoload: array{classmap: array<int, string>, files: array<int, string>, psr-4: array<string, array<int, string>>, psr-0: array<string, array<int, string>>}, config: array{vendor-dir: array<string, array<int, string>>}}|null $composer */
+        $composer  = json_decode((string) file_get_contents($composerJsonPath), true);
```
Ocramius, line 53: BTW, since we now support 7.4+, we can use `JSON_THROW` here

src/SourceLocator/Type/Composer/Factory/MakeLocatorForComposerJsonAndInstalledJson.php
```diff
@@ -169,9 +174,9 @@ private function packageToFilePaths(array $package): array
     /**
      * @param array{name: string} $package
      */
-    private function packagePrefixPath(string $trimmedInstallationPath, array $package): string
+    private function packagePrefixPath(string $trimmedInstallationPath, array $package, string $vendorDir = 'vendor'): string
```
Ocramius: `$vendorDir` no longer needs a default, since it is always declared and passed in explicitly

src/SourceLocator/Type/Composer/Factory/MakeLocatorForComposerJsonAndInstalledJson.php
```diff
@@ -180,9 +185,9 @@ private function packagePrefixPath(string $trimmedInstallationPath, array $packa
      *
      * @return array<string, array<int, string>>
      */
-    private function prefixWithPackagePath(array $paths, string $trimmedInstallationPath, array $package): array
+    private function prefixWithPackagePath(array $paths, string $trimmedInstallationPath, array $package, string $vendorDir = 'vendor'): array
```
Ocramius: `$vendorDir` no longer needs a default, since it is always declared and passed in explicitly

src/SourceLocator/Type/Composer/Factory/MakeLocatorForComposerJsonAndInstalledJson.php
Ocramius: No, `??` is very much lax :-)

src/SourceLocator/Type/Composer/Factory/MakeLocatorForComposerJsonAndInstalledJson.php
```diff
@@ -38,19 +38,22 @@ public function __invoke(string $installationPath, Locator $astLocator): SourceL
...
 
-        $composerJsonPath  = $realInstallationPath . '/composer.json';
-        $installedJsonPath = $realInstallationPath . '/vendor/composer/installed.json';
+        $composerJsonPath = $realInstallationPath . '/composer.json';
 
         if (! file_exists($composerJsonPath)) {
             throw MissingComposerJson::inProjectPath($installationPath);
         }
 
+        /** @var array{autoload: array{classmap: array<int, string>, files: array<int, string>, psr-4: array<string, array<int, string>>, psr-0: array<string, array<int, string>>}, config: array{vendor-dir: array<string, array<int, string>>}}|null $composer */
```
Ocramius: Can we split this into multiline?

```php
/**
 * @var array{
 * ...
 * }
 */
```

Ocramius: `vendor-dir` is incorrectly mapped here.

Currently: `vendor-dir: array<string, array<int, string>>`

Correct: `vendor-dir?: string`

Ocramius (changes requested):
Looking good! Need some minor adjustments, but overall looks like we're almost there :)

tomasnorre:
> Looking good! Need some minor adjustments, but overall looks like we're almost there :)

Super. I hope I'll be able to finished with respect to your feedback within the next days.

src/SourceLocator/Type/Composer/Factory/MakeLocatorForComposerJsonAndInstalledJson.php
tomasnorre: I'm not sure I have done this correctly. 
I have added the default `512` as depth.

src/SourceLocator/Type/Composer/Factory/MakeLocatorForComposerJsonAndInstalledJson.php
tomasnorre: The `phpcs` complaines about the multiline annotation. 

``` 
 49 | ERROR | Invalid inline documentation comment format "@var array{
    |       |          * autoload: array{classmap: array<int, string>, files: array<int, string>, psr-4: array<string, array<int,
    |       | string>>, psr-0: array<string, array<int, string>>},
    |       |          * config: array{vendor-dir?: string}
    |       |          * }|null $composer", expected "@var type $variable".
    |       | (SlevomatCodingStandard.Commenting.InlineDocCommentDeclaration.InvalidFormat)

``` 
But don't know how to deal with it.

src/SourceLocator/Type/Composer/Factory/MakeLocatorForComposerJsonAndInstalledJson.php
```diff
@@ -38,19 +40,26 @@ public function __invoke(string $installationPath, Locator $astLocator): SourceL
...
         if (! file_exists($composerJsonPath)) {
             throw MissingComposerJson::inProjectPath($installationPath);
         }
 
+        /** @var array{
+         * autoload: array{classmap: array<int, string>, files: array<int, string>, psr-4: array<string, array<int, string>>, psr-0: array<string, array<int, string>>},
+         * config: array{vendor-dir?: string}
+         * }|null $composer
+         */
+        $composer  = json_decode((string) file_get_contents($composerJsonPath), true, 512, JSON_THROW_ON_ERROR);
```
tomasnorre: I think I have done the `JSON_THROW_ON_ERROR` wrong. The test `testWillFailToProduceLocatorForProjectWithInvalidComposerJson` fails now as the exection type isn't the one expected anymore. 

I didn't know the `JSON_THROW` you talked about but this was what found, so really not sure if that was what you meant @Ocramius?

src/SourceLocator/Type/Composer/Factory/MakeLocatorForComposerJsonAndInstalledJson.php
Ocramius: Oh, use `/** @psalm-var` and phpcs will shut up :-)

src/SourceLocator/Type/Composer/Factory/MakeLocatorForComposerJsonAndInstalledJson.php
Ocramius: Oh, I thought `testWillFailToProduceLocatorForProjectWithInvalidComposerJson` was not handled at all. If we already have tests for that, I would not add `JSON_THROW_ON_ERROR` :-)

src/SourceLocator/Type/Composer/Factory/MakeLocatorForComposerJsonAndInstalledJson.php
tomasnorre: If you say so, didn't want to ignore it till getting feedback. I'm trying to learn and contribute, not to screw your library :)

tomasnorre:
BTW do you squash before merge? Or should I do that before changing this from Draft to a PR?

Ocramius:
@tomasnorre up to you - the only thing I'd remove is the `[WIP]` from the commit messages

Ocramius (approved):
LGTM :ship:

tomasnorre:
> LGTM

Looks good to merge?

Issue #701, opened 2020-10-30
PHP 8 compatibility


ondrejmirtes:
I already did most of these steps [in my fork](https://github.com/ondrejmirtes/BetterReflection) because I want PHPStan to be ready for PHP 8 on day one or sonner.

This is what needs to be done in BetterReflection in order to fully support PHP 8:

- [x] Adapter signatures compatibility (https://github.com/ondrejmirtes/BetterReflection/commit/0acd4522bd500547cf2574acb3b1a04b177869ae)
- [x] `mixed` type (https://github.com/ondrejmirtes/BetterReflection/commit/fabd25923c2a1ec96e3bedbdd09267d88a150f15)
- [x] `static` type (https://github.com/ondrejmirtes/BetterReflection/commit/e6f9b722e8af66566caad705cd541c213bd4d106)
- [x] Reflecting union types (https://github.com/ondrejmirtes/BetterReflection/commit/ce858eba391542639e8d2ea725df27ffaeca4e01 - this commit needs to be broken up because it mixes more things)
- [x] `ReflectionType::__toString` breaking change in PHP 8 - nullable type includes `?` again (https://3v4l.org/CmTA1)
- [x] `ReflectionSourceStubber` misses to generate return types (https://github.com/ondrejmirtes/BetterReflection/commit/6a87721f5c95792d754b30931b283c91897995a6)
- [x]  `isBuiltin()` is supposed to be only on `ReflectionNamedType` in PHP 8 (https://github.com/ondrejmirtes/BetterReflection/commit/86c5ce76560c561209cc36f05d576d0bc761ede0)
- [x] `Stringable` (https://github.com/ondrejmirtes/BetterReflection/commit/e9f99da6586dab440cc4117379fdbdb912a1697e)
- [x] `PhpStormStubsSourceStubberTest` will fail a lot when building on PHP 8 because a lot of stuff changed between PHP 8 and PHP 7
- [x] Promoted properties in constructor - I plan to solve this by making ReflectionProperty an interface with two implementations - one based on `Property` node, the other based on `Param` node.
- [x] Only `__construct` methods can be considered constructors
- [x] Support for reading Attributes
- [x] Constant of value `null` as a default parameter value no longer makes the parameter type nullable

The `Stringable` commit and the obsoleted old-style constructors require an `if` based on the PHP version. I did that with a public static property on `BetterReflection` but a different way can be fine too.

Some of these items require a new major version, but some of them can be done in a minor version before that.

[...]

PR #712, opened 2020-12-12
PHP8 - Adapter Signatures Compatibility

michaeljoelphillips:
This PR takes a stab at the first item listed in #701 concerning PHP8 support.

What should be done about the missing constants on `\ReflectionClassConstant` for PHP < 8.0?  In this PR I am declaring them on `Roave\BetterReflection\Reflection\ReflectionClassConstant`.  I am assuming that the expectation for users on PHP7 will be to have some access to constants for the filter values, but I am open to feedback.

Since they are currently undocumented, here are the constants for reference:
```
php > $reflection = new ReflectionClass(ReflectionClassConstant::class);
php > var_dump($reflection->getConstants());
array(3) {
  ["IS_PUBLIC"]=>
  int(1)
  ["IS_PROTECTED"]=>
  int(2)
  ["IS_PRIVATE"]=>
  int(4)
}
```

### DIFF
--- a/src/Reflection/ReflectionClass.php
+++ b/src/Reflection/ReflectionClass.php
@@ -490,16 +490,14 @@ class ReflectionClass implements Reflection
     /**
      * Get an associative array of the defined constants in this class,
      * with keys as constant names and values as constant values.
-
-     * @see ReflectionClass::getConstants() for the usage of filter
      *
      * @return array<string, scalar|array<scalar>|null>
      */
-    public function getConstants(?int $filter = null): array
+    public function getConstants(): array
     {
         return array_map(static function (ReflectionClassConstant $classConstant) {
             return $classConstant->getValue();
-        }, $this->getReflectionConstants($filter));
+        }, $this->getReflectionConstants());
     }
 
     /**
@@ -586,11 +584,9 @@ class ReflectionClass implements Reflection
      * Get an associative array of the defined constants in this class,
      * with keys as constant names and values as {@see ReflectionClassConstant} objects.
      *
-     * @see ReflectionClass::getConstants() for the usage of filter
-     *
      * @return array<string, ReflectionClassConstant> indexed by name
      */
-    public function getReflectionConstants(?int $filter = null): array
+    public function getReflectionConstants(): array
     {
         // Note: constants are not merged via their name as array index, since internal PHP constant
         //       sorting does not follow `\array_merge()` semantics
@@ -628,16 +624,7 @@ class ReflectionClass implements Reflection
             $reflectionConstants[$constantName] = $constant;
         }
 
-        if ($filter === null) {
-            return $reflectionConstants;
-        }
-
-        return array_filter(
-            $reflectionConstants,
-            static function (ReflectionClassConstant $constant) use ($filter): bool {
-                return (bool) ($filter & $constant->getModifiers());
-            },
-        );
+        return $reflectionConstants;
     }
 
     /**
--- a/src/Reflection/ReflectionClass.php
+++ b/src/Reflection/ReflectionClass.php
@@ -490,14 +490,16 @@ class ReflectionClass implements Reflection
     /**
      * Get an associative array of the defined constants in this class,
      * with keys as constant names and values as constant values.
+
+     * @see ReflectionClass::getConstants() for the usage of filter
      *
      * @return array<string, scalar|array<scalar>|null>
      */
-    public function getConstants(): array
+    public function getConstants(?int $filter = null): array
     {
         return array_map(static function (ReflectionClassConstant $classConstant) {
             return $classConstant->getValue();
-        }, $this->getReflectionConstants());
+        }, $this->getReflectionConstants($filter));
     }
 
     /**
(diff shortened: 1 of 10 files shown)


src/Reflection/ReflectionClassConstant.php
```diff
@@ -15,6 +15,10 @@
 
 class ReflectionClassConstant
 {
+    public const IS_PUBLIC    = 1;
```
Ocramius: I think these shouldn't really be declared here - perhaps we should just use the built-in constants?

src/Reflection/ReflectionClass.php
```diff
@@ -490,14 +490,16 @@ public function getImmediateConstants(): array
     /**
      * Get an associative array of the defined constants in this class,
      * with keys as constant names and values as constant values.
+
+     * @see ReflectionClass::getConstants() for the usage of filter
      *
      * @return array<string, scalar|array<scalar>|null>
      */
-    public function getConstants(): array
+    public function getConstants(?int $filter = null): array
```
Ocramius: Given that child classes may extend this, I think it's to be labeled as BC break :thinking:

src/Reflection/ReflectionClass.php
```diff
@@ -584,9 +586,11 @@ static function (ReflectionClassConstant $constant): string {
      * Get an associative array of the defined constants in this class,
      * with keys as constant names and values as {@see ReflectionClassConstant} objects.
      *
+     * @see ReflectionClass::getConstants() for the usage of filter
+     *
      * @return array<string, ReflectionClassConstant> indexed by name
      */
-    public function getReflectionConstants(): array
+    public function getReflectionConstants(?int $filter = null): array
```
Ocramius: I don't think the `$filter` as a bitmask is a good API: sure, it perhaps worked in php-src, but it's a very much shitty approach that C programmers keep perpetuating and imposing upon sensible developers and scopes where memory utilization isn't needed to this level.

I think this `array_filter()` should be done in `Adapter\ReflectionClass#getReflectionConstants()` instead.

Ocramius (changes requested):
Patch looks good, but we need to get rid of the design implications of core reflection API leaking into the new API: filtering logic should therefore be moved from the `src/Reflection` to the `src/Reflection/Adapter` logic, to avoid having ugly/complex logic like bitmask filtering exposed in the BC-compliant API.

Ocramius:
Moved to `4.13.0`. Reminder to self: change base branch after `4.12.0` is released

src/Reflection/ReflectionClass.php
michaeljoelphillips: Thanks for pointing this out, the filter logic has been moved to the adapter.  :+1:

src/Reflection/ReflectionClassConstant.php
michaeljoelphillips: I wasn't sure what to do about them for users of PHP7, but now that the filter logic is gone, so are these constants.

michaeljoelphillips:
@Ocramius, I moved the filter logic to the adapter layer for both `ReflectionClass` and `ReflectionObject` and added tests.  Thanks for the feedback.

test/unit/Reflection/Adapter/ReflectionClassTest.php
```diff
@@ -78,7 +79,6 @@ public function methodExpectationProvider(): array
             ['getProperties', null, [$mockProperty], []],
             ['hasConstant', null, true, ['foo']],
             ['getConstant', null, 'a', ['foo']],
-            ['getConstants', null, ['a', 'b'], []],
```
Ocramius, line 81: Even considering the new test additions, shouldn't these data-provider entries be kept? In theory, the changes you introduced are BC compliant, so this should work as before 🤔

Ocramius (changes requested):

Ocramius:
@michaeljoelphillips can you expand on https://github.com/Roave/BetterReflection/blob/ecd6c1e0af34df5bc2d94f850f2b37f1dc29b782/.github/workflows/phpunit.yml#L17-L27 to add `php: 8.0` to it, perhaps?

test/unit/Reflection/Adapter/ReflectionClassTest.php
michaeljoelphillips: The previous implementation relied upon `Reflection\ReflectionClass::getConstants()` which returns `array<string, scalar|array<scalar>|null>`.  Since we are keeping the filter logic out of the BR API, I chose not to delegate to `Reflection\ReflectionClass::getConstants()` but instead to `Reflection\ReflectionClass::getReflectionConstants()` so that I could filter on the visibility with `Reflection\ReflectionClassConstant::getModifiers()`.

I agree that removing this line in the data provider feels wrong.  I'm open to suggestions on alternatives.  Not changing this method's delegation to `ReflectionClass::getConstants()` will still require access to each `ReflectionClassConstant` object in order to compare visibility.

Ocramius (approved):
LGTM :ship: 

The addition of PHP 8.x to CI will happen separately, I suppose :-)

Thanks for the clarification comments and work on this, @michaeljoelphillips!

PR #713, opened 2020-12-13
PHP8 - Add support for `mixed` and `static` types

michaeljoelphillips:
This PR addresses two more items in #701.

The tests for this PR are failing on 7.4 with the addition of a `static` return type in one of the fixture class files.  This, and the issue with class constants that I mentioned in #712 makes me wonder if there will be a plan to drop PHP7 support with the release of BR supporting PHP8.  Please advise on how I should handle these test cases.

Also, please let me know if I'm missing tests, I'm not sure that I caught all of the places that needed refactoring.

### DIFF
--- a/src/Reflection/Adapter/ReflectionNamedType.php
+++ b/src/Reflection/Adapter/ReflectionNamedType.php
@@ -44,7 +44,7 @@ class ReflectionNamedType extends CoreReflectionNamedType
     {
         $type = (string) $this->betterReflectionType;
 
-        if ($type === 'self' || $type === 'parent') {
+        if ($type === 'self' || $type === 'parent' || $type === 'static') {
             return false;
         }
 
--- a/src/Reflection/Adapter/ReflectionType.php
+++ b/src/Reflection/Adapter/ReflectionType.php
@@ -39,7 +39,7 @@ class ReflectionType extends CoreReflectionType
     {
         $type = (string) $this->betterReflectionType;
 
-        if ($type === 'self' || $type === 'parent') {
+        if ($type === 'self' || $type === 'parent' || $type === 'static') {
             return false;
         }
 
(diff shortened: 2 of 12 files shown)


Ocramius:
PHPUnit tests seem to fail on `static` due to the parser using PHP 7 compatibility:

```
                                                                                
         There was 1 error:                                                     
                                                                                
         1)                                                                     
         Roave\BetterReflectionTest\SourceLocator\SourceStubber\ReflectionSource
         StubberTest::testClassStub                                             
         ParseError: syntax error, unexpected 'static' (T_STATIC)               
                                                                                
         /home/runner/work/BetterReflection/BetterReflection/test/unit/Fixture/C
         lassForSourceStubber.php:147                                           
                             
```

I think the defaults of the parser need to be changed:

https://github.com/Roave/BetterReflection/blob/df8a13c8aafc5036d2c65884acd02d006905a433/src/BetterReflection.php#L80

Ocramius (approved):
Besides parser changes, this LGTM :+1: 

Waiting for green to merge though :D

Ocramius:
Moved to `4.13.0`. Reminder to self: change base branch after `4.12.0` is released

michaeljoelphillips:
@Ocramius, I think [this is the offending block of code](https://github.com/Roave/BetterReflection/blob/4.12.0/test/unit/SourceLocator/SourceStubber/ReflectionSourceStubberTest.php#L105) causing the error:
```php
public function testClassStub(): void
{
    require __DIR__ . '/../../Fixture/ClassForSourceStubber.php';
    ...
}
```
Since this is being `required`, I believe it is the PHP74 runtime balking at the `static` return type.  Should I create a separate fixture and test case that can be skipped if not running on PHP8?

michaeljoelphillips:
@Ocramius, please check 9c164946b and let me know if that works for you.

Ocramius (approved):
>  Should I create a separate fixture and test case that can be skipped if not running on PHP8?

I see you already did 😁

Excellent work, thank you!

Ocramius:
@michaeljoelphillips I just noticed that our CI pipeline has no PHP 8.0 in it yet - big problem, but will bring it up in the next patch.

PR #716, opened 2020-12-16
[BUGFIX] Add custom `vendor-dir` to `MakeLocationForInstalledJson` + Correcting exception messages

### LINKED ISSUE #714: Misleading exception message
After adding #711, the exception message should be correct. I overlooked this as part of my PR before. 

I've just seen that the exception needs to be adjusted. It still contains path-info to vendor/composer/installed.json

https://github.com/Roave/BetterReflection/blob/ecd6c1e0af34df5bc2d94f850f2b37f1dc29b782/src/SourceLocator/Type/Composer/Factory/Exception/MissingInstalledJson.php#L16

It's not critical, but misleading, but I would need to add the vendor-dir to the input variables for the exception otherwise the vendor-dir would be unknown to the function. 

Feel free to assign me and I'll try to cleanup my mess.

tomasnorre:
Resolves #714

### DIFF
--- a/src/SourceLocator/Type/Composer/Factory/Exception/MissingInstalledJson.php
+++ b/src/SourceLocator/Type/Composer/Factory/Exception/MissingInstalledJson.php
@@ -13,7 +13,7 @@ final class MissingInstalledJson extends UnexpectedValueException implements Exc
     public static function inProjectPath(string $path): self
     {
         return new self(sprintf(
-            'Could not locate a "vendor/composer/installed.json" file in "%s"',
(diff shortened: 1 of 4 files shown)


asgrim (approved):

asgrim:
@Ocramius milestone 4.12.1 correct for this? Also does `laminas/automatic-releases` create a forward-port to 4.13.x / 5.0.x branches?

Ocramius (approved):
LGTM :+1:

Ocramius:
Thanks @tomasnorre! Rolling a release now :-)

PR #719, opened 2020-12-16
Use `$vendorDir` in `packagePrefixPath` to avoid hardcoded directory

### LINKED ISSUE #718: Custom `vendor-dir` not working in `Psr4Mapping`
I'm really sorry, that I'm not doing my homework good enough. I have found another issue with my change. Will try to have it don't fully this time. 

Sorry. 

**Affected:** 4.12.1 & 4.13.x

**Problem:** 

Even with the latest fix from 4.12.1 the custom vendor-dir isn't working, the `Psr4Mapping` is throwing an exception as it still thin that the folders should be found in `*/vendor/*`which isn't the case.

```shell
Provided path "Assert\" for prefix "/tmp/api-compare-06bfa8083aa22f6fdfe83de9807969b11f1e90c5_5fda50f9734d9/vendor/beberlei/assert/lib/Assert" is not a directory   
```
I have been able to verify this over tests, which I didn't add last time, otherwise I would have found this before merging.
Will grep through the repository for all `vendor` occurrences to "make sure" then are fixed this time. I'm really feeling bad about this.

Here is one of the problems:
https://github.com/Roave/BetterReflection/blob/9bfdcdd08da048c5522c7feed95bed85d3ef5254/src/SourceLocator/Type/Composer/Factory/MakeLocatorForInstalledJson.php#L178

tomasnorre:
I hope I got it right this time. 
Resolves: #718

### DIFF
--- a/src/SourceLocator/Type/Composer/Factory/MakeLocatorForInstalledJson.php
+++ b/src/SourceLocator/Type/Composer/Factory/MakeLocatorForInstalledJson.php
@@ -27,6 +27,7 @@ use function is_array;
 use function is_dir;
 use function json_decode;
 use function realpath;
+use function rtrim;
 
 final class MakeLocatorForInstalledJson
 {
--- a/src/SourceLocator/Type/Composer/Factory/MakeLocatorForInstalledJson.php
+++ b/src/SourceLocator/Type/Composer/Factory/MakeLocatorForInstalledJson.php
@@ -71,10 +71,10 @@ final class MakeLocatorForInstalledJson
 
         $classMapPaths       = array_merge(
             [],
-            ...array_map(function (array $package) use ($realInstallationPath): array {
+            ...array_map(function (array $package) use ($realInstallationPath, $vendorDir): array {
                 return $this->prefixPaths(
                     $this->packageToClassMapPaths($package),
-                    $this->packagePrefixPath($realInstallationPath, $package),
+                    $this->packagePrefixPath($realInstallationPath, $package, $vendorDir),
                 );
             }, $installed),
         );
@@ -82,10 +82,10 @@ final class MakeLocatorForInstalledJson
         $classMapDirectories = array_filter($classMapPaths, 'is_dir');
         $filePaths           = array_merge(
             [],
-            ...array_map(function (array $package) use ($realInstallationPath): array {
+            ...array_map(function (array $package) use ($realInstallationPath, $vendorDir): array {
                 return $this->prefixPaths(
                     $this->packageToFilePaths($package),
-                    $this->packagePrefixPath($realInstallationPath, $package),
+                    $this->packagePrefixPath($realInstallationPath, $package, $vendorDir),
                 );
             }, $installed),
         );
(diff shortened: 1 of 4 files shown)


src/SourceLocator/Type/Composer/Factory/MakeLocatorForInstalledJson.php
```diff
@@ -173,9 +175,9 @@ private function packageToFilePaths(array $package): array
     /**
      * @param array{name: string, autoload: array{classmap: array<int, string>, files: array<int, string>, psr-4: array<string, array<int, string>>, psr-0: array<string, array<int, string>>}} $package
      */
-    private function packagePrefixPath(string $trimmedInstallationPath, array $package): string
+    private function packagePrefixPath(string $trimmedInstallationPath, array $package, string $vendorDir): string
     {
-        return $trimmedInstallationPath . '/vendor/' . $package['name'] . '/';
+        return $trimmedInstallationPath . '/' . $vendorDir . '/' . $package['name'] . '/';
```
Ocramius, line 182: Can `$vendorDir` include leading or trailing `/`?

Ocramius (changes requested):

src/SourceLocator/Type/Composer/Factory/MakeLocatorForInstalledJson.php
tomasnorre: Will check. But valid question :)

tomasnorre: A tailing `/` in the composer.json config, breaks that test, so I'll look into this.

tomasnorre: I would need to sanitize the `$vendorDir` before handing it over. A `ltrim($vendorDir, "/")` and `rtrim($vendorDir, "/")` could be an option, but not really sure how I like it, and where to handel this. 

Do you have any input on this? 

I'll add a test case with leading and tailing `/` to catch this if anything changes. 
```json
{
  "config": {
    "vendor-dir": "/custom-vendor/"
  }
}
```

Ocramius: I think it's only `rtrim()`, since `/` means "absolute path", such as `/tmp/vendor`

tomasnorre: True. But the code base doesn't read it as an absolute path, This is the errors from the config example above. So if someone sets it with both leading and tailing, it has to be solved, or is that out of scope and one should know?

```shell
0 => '...custom-vendor-dir-using-composer-v2//custom-vendor//a/b/src/ab_PSR-4_Sources'
```

Ocramius: > or is that out of scope and one should know?

I'd say out of scope => new ticket for later. Using an absolute path for `vendor` is very harmful, and certainly goes against all principles on which composer is based upon. In addition to that, tools like `roave/backward-compatibility-check` wouldn't even work then (due to path collisions)

tomasnorre: I don't know if we are misunderstanding each other? 

Composer doesn't see the `/` in config: vendor-dir as an absolute path, it replaces the "vendor" in the full path. 

e.g. 
`/path/to/project/vendor/....` will be `/path/to/project/custom-vendor/....`

The question would be if composer validation should handle that as invalid configuration. To be honest I don't know how much validation is in the composer.json schema.

Ocramius: That's alright, hence why I propose to defer handling of a leading `/`, and just adjust things with `rtrim()` only

Ocramius (approved):
LGTM :+1:

Ocramius:
Thanks @tomasnorre!

tomasnorre:
You're welcome. Hopefully I got it right this time. At least the non-standard composer project works now on my setup, so probably it got fixed now.

PR #742, opened 2021-05-18
Use repeatable cache key from `spl_object_id()` instead of `spl_object_hash()` to reduce memory overhead

mvorisek:
see https://3v4l.org/b4Z1G

also more performant, ref https://github.com/php/php-src/blob/php-8.0.6/ext/spl/php_spl.c#L669

### DIFF
--- a/src/SourceLocator/Type/MemoizingSourceLocator.php
+++ b/src/SourceLocator/Type/MemoizingSourceLocator.php
@@ -11,7 +11,7 @@ use Roave\BetterReflection\Reflector\Reflector;
 
 use function array_key_exists;
 use function get_class;
-use function spl_object_hash;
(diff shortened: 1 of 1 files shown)


Ocramius:
Overall only mergeable when/if CI runs with PHP 8.x.

mvorisek:
this can be merged immediatelly into 4.12.x, there was a condition for PHP 7.1 (which is still supported by phpstan), but as 4.12.x already does not support PHP 7.1, no condition is needed (as has nothing to do with PHP 8.0)

Ocramius:
Ah, just saw that you removed the conditional :+1:

Ocramius (approved):
Thanks @mvorisek!

PR #746, opened 2021-06-22
Update ReflectionFunctionAbstract.php

lanlin:
Fix The Error : Typed static property Roave\BetterReflection\Reflection\ReflectionFunctionAbstract::$parser must not be accessed before initialization

### DIFF
--- a/test/unit/Reflection/ReflectionFunctionAbstractTest.php
+++ b/test/unit/Reflection/ReflectionFunctionAbstractTest.php
@@ -745,6 +745,32 @@ PHP;
         self::assertTrue($function->isDeprecated());
     }
 
+    /**
+     * Set $parser = null in case the error below occur when excute with phpunit's data provider
+     * Error : Typed static property Roave\BetterReflection\Reflection\ReflectionFunctionAbstract::$parser must not be accessed before initialization
+     *
+     * @dataProvider getDataForTypedAccessTest
+     */
+    public function testAccessTypedConstantWithDataProvider(array $data): void
+    {
+        $x = ReflectionFunctionAbstract::CLOSURE_NAME;
+
+        $this->assertEquals('{closure}', $x);
+    }
+
+    /**
+     * data provider for test access typed constant
+     * 
+     * @return array[]
+     */
+    public function getDataForTypedAccessTest(): array
+    {
+        return [
+            [['aaa']],
+            [['bbb']],
+        ];
+    }
+
     public function deprecatedDocCommentsProvider(): array
     {
         return [
--- a/src/Reflection/ReflectionFunctionAbstract.php
+++ b/src/Reflection/ReflectionFunctionAbstract.php
@@ -55,7 +55,7 @@ abstract class ReflectionFunctionAbstract
 
     private Reflector $reflector;
 
-    private static ?Parser $parser;
+    private static ?Parser $parser = null;
 
     protected function __construct()
     {


Ocramius:
Access to this property is indeed not supposed to happen upfront: how can
we reproduce the crash?

Can you write a failing test case please?

lanlin:
It takes me much time to reproduce this error

PHP Version: 8.0.1
PHPUnit Version: 9.5

```php
require 'vendor/autoload.php';

use PHPUnit\Framework\TestCase;
use Roave\BetterReflection\Reflection\ReflectionFunctionAbstract;

class Tests extends TestCase
{
    // ------------------------------------------------------------------------------

    public function testA(): void
    {
        $a = ReflectionFunctionAbstract::CLOSURE_NAME;
    }

    // ------------------------------------------------------------------------------

    /**
     * @dataProvider getData
     */
    public function testB(array $data): void
    {
        $b = ReflectionFunctionAbstract::CLOSURE_NAME;
    }

    // ------------------------------------------------------------------------------

    /**
     * @return array[]
     */
    public function getData(): array
    {
        return [
            [['aaa']],
            [['bbb']],
        ];
    }

    // ------------------------------------------------------------------------------
}
```

when using with PHPUnit's `@dataProvider` this error occurred.

Ocramius:
> It takes me much time to reproduce this error

It takes ages to maintain libraries without test cases :-) 

Could you factor the above in a test in either https://github.com/Roave/BetterReflection/blob/86135105e7f6e9684efd65c688ab2c49cf240143/test/unit/Reflection/ReflectionFunctionAbstractTest.php#L44 or any of the concrete types around it?

lanlin:
@Ocramius ok, give me some time

test/unit/Reflection/ReflectionFunctionAbstractTest.php
```diff
@@ -745,6 +745,32 @@ function foo() {}', $comment);
...
+     * Set $parser = null in case the error below occur when excute with phpunit's data provider
+     * Error : Typed static property Roave\BetterReflection\Reflection\ReflectionFunctionAbstract::$parser must not be accessed before initialization
+     *
+     * @dataProvider getDataForTypedAccessTest
+     */
+    public function testAccessTypedConstantWithDataProvider(array $data): void
+    {
+        $x = ReflectionFunctionAbstract::CLOSURE_NAME;
+
+        $this->assertEquals('{closure}', $x);
```
Ocramius, line 758: This is verifying a constant value is equal something, and it does not use `$data`? :thinking: 

Is this a WIP test?

Ocramius (changes requested):

test/unit/Reflection/ReflectionFunctionAbstractTest.php
lanlin: it's nothing about the constant or the $data, run this test case with the original `private static ?Parser $parser;` will get error, but `private static ?Parser $parser = null;` not.

Ocramius: So the triggering factor is `ReflectionFunctionAbstract::CLOSURE_NAME;`?

Asking because I'm mostly trying to reduce this further into a smaller test case, rather than a side-effect of code floating within the test suite.

Ocramius: @lanlin looking at this again, can the test be formulated in form of something that actually calls `ReflectionFunctionAbstract`?

Does this test crash also when `@runInSeparateProcess` is used?

Ocramius:
Closing here: please re-open once there's a failing test case that shows an explicit failure, not by transitivity

PR #753, opened 2021-09-15
Coding standard with PHP 8 features

kukulich:
Each sniff enabled in separate commit so it's easy to remove the sniff if not approved.

### DIFF
--- a/src/NodeCompiler/CompileNodeToValue.php
+++ b/src/NodeCompiler/CompileNodeToValue.php
@@ -29,7 +29,7 @@ class CompileNodeToValue
      *
      * @throws Exception\UnableToCompileNode
      */
-    public function __invoke(Node $node, CompilerContext $context)
+    public function __invoke(Node $node, CompilerContext $context): string|int|float|bool|array|null
     {
         if ($node instanceof Node\Stmt\Expression) {
             return $this($node->expr, $context);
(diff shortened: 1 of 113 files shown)


kukulich:
@WyriHaximus For you info.

kukulich:
@Ocramius Next one :)

src/NodeCompiler/CompileNodeToValue.php
```diff
@@ -29,7 +29,7 @@ class CompileNodeToValue
      *
      * @throws Exception\UnableToCompileNode
      */
-    public function __invoke(Node $node, CompilerContext $context)
+    public function __invoke(Node $node, CompilerContext $context): string|int|float|bool|array|null
```
Ocramius, line 32: Ooooh, I like how `scalar` was implemented here :-)

test/unit/Reflection/ReflectionFunctionTest.php
```diff
@@ -119,18 +119,22 @@ public function testStaticCreationFromName(): void
 
     public function testCreateFromClosure(): void
     {
-        $myClosure  = static function () {
+        // phpcs:disable SlevomatCodingStandard.Functions.RequireArrowFunction
```
Ocramius, line 122: Good catch not to change these 👍

test/unit/Reflection/ReflectionTypeTest.php
```diff
@@ -18,8 +18,7 @@
  */
 class ReflectionTypeTest extends TestCase
 {
-    /** @var Reflector|MockObject */
```
Ocramius, line 21: Not important, but we'll need to use `&` here in future.

Ocramius (approved):

Ocramius:
Awesome, thanks! 👍

PR #758, opened 2021-09-16
Added support for passing `ReflectionClass` and `ReflectionObject` instances to `isSubclassOf` and `implementsInterface`

kukulich:


src/Reflection/Adapter/ReflectionClass.php
```diff
@@ -411,19 +411,20 @@ public function getParentClass()
     }
 
     /**
-     * {@inheritDoc}
+     * @psalm-suppress MethodSignatureMismatch
```
Ocramius, line 414: Is this suppression due to psalm's stubs being out of sync?

Ocramius (approved):

Ocramius:
🚢 

Didn't even know the API allowed for this 😵

src/Reflection/Adapter/ReflectionClass.php
kukulich: `MethodSignatureMismatch - src/Reflection/Adapter/ReflectionClass.php:417:61 - Argument 1 of Roave\BetterReflection\Reflection\Adapter\ReflectionClass::isSubclassOf has wrong type 'ReflectionClass|string', expecting 'ReflectionClass|class-string' as defined by ReflectionClass::isSubclassOf (see https://psalm.dev/042)`

I don't know how to solve it. Is there a way? `@param` does not work...

PR #759, opened 2021-09-16
Removed support of PHP4 style constructors

kukulich:


Ocramius (approved):
Happy to see this gone 💪

PR #761, opened 2021-09-17
Fixed `PhpStormStubsSourceStubber` for `Generator#throw()`

### LINKED ISSUE #658: BetterReflection says method Generator::throw() does not exist
This one is really tricky. The method comes prefixed in the stubs (https://github.com/JetBrains/phpstorm-stubs/blob/b1041ca15222138a1d3d3aa4afb096df215bdc2e/standard/_types.php#L189-L194) because the PHP-Parser couldn't parse the file otherwise - there can't be a userland method called `throw`.

So we can't generate stub data as `public function throw(...`. The question is on which layer this should be fixed. A simple if in ReflectionClass would do but that's ugly. But source locators are all just about source code as string, so we have to patch the AST (modifying the name the ClassMethod node) somewhere...

kukulich:


Ocramius:
Fixes #658

Ocramius (approved):
Nice and clean: a bit hacky with the magic strings, but it is what it is with prefixing 🤷‍♀️

PR #762, opened 2021-09-17
Trait case sensitivity and aliasing collision fixes

### LINKED ISSUE #663: Changed trait method visibility is not taken into account
```php
<?php
trait HelloWorld {
    public function sayHello() {
        echo 'Hello World!';
    }
}

// Change visibility of sayHello
class MyClass1 {
    use HelloWorld { sayHello as protected; }
}

// Alias method with changed visibility
// sayHello visibility not changed
class MyClass2 {
    use HelloWorld { sayHello as private myPrivateHello; }
}
```

BetterReflection keeps the original visibility from the trait.

In the AST this information is communicated through `PhpParser\Node\Stmt\TraitUseAdaptation\Alias::$newModifier` (integer constants from `Class_` most likely). Looks like this property isn't read in the source code at all.

Original issue: https://github.com/phpstan/phpstan/issues/3465#issuecomment-643757521

kukulich:
Reported by @ondrejmirtes

Fixes https://github.com/Roave/BetterReflection/issues/663 and other bugs.

### DIFF
--- a/src/Reflection/ReflectionClass.php
+++ b/src/Reflection/ReflectionClass.php
@@ -1068,7 +1068,7 @@ class ReflectionClass implements Reflection
         return sprintf(
             '%s::%s',
             $className,
-            $methodName,
+            strtolower($methodName),
         );
     }
 
--- a/src/Reflection/ReflectionClass.php
+++ b/src/Reflection/ReflectionClass.php
@@ -82,6 +82,9 @@ class ReflectionClass implements Reflection
     /** @var array<string, string>|null */
     private ?array $cachedTraitAliases = null;
 
+    /** @var array<string, int>|null */
+    private ?array $cachedTraitModifiers = null;
+
     /** @var array<string, string>|null */
     private ?array $cachedTraitPrecedences = null;
 
@@ -237,11 +240,18 @@ class ReflectionClass implements Reflection
     {
         $traitAliases     = $this->getTraitAliases();
         $traitPrecedences = $this->getTraitPrecedences();
+        $traitModifiers   = $this->getTraitModifiers();
 
         $methodAst = $method->getAst();
         assert($methodAst instanceof ClassMethod);
 
-        $methodHash   = $this->methodHash($method->getImplementingClass()->getName(), $method->getName());
+        $methodHash = $this->methodHash($method->getImplementingClass()->getName(), $method->getName());
+
+        if (array_key_exists($methodHash, $traitModifiers)) {
+            $methodAst        = clone $methodAst;
+            $methodAst->flags = ($methodAst->flags & ~ Node\Stmt\Class_::VISIBILITY_MODIFIER_MASK) | $traitModifiers[$methodHash];
+        }
+
         $createMethod = fn (?string $aliasMethodName): ReflectionMethod => ReflectionMethod::createFromNode(
             $this->reflector,
             $methodAst,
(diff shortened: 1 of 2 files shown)


src/Reflection/ReflectionClass.php
```diff
@@ -1019,16 +1030,44 @@ private function getTraitPrecedences(): array
...
+        $this->parseTraitUsages();
+
+        /** @return array<string, int> */
+        return $this->cachedTraitModifiers;
+    }
+
     private function parseTraitUsages(): void
     {
-        if ($this->cachedTraitAliases !== null && $this->cachedTraitPrecedences !== null) {
+        if ($this->cachedTraitAliases !== null && $this->cachedTraitModifiers !== null && $this->cachedTraitPrecedences !== null) {
```
Ocramius, line 1062: Having 3 different caches for the same thing is quite worrisome :D

src/Reflection/ReflectionClass.php
```diff
@@ -236,11 +240,18 @@ private function createMethodsFromTrait(ReflectionMethod $method): array
...
         $traitPrecedences = $this->getTraitPrecedences();
+        $traitModifiers   = $this->getTraitModifiers();
 
         $methodAst = $method->getAst();
         assert($methodAst instanceof ClassMethod);
 
-        $methodHash   = $this->methodHash($method->getImplementingClass()->getName(), $method->getName());
+        $methodHash = $this->methodHash($method->getImplementingClass()->getName(), $method->getName());
+
+        if (isset($traitModifiers[$methodHash])) {
```
Ocramius: Let's use `array_key_exists()` for these - `isset()` is just generally too risky for type-level regressions later on

src/Reflection/ReflectionClass.php
```diff
@@ -1038,12 +1077,20 @@ private function parseTraitUsages(): void
             foreach ($adaptations as $adaptation) {
                 $usedTrait = $adaptation->trait;
                 if ($usedTrait === null) {
-                    $usedTrait = $traitNames[0];
+                    $usedTrait = end($traitNames);
```
Ocramius, line 1080: Did the index change in upstream `nikic/php-parser`?

test/unit/Reflection/ReflectionClassTest.php
```diff
@@ -1882,4 +1882,122 @@ class Foo
...
+            {
+                protected function myMethod() : void{
+                
+                }
+            }
+            
+            class HelloWorld
+            {
+                use MyTrait {
+                    MyMethod as myRenamedMethod;
```
Ocramius, line 1901: Sneaky casing issue :O

Ocramius (changes requested):

src/Reflection/ReflectionClass.php
kukulich: No, we just need to read the latest.

src/Reflection/ReflectionClass.php
kukulich: I will remove the new one and combine with one of the previous. It should be easy.

kukulich: Hmm, `getTraitAliases()` is `public` so it would be more difficult...

kukulich: I suggest to ignore it for now... :)

Ocramius: Let's leave it alone - wasn't suggesting action is needed, but rather warning about ever-raising complexity.

Ocramius (approved):
LGTM 👍

PR #763, opened 2021-09-17
expand `CompileNodeToValue` to support magic constants in functions, traits and constant declarations

### LINKED ISSUE #434: Extend `CompilerContext` so that it doesn't just track the current class, but also the current function
As discussed in #425, the `CompilerContext` is very limited in scope, and doesn't tell us whether it ia acting in a function or file, but just which class it is currently about: this means that exceptions will not contain any useful information if a failure in evaluating expressions happens inside a function.

### LINKED ISSUE #645: CompileNodeToValue does not recognize all magic constants
PoC fix: https://github.com/ondrejmirtes/BetterReflection/commit/5a662db404d78d8a9e5b55f408c3973d5d04ebe5

Original issue: https://github.com/phpstan/phpstan/issues/3405

### LINKED ISSUE #657: CompilerContext::getFileName() works only in class context
When a constant is defined outside of class:

```
define('fizz', __DIR__);
```

CompileNodeToValue fails: `Uncaught RuntimeException: The current context does not have a class for self`

Original issue: https://github.com/phpstan/phpstan/issues/3433

The fix is to pass in filename separately in CompilerContext's constructor and do not rely on `$this–>self` which contains ReflectionClass.

kukulich:
Fixes #434
Fixes #645 
Fixes #657

### DIFF
--- a/src/NodeCompiler/CompileNodeToValue.php
+++ b/src/NodeCompiler/CompileNodeToValue.php
@@ -17,6 +17,7 @@ use function defined;
 use function dirname;
 use function realpath;
 use function reset;
+use function sprintf;
 
 class CompileNodeToValue
 {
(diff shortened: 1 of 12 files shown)


src/NodeCompiler/Exception/UnableToCompileNode.php
```diff
@@ -66,11 +69,25 @@ public static function becauseOfNotFoundConstantReference(
         return $exception;
     }
 
+    private static function getFileName(CompilerContext $fetchContext): string
+    {
+        return $fetchContext->getFileName() ?? '""';
```
Ocramius, line 74: Is it worth stating "no file" rather than "" in this case? Not important, since we didn't even have the file name before

src/NodeCompiler/CompilerContext.php
```diff
@@ -5,41 +5,59 @@
...
+    }
+
+    public function getNamespace(): ?string
     {
-        return $this->self !== null;
+        return $this->namespace;
     }
 
-    public function getSelf(): ReflectionClass
+    public function inClass(): bool
```
Ocramius, line 36: Overall no big deal with this design, since it's `@internal`, but we should probably avoid `hasFoo()` and `getFoo` design when possible. A union type with an enumeration or such would be preferable in future.

src/NodeCompiler/CompileNodeToValue.php
```diff
@@ -48,10 +49,49 @@ public function __invoke(Node $node, CompilerContext $context): string|int|float
...
+                return $context->getNamespace() ?? '';
+            }
+
+            if ($node instanceof Node\Scalar\MagicConst\Method) {
+                if ($context->inClass() && $context->inFunction()) {
+                    return sprintf('%s::%s', $context->getClass()->getName(), $context->getFunction()->getName());
+                }
+
+                if ($context->inFunction()) {
+                    return $context->getFunction()->getName();
```
Ocramius, line 74: Didn't even know `__METHOD__` worked in functions 😯

https://3v4l.org/fr6E6

Ocramius (approved):
🚢  stellar work, @kukulich! 🤯

src/NodeCompiler/CompilerContext.php
kukulich: I'm happy to remove this methods. I used the pattern just because it was there... Give me 10 minutes :)

Ocramius: No need to, just keep it in mind for your future machinations 😁

kukulich: https://github.com/Roave/BetterReflection/pull/766 ;)

PR #764, opened 2021-09-17
It's possible to get default parameter value for internal methods/functions in PHP 8

kukulich:


test/unit/SourceLocator/SourceStubber/ReflectionSourceStubberTest.php
```diff
@@ -295,11 +295,16 @@ private function assertSameParameterAttributes(
...
     ): void {
-        $parameterName = $original->getDeclaringClass()->getName()
-            . '#' . $originalMethod->getName()
-            . '.' . $original->getName();
+        $methodName    = $original->getDeclaringClass()->getName() . '#' . $originalMethod->getName();
+        $parameterName = $methodName . '.' . $original->getName();
 
         self::assertSame($original->getName(), $stubbed->getName(), $parameterName);
+
+        // PHP sometimes reports default value available but it throws error when trying to get the value
```
Ocramius: This comment seems to be out of sync with the code: I guess we have more default values declared in stubs than in PHP?

Ocramius (changes requested):

test/unit/SourceLocator/SourceStubber/ReflectionSourceStubberTest.php
kukulich: I've to check it again. I'm not able to understand my yesterday's code...

kukulich: It looks I fixed something else in the night... Fixed the right place now :)

Ocramius (approved):
🚢  (yes, will wait for CI)

Ocramius:
This is gonna simplify a lot of downstream usage by a lot 💪

PR #765, opened 2021-09-17
Adapter `ReflectionProperty`: Added `hasDefaultValue()` and `getDefaultValue()`

kukulich:


Ocramius (approved):
LGTM 👍

src/Reflection/Adapter/ReflectionProperty.php
```diff
@@ -161,4 +161,14 @@ public function isAccessible(): bool
     {
         return $this->accessible || $this->isPublic();
     }
+
+    public function hasDefaultValue(): bool
```
Ocramius, line 165: The fact that our test suite was green, yet wr had missing methods on the adapters, seems to indicate that we didn't check if all methods were re-declared in the child class 🤔

PR #766, opened 2021-09-18
Reduce `CompilerContext`  API to avoid temporal coupling of method calls

kukulich:


src/NodeCompiler/CompileNodeToValue.php
```diff
@@ -198,17 +204,20 @@ private function compileFileConstant(CompilerContext $context): string
...
     private function compileClassConstant(CompilerContext $context): string
     {
-        return $context->inClass() ? $context->getClass()->getName() : '';
+        return $context->getClass()?->getName() ?? '';
     }
 
     private function resolveClassNameForClassNameConstant(string $className, CompilerContext $context): string
     {
+        $classContext = $context->getClass();
+        assert($classContext instanceof ReflectionClass);
```
Ocramius: Let's use a `!== null` assertion here: avoids having an implicit inference to `ReflectionClass` if this type ever expands

Ocramius (changes requested):

Ocramius (approved):
🚢  thanks @kukulich!

PR #769, opened 2021-09-18
Added all missing methods to adapters and fixed tests

kukulich:


Ocramius (changes requested):
Note: we still may want to have a test that compares declared API on our adapters with declared API on core classes, and makes sure we re-declare all of it

kukulich:
@Ocramius It's part of this PR :)

I found the bug why the tests were not failing.

https://github.com/Roave/BetterReflection/pull/769/files#diff-7074af2cec059f8ff54c9478f98676c03aa3bab7776729c50f67e97fd0e76d83R38

Ocramius:
Hmm, I don't see a loop over any core reflection information though? 🤔

kukulich:
Test `testCoreReflectionMethods` has data provider `coreReflectionMethodNamesProvider`.

The `testCoreReflectionMethods` now checks that declaring class of all methods is adapter. Previously there was only `hasMethod` check and it does not work because all adapters extend core classes so all adapters always have all methods from core classes :)

Ocramius:
Aha, now I do understand, thanks 💪

Ocramius (approved):
Heh, sorry for thr hoops to get MT passing, but we did indeed add a lot of uncovered API 😅

PR #771, opened 2021-09-18
By default, configure PHP-Parser to always consider sources as if they were PHP 7+

kukulich:


Ocramius (approved):
Simple/easy improvement - users can still provide their own parser 👍

PR #772, opened 2021-09-18
One reflector to rule them all

kukulich:
One reflector to rule them all
one reflector to find them
one reflector to bring them all
and in the reflection bind them


kukulich:
This is PoC of big BC break :)

Motivation:

```php
<?php

namespace Foo;

const SOME_CONSTANT = 0;

class Boo
{
     public function method($parameter = SOME_CONSTANT)
     {}
}
```
It's not possible with `ClassReflector` and `StringSourceLocator`/`SingleFileSourceLocator` to resolve default value of the parameter.
`ClassReflector` can create the `ReflectionClass` but it's not possible to get the default value because it has to be resolved by `ConstantReflector` and there's no `ConstantReflector` in the `ReflectionClass`.  It only partly works thanks to `CompileNodeToValue` and global functions `defined` and `constant`.

I propose to have one `Reflector` that can resolve everything based on the `SourceLocator`.

It can also improve the types. There are a lot of places where current interface `Reflector` is as the parameter but we almost everytime expect that it's `ClassReflector`.

I've also removed the ugly circular reference in https://github.com/Roave/BetterReflection/blob/5.0.x/src/BetterReflection.php#L89

---

There's one class `Reflector` in the PR but it should be `interface` and there should be one implementation, eg. `BetterReflector`, `OneReflector (as 1Password) or any other name. Or the interface can be eg. `ReflectorInterface`.

kukulich:
@Ocramius Please look at this PR (when you have time) and tell me if it will be mergeable :) If yes, please also add the opinion about the names :) I’ll then make the build green.

asgrim (commented):
The concept here seems fine; there is a finite number of "types of symbols to reflect on" so growth is reasonably controlled, I think our concern originally (why we split into `ClassReflector`/`FunctionReflector`) is uncertainty over how many symbol types we'd want to reflect about. Now the library is much more defined, it seems reasonable.

I do also like that it solves the circular dependency issue :+1:

asgrim:
> There's one class `Reflector` in the PR but it should be `interface` and there should be one implementation, eg. `BetterReflector`, `OneReflector (as 1Password) or any other name. Or the interface can be eg. `ReflectorInterface`.

I'd say keep the interface named `Reflector` (this becomes clear where the BC break is then - because the interface is changed, rather than changing interface into a class). For the implementation name I'm not worried about, maybe `BetterReflector` is fine, but I'll let @Ocramius opine later on that :grin:

src/Reflector/Reflector.php
```diff
@@ -4,18 +4,131 @@
...
+            throw Exception\IdentifierNotFound::fromIdentifier($identifier);
+        }
+
+        return $classInfo;
+    }
+
+    /**
+     * Get all the classes available in the scope specified by the SourceLocator.
+     *
+     * @return ReflectionClass[]
```
Ocramius: Since the signature changed, I think it might be a good idea to return an `iterable` here for future-proofing our ability to use a lazy iterator

src/Reflector/Reflector.php
```diff
@@ -4,18 +4,131 @@
...
+            throw Exception\IdentifierNotFound::fromIdentifier($identifier);
+        }
+
+        return $functionInfo;
+    }
+
+    /**
+     * Get all the functions available in the scope specified by the SourceLocator.
+     *
+     * @return ReflectionFunction[]
```
Ocramius: Since the signature changed, I think it might be a good idea to return an `iterable` here for future-proofing our ability to use a lazy iterator

src/Reflector/Reflector.php
```diff
@@ -4,18 +4,131 @@
...
+            throw Exception\IdentifierNotFound::fromIdentifier($identifier);
+        }
+
+        return $constantInfo;
+    }
+
+    /**
+     * Get all the constants available in the scope specified by the SourceLocator.
+     *
+     * @return array<int, ReflectionConstant>
```
Ocramius: Since the signature changed, I think it might be a good idea to return an `iterable` here for future-proofing our ability to use a lazy iterator

Ocramius (changes requested):
@kukulich from my PoV, approach is very much valid! It's a major BC break, but easy to migrate to it

kukulich:
@Ocramius what do you think about the names of interface and the class?

kukulich:
@Ocramius @asgrim Could you please give me a clue about the class and interface names? :) This PR blocks other my PR.

Ocramius:
@kukulich `Reflector` for the interface, `DefaultReflector` for the implementation? WDYT?

`DefaultReflector` could also be `SourceLocatorBasedReflector` or such, but IMO not that important there.

kukulich:
@Ocramius @asgrim Ready to review :)

src/Reflector/DefaultReflector.php
```diff
@@ -0,0 +1,127 @@
...
+use Roave\BetterReflection\Identifier\IdentifierType;
+use Roave\BetterReflection\Reflection\ReflectionClass;
+use Roave\BetterReflection\Reflection\ReflectionConstant;
+use Roave\BetterReflection\Reflection\ReflectionFunction;
+use Roave\BetterReflection\Reflector\Exception\IdentifierNotFound;
+use Roave\BetterReflection\SourceLocator\Type\SourceLocator;
+
+use function assert;
+
+class DefaultReflector implements Reflector
```
Ocramius: We should make this `final`

Ocramius (changes requested):
One last change needed :D

Ocramius (approved):
:ship:

Ocramius:
Awesome, thanks @kukulich!

PR #777, opened 2021-09-20
Use `is_file()` instead of `file_exists()` because we really need to check only files

kukulich:


Ocramius (approved):
Straightforward improvement :+1:  :ship:

PR #778, opened 2021-09-20
Make sure that `PhpStormStubsSourceStubber` updates constants based on current runtime environment, when outdated or not defined in stubs

kukulich:


Ocramius:
I don't understand this patch: can the commit description be expanded a bit? :sweat_smile:

kukulich:
Stubs are often outdated or constants have different values on different systems, see comment for method `updateConstantValue` (the comment was already there - it just moved now)

So we already update the constant value based on the real value (based on `constant()` call).

The current version only updates constants in stubs defined by `define()` call. The patch fixes it so all types of constants are updated - class constants and constants defined by `const` keyword too.

I've spotted this when preparing PHP 8.1 version - some constants change in PHP 8.1.

test/unit/SourceLocator/SourceStubber/PhpStormStubsSourceStubberTest.php
```diff
@@ -587,4 +589,45 @@ public function testCaseSensitiveConstant(string $constantName): void
...
+     */
+    public function testUpdateConstantValue(): void
+    {
+        require __DIR__ . '/../../Fixture/FakeConstants.php';
+
+        $sourceStubber = new PhpStormStubsSourceStubber(BetterReflectionSingleton::instance()->phpParser());
+
+        $stubberReflection = new CoreReflectionClass($sourceStubber);
+
+        $stubDirectoryReflection = $stubberReflection->getProperty('stubsDirectory');
```
Ocramius, line 604: Hacky :P

Ocramius (approved):
LGTM :ship:

test/unit/SourceLocator/SourceStubber/PhpStormStubsSourceStubberTest.php
kukulich: Yes, two hours of creating this test :D

Ocramius: I know the feeling - having one of those come up for me too, on a client project :-\

PR #779, opened 2021-09-20
Clean ReflectionClass constants API

kukulich:
BetterReflection should have "better" API. Only adapters need ugly API.

It's BC break but why not in major version?

### DIFF
--- a/src/Reflection/Adapter/ReflectionObject.php
+++ b/src/Reflection/Adapter/ReflectionObject.php
@@ -202,7 +202,7 @@ class ReflectionObject extends CoreReflectionObject
      */
     public function getConstants(?int $filter = null): array
     {
-        $reflectionConstants = $this->betterReflectionObject->getReflectionConstants();
+        $reflectionConstants = $this->betterReflectionObject->getConstants();
 
         if ($filter !== null) {
             $reflectionConstants = array_filter(
@@ -222,7 +222,7 @@ class ReflectionObject extends CoreReflectionObject
      */
     public function getConstant($name)
     {
-        return $this->betterReflectionObject->getConstant($name);
+        return $this->betterReflectionObject->getConstantValue($name);
     }
 
     /**
@@ -230,7 +230,7 @@ class ReflectionObject extends CoreReflectionObject
      */
     public function getReflectionConstant($name)
     {
-        $betterReflectionConstant = $this->betterReflectionObject->getReflectionConstant($name);
+        $betterReflectionConstant = $this->betterReflectionObject->getConstant($name);
 
         if ($betterReflectionConstant === null) {
             return false;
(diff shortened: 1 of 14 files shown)


kukulich:
@Ocramius What do you think?

src/Reflection/Adapter/ReflectionObject.php
```diff
@@ -222,15 +222,15 @@ public function getConstants(?int $filter = null): array
      */
     public function getConstant($name)
     {
-        return $this->betterReflectionObject->getConstant($name);
+        return $this->betterReflectionObject->getConstantValue($name);
```
Ocramius, line 225: Do we even need `getConstantValue` in our API? Can't a `ReflectionConstant` provide that, when asked for it?

src/Reflection/ReflectionClass.php
```diff
@@ -501,9 +501,9 @@ public function hasMethod(string $methodName): bool
      *
      * @return array<string, scalar|array<scalar>|null>
      */
-    public function getImmediateConstants(): array
+    public function getImmediateConstantsValues(): array
```
Ocramius, line 504: IMO, let's drop this completely - let the userland consumer do the `array_map()` operations

src/Reflection/ReflectionClass.php
```diff
@@ -512,9 +512,9 @@ public function getImmediateConstants(): array
      *
      * @return array<string, scalar|array<scalar>|null>
      */
-    public function getConstants(): array
+    public function getConstantsValues(): array
```
Ocramius, line 515: IMO, let's drop this completely - let the userland consumer do the `array_map()` operations

src/Reflection/ReflectionClass.php
```diff
@@ -540,17 +540,17 @@ public function getConstant(string $name): string|int|float|bool|array|null
...
+        return $this->getConstant($name) !== null;
     }
 
     /**
      * Get the reflection object of the specified class constant.
      *
      * Returns null if not specified.
      */
-    public function getReflectionConstant(string $name): ?ReflectionClassConstant
+    public function getConstant(string $name): ?ReflectionClassConstant
```
Ocramius, line 551: In order to ease the migration to `5.0`, I think we should have a different name for any symbols that changed return type, or else the BC break becomes more subtle.

Yes, that muds things a bit, since we'd have `get*ReflectionConstant` or such, but it's still better than a hard break at type level, with the same name :thinking:

Ocramius (changes requested):

src/Reflection/ReflectionClass.php
kukulich: I'm not sure if the "cleanup" make sense then. Now it could make `getProperty()`, `getMethod()` and `getConstant()` consistent...

Ocramius: Let's leave it be for now then?

kukulich: Ok :)

PR #781, opened 2021-09-20
Removed dead code / unused psalm error suppressions

kukulich:


src/Reflection/ReflectionConstant.php
```diff
@@ -140,10 +140,7 @@ public function getName(): string
             return $this->getNameFromDefineFunctionCall($this->node);
         }
 
-        /**
-         * @psalm-suppress UndefinedPropertyFetch
```
Ocramius, line 144: Can we add `findUnusedPsalmSuppress="true"` to `psalm.xml`?

Ocramius (changes requested):

src/Reflection/ReflectionConstant.php
kukulich: Done.

Ocramius: Awesome, thanks!

PR #783, opened 2021-09-21
Use `is_file()` instead of `file_exists()` when we need to be sure that the checked locations are files, and not directories

kukulich:
I had probably some active filter in PHPStorm yesterday so I did not spot all places...


Ocramius:
I was wondering why all changes were only on `test/` :D

Ocramius (approved):

PR #784, opened 2021-09-21
Types, tests, cleanup and code coverage

kukulich:


src/SourceLocator/Ast/Strategy/NodeToReflection.php
```diff
@@ -24,11 +22,11 @@ class NodeToReflection implements AstConversionStrategy
      */
     public function __invoke(
         Reflector $reflector,
-        Node $node,
+        Node\Stmt\ClassLike|Node\Stmt\Function_|Node\Expr\Closure|Node\Expr\ArrowFunction|Node\Stmt\Const_|Node\Expr\FuncCall $node,
         LocatedSource $locatedSource,
         ?Node\Stmt\Namespace_ $namespace,
         ?int $positionInNode = null,
-    ): ?Reflection {
```
Ocramius, line 31: Good riddance of nullability :+1:

src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
```diff
@@ -396,6 +396,8 @@ private function getStubsDirectory(): string
             }
         }
 
+        // @codeCoverageIgnoreStart
```
Ocramius, line 399: Should drop this :thinking:

src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
```diff
@@ -396,6 +396,8 @@ private function getStubsDirectory(): string
...
         throw CouldNotFindPhpStormStubs::create();
+        // @codeCoverageIgnoreEnd
```
Ocramius, line 402: Should drop this :thinking:

src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
Ocramius: Alternatively: document why it's not covered (impossible scenario, I suppose?)

test/unit/SourceLocator/Ast/FindReflectionsInTreeTest.php
```diff
@@ -235,9 +235,8 @@ public function testInvokeCallsReflectNodesForNotGlobalDefine(): void
         $functionReflector->method('reflect')
             ->willReturn($mockFunctionReflection);
 
-        $strategy->expects($this->once())
-            ->method('__invoke')
-            ->willReturn($mockFunctionReflection);
+        $strategy->expects($this->never())
```
Ocramius, line 238: This is quite a bold expectation change? :thinking:

Ocramius (changes requested):

src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
kukulich: Yes, impossible scenario - not possible to change value of private constant.

test/unit/SourceLocator/Ast/FindReflectionsInTreeTest.php
kukulich: Yes, it's. The invoke is not called because now we don't create the `ReflectionFunction` instance from `function` node when we are looking for `constant` identifier.

And the `define` is invalid (not global `define`) so we don't call the `invoke` for the constant too.

Ocramius (approved):
:ship:

Ocramius:
Thanks @kukulich!

PR #785, opened 2021-09-23
Promoted properties support

kukulich:


src/Reflection/ReflectionParameter.php
```diff
@@ -449,6 +449,11 @@ public function canBePassedByValue(): bool
         return ! $this->isPassedByReference();
     }
 
+    public function isPromoted(): bool
+    {
+        return $this->node->flags !== 0;
```
Ocramius, line 454: Is there a more precise flag value to check? :thinking:

Ocramius (approved):
This turned out to be much cleaner than I though :O

src/Reflection/ReflectionClass.php
```diff
@@ -664,21 +665,52 @@ public function getImmediateProperties(?int $filter = null): array
...
+                        $this->declaringNamespace,
+                        $this,
+                        $this,
+                        false,
+                    );
+                    $properties[$property->getName()] = $property;
+                }
+            }
+
+            foreach ($this->node->getMethods() as $methodNode) {
```
Ocramius, line 683: Is there a faster way to find the ctor, than looping through these? (fine also if not)

kukulich: I think this is the most optimized way. We don't need to create any reflection, just iterate existing AST.

src/Reflection/ReflectionParameter.php
kukulich: Unfortunately not: https://github.com/nikic/PHP-Parser/blob/master/lib/PhpParser/Node/Param.php#L40

Ocramius: Yeah, only found one cryptic usage: `$this->semStack[$stackPos-(6-2)]` :D 

Let's roll it as-is :+1:

Ocramius:
Awesome++ 

Thanks @kukulich!

PR #788, opened 2021-09-25
Simplified `CompilerContext` internals, making it compatible with more reflection symbols

kukulich:


Ocramius (approved):
Nice simplification! 👍

PR #790, opened 2021-09-26
Improved precision of declared types in internal code and exposed return values

kukulich:


Ocramius (approved):
SO GOOD! Thanks @kukulich! :rocket:

PR #792, opened 2021-10-03
Add `ReflectionParameter::hasType()`

WinterSilence:
also it's fix invalid result of isArray/isCallable for union type

### DIFF
--- a/src/Reflection/ReflectionParameter.php
+++ b/src/Reflection/ReflectionParameter.php
@@ -478,7 +478,7 @@ class ReflectionParameter
     }
 
     /**
-     * Checks parameter type.
+     * Checks if a parameter has the type specified in `$type`
      */
     public function hasType(string|ReflectionNamedType $type, bool $ignoreDefaultValue = false): bool
     {
--- a/src/Reflection/ReflectionParameter.php
+++ b/src/Reflection/ReflectionParameter.php
@@ -29,13 +29,16 @@ use RuntimeException;
 use function assert;
 use function count;
 use function get_class;
+use function gettype;
 use function in_array;
+use function is_a;
 use function is_array;
 use function is_object;
 use function is_string;
 use function preg_match;
 use function sprintf;
 use function strtolower;
+use function ltrim;
 
 class ReflectionParameter
 {
@@ -47,14 +50,20 @@ class ReflectionParameter
 
     private int $parameterIndex;
 
-    /** @var scalar|array<scalar>|null */
-    private string|int|float|bool|array|null $defaultValue = null;
+    /** 
+     * @var scalar|array<scalar>|null
+     */
+    private string|int|float|bool|array|null $defaultValue;
 
     private bool $isDefaultValueConstant = false;
 
     private ?string $defaultValueConstantName;
 
     private Reflector $reflector;
+
+    private ?bool $isArray;
+
+    private ?bool $isCallable;
 
     private function __construct()
     {
@@ -396,6 +405,7 @@ class ReflectionParameter
      */
     public function setType(string $newParameterType): void
     {
+        $this->removeType();
         $this->node->type = new Node\Name($newParameterType);
     }
 
@@ -405,6 +415,8 @@ class ReflectionParameter
     public function removeType(): void
     {
         $this->node->type = null;
+        $this->isArray = null;
+        $this->isCallable = null;
     }
 
     /**
@@ -412,7 +424,10 @@ class ReflectionParameter
      */
     public function isArray(): bool
     {
-        return preg_match('~^\??array$~i', (string) $this->getType()) === 1;
+        if ($this->isArray === null) {
+            $this->isArray = $this->hasType('array');
+        }
+        return $this->isArray;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


Ocramius (changes requested):
This kind of change requires test additions to validate it

src/Reflection/ReflectionParameter.php
```diff
@@ -459,6 +477,48 @@ public function isDefaultValueConstant(): bool
         return $this->isDefaultValueConstant;
     }
 
+    /**
+     * Checks parameter type.
```
asgrim:
```suggestion
     * Checks if a parameter has the type specified in `$type`
```
or something similar? Just need something more descriptive

src/Reflection/ReflectionParameter.php
```diff
@@ -405,22 +415,30 @@ public function setType(string $newParameterType): void
     public function removeType(): void
     {
         $this->node->type = null;
+        $this->isArray = null;
```
asgrim, line 418: As these values are memoized, we either need to:

 * Remove the memoization of the `isArray` / `isCallable` checks
 * Add a test case that ensure memoization recalls the value, and also is reset when type is removed

src/Reflection/ReflectionParameter.php
```diff
@@ -396,6 +405,7 @@ public function hasType(): bool
      */
     public function setType(string $newParameterType): void
     {
+        $this->removeType();
```
asgrim, line 408: As the checks are memoized, we either need to:

 * Remove the memoization of the `isArray` / `isCallable` checks
 * Add a test case to ensure that memoization is removed when setting a new type of the parameter

src/Reflection/ReflectionParameter.php
```diff
@@ -459,6 +477,48 @@ public function isDefaultValueConstant(): bool
         return $this->isDefaultValueConstant;
     }
 
+    /**
+     * Checks parameter type.
+     */
+    public function hasType(string|ReflectionNamedType $type, bool $ignoreDefaultValue = false): bool
+    {
+        $aliases = ['boolean' => 'bool', 'integer' => 'int', 'double' => 'float', 'NULL' => 'null'];
```
asgrim, line 485: Could possibly be defined in a `private const`?

src/Reflection/ReflectionParameter.php
```diff
@@ -459,6 +477,48 @@ public function isDefaultValueConstant(): bool
...
+                return true;
+            }
+        }
+
+        if ($this->hasType()) {
+            $paramType = $this->getType();
+            $paramTypes = $paramType instanceof ReflectionNamedType
+                ? [$paramType]
+                : $paramType->getTypes();
+            foreach ($paramTypes as $paramType) {
```
asgrim, line 509: `$paramType` is redefined here (originally declared on L505), slightly confusing naming re-use.

test/unit/Reflection/ReflectionParameterTest.php
```diff
@@ -426,6 +426,26 @@ public function testIsArray(): void
         $arrayParam = $method->getParameter('arrayParameter');
         self::assertTrue($arrayParam->isArray());
     }
+    
+
+    public function testHasType(): void
```
asgrim, line 431: Think this test could be improved by adding a `@dataProvider` and checking every possible type we can. I worry that some types might trip this up without test cases proving otherwise

asgrim (changes requested):

asgrim:
> also it's fix invalid result of isArray/isCallable for union type

A separate test case should be added to verify the original bug is also fixed.

src/Reflection/ReflectionParameter.php
```diff
@@ -459,6 +477,48 @@ public function isDefaultValueConstant(): bool
         return $this->isDefaultValueConstant;
     }
 
+    /**
+     * Checks parameter type.
+     */
+    public function hasType(string|ReflectionNamedType $type, bool $ignoreDefaultValue = false): bool
```
kukulich, line 483: The method `hasType` already exists: https://github.com/Roave/BetterReflection/blob/5.0.x/src/Reflection/ReflectionParameter.php#L386

src/Reflection/ReflectionParameter.php
WinterSilence: @asgrim I'm don't see reason for this now

src/Reflection/ReflectionParameter.php
WinterSilence: @asgrim I just ran out of ideas for variable names (:

src/Reflection/ReflectionParameter.php
WinterSilence: @asgrim can you comment this? i can update method ` public function hasType(string|ReflectionNamedType $type = null, bool $ignoreDefaultValue = false): bool` to keep LSP

Ocramius:
@WinterSilence I've merged #803, which includes these changes - can you give it a shot and see if it works for you?

WinterSilence:
@Ocramius copying my changes without merging PR is not so friendly

Ocramius:
> @Ocramius #803 not same to current PR, only define method with same name


@WinterSilence if it's not the same scenario, I'd gladly adjust: does your test case still fail?

 > @Ocramius it's bad habit to close PR without understanding it

PR got closed by merge via `Fixes X` comment provided by @kukulich: @kukulich did put in the time to understand your work. Please don't assume malice.

 > @Ocramius copying my changes without merging PR is not so friendly

No copying, AFAIK - this was done in a different way, from what I see from the diff. Again, please don't assume malice: we're really just trying to improve the codebase, all of us in our free time, I'd add.

PR #793, opened 2021-10-03
Update infection badge target branch to 5.0.x

sasezaki:

PR #794, opened 2021-10-03
Change `push:` rule branch from `master` to `5.0.x` in workflow definitions

sasezaki:
Sorry. After #793, I noticed to have to change workflow branch to hook actions.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -6,7 +6,7 @@ on:
   pull_request:
   push:
     branches:
-      - "master"
+      - "5.0.x"
 
 jobs:
   benchmarks:
--- a/.github/workflows/coding-standards.yml
+++ b/.github/workflows/coding-standards.yml
@@ -6,7 +6,7 @@ on:
   pull_request:
   push:
     branches:
-      - "master"
+      - "5.0.x"
 
 jobs:
   coding-standards:
--- a/.github/workflows/compatibility.yml
+++ b/.github/workflows/compatibility.yml
@@ -6,7 +6,7 @@ on:
   pull_request:
   push:
     branches:
-      - "master"
+      - "5.0.x"
 
 jobs:
   compatibility:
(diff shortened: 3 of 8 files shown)


sasezaki:
Notes:

On this PR, I would to change only workflow files.

There remains `master` on other markdown.
```
$ git ls-files | xargs grep master | grep -v composer.lock | grep -v CONTRIBUTING.md
README.md:* [Test suite](https://github.com/Roave/BetterReflection/blob/master/test/README.md)
docs/features.md:See [Loading a class from a string](https://github.com/Roave/BetterReflection/tree/master/docs/usage.md#Loading-a-class-from-a-string)
docs/features.md:See [types documentation](https://github.com/Roave/BetterReflection/tree/master/docs/types.md)
docs/features.md:See [AST extraction documentation](https://github.com/Roave/BetterReflection/tree/master/docs/ast-extraction.md)
Binary file test/unit/Fixture/autoload.phar matches
```

But, these can be redirected to main branch(5.0.x) by Github.

.github/workflows/benchmarks.yml
```diff
@@ -6,7 +6,7 @@ on:
   pull_request:
   push:
     branches:
-      - "master"
+      - "5.0.x"
```
asgrim, line 9: I really wish there was a better way of doing this; the default branch changes every time there is a minor release. I found [this on Stack Overflow](https://stackoverflow.com/questions/64781462/github-actions-default-branch-variable) but I think it only goes in templates, not actual workflow files >.<

asgrim (approved):

asgrim:
This is otherwise fine :+1:

Ocramius:
My alternative for the future is to use https://github.com/laminas/laminas-ci-matrix-action/tree/71aa5bef47be3a9d3e1a3865f43caf16708ed847#usage

That would automate most of this stuff for us.

For now, this change is good as-is :)

Ocramius (approved):

Ocramius:
Thanks @sasezaki!

PR #795, opened 2021-10-03
Update ReflectionParameterTest.php

### LINKED ISSUE #791: Result of `BetterReflection\ReflectionParameter::isDefaultValueAvailable()` not similar to `\ReflectionParameter::isDefaultValueAvailable()`
~~~php
class Demo
{
    public function test($value = null)
    {
    }
    public function isDefaultValueAvailable(): bool
    {
        return (new \ReflectionParameter([$this, 'test'], 0))->isDefaultValueAvailable();
    }
}

$demo = new Demo();
var_dump($demo->isDefaultValueAvailable()); // result: true
~~~
`Roave\BetterReflection\Reflection\ReflectionParameter::isDefaultValueAvailable():`
https://github.com/Roave/BetterReflection/blob/74aaf7e9a446590f5a461203c379d0580c614faa/src/Reflection/ReflectionParameter.php#L278

WinterSilence:
Fixes #791

### DIFF
--- a/test/unit/Reflection/ReflectionParameterTest.php
+++ b/test/unit/Reflection/ReflectionParameterTest.php
@@ -489,7 +489,7 @@ class ReflectionParameterTest extends TestCase
 
         $betterParam = $method->getParameter('optionalParameter');
         $nativeParam = new \ReflectionParameter([Methods::class, 'methodWithOptionalParameters'], 1);
-        self::assertSame($betterParam->isDefaultValueAvailable(), $nativeParam>isDefaultValueAvailable());
+        self::assertSame($betterParam->isDefaultValueAvailable(), $nativeParam->isDefaultValueAvailable());
     }
 
     public function testParameterWithDefaultValueBeforeVariadicParameterShouldBeOptional(): void
--- a/test/unit/Reflection/ReflectionParameterTest.php
+++ b/test/unit/Reflection/ReflectionParameterTest.php
@@ -481,6 +481,16 @@ class ReflectionParameterTest extends TestCase
         self::assertFalse($secondParam->isOptional());
         self::assertFalse($secondParam->isDefaultValueAvailable());
     }
+    
+    public function testIsDefaultValueAvailable(): void
+    {
+        $classInfo = $this->reflector->reflect(Methods::class);
+        $method    = $classInfo->getMethod('methodWithOptionalParameters');
+
+        $betterParam = $method->getParameter('optionalParameter');
+        $nativeParam = new \ReflectionParameter([Methods::class, 'methodWithOptionalParameters'], 1);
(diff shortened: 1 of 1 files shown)


kukulich:
@WinterSilence Did you run the test on your localhost? I don’t think so because there’s parse error…

WinterSilence:
@kukulich you catch me (: fixed

kukulich:
@WinterSilence Still don't understand. Native reflection returns `true`, better reflection returns `true` too.

The PHPUnit test reports error - you still dit not run the test on your localhost...

WinterSilence:
@kukulich 
> Class Roave\BetterReflectionTest\Fixture\Methods was loaded during test testIsDefaultValueAvailable
Failed asserting that true is false.

that's not error in test, that's error in `ReflectionParameter::isDefaultValueAvailable()`

kukulich:
@WinterSilence no, the test should not report such error.

WinterSilence:
@kukulich you asked for test? here it is. test perfectly shows inconsistency of method's implementation

> Native reflection returns true, better reflection returns true too.

Did you run the test on your localhost? I don’t think so…

kukulich:
@WinterSilence Yes, I run the test on my localhost. And there’s no error when I fixed the PHPUnit report.

asgrim:
The test causes a class `\Roave\BetterReflectionTest\Fixture\Methods` to be loaded, which is not allowed in our test suite; since all of BR should be *static* analysed only.

That said, I disabled the assertion in `\Roave\BetterReflectionTest\TestListener::assertClassNotLoaded`, and removed method `\Roave\BetterReflectionTest\Fixture\Methods::methodWithNonOptionalDefaultValue` (since loading this class causes an error), and the test submitted here already passes:

```
/usr/bin/php /home/james/workspace/better-reflection/vendor/phpunit/phpunit/phpunit --configuration /home/james/workspace/better-reflection/phpunit.xml.dist --filter "/(Roave\\BetterReflectionTest\\Reflection\\ReflectionParameterTest::testIsDefaultValueAvailable)( .*)?$/" --test-suffix ReflectionParameterTest.php /home/james/workspace/better-reflection/test/unit/Reflection --teamcity
Testing started at 14:21 ...
PHPUnit 9.5.9 by Sebastian Bergmann and contributors.



Time: 00:00.025, Memory: 18.00 MB

OK (1 test, 1 assertion)
Process finished with exit code 0
```

asgrim:
@WinterSilence not sure exactly why your test does not pass on your machine, but as it stands, neither @kukulich nor myself can reproduce any issue using your test case I'm afraid :( if you can manage to make a test that does fail, please do ask for this to be re-opened with a failing test and we can investigate.

PR #796, opened 2021-10-06
Improved constants resolving to support constants declared on traits and outside classes

### LINKED ISSUE #519: Crash when asking for default value of LevelDB::__construct() $options parameter
```php
UnableToCompileNode: Could not locate constant "LEVELDB_SNAPPY_COMPRESSION" while evaluating expression in LevelDB at line 22
```

The constant is defined in PhpStorm stubs in the same file as LevelDB class.

The problem is when I get ReflectionParameter for the `$options` parameter and ask for the default value.

The resolution does not go through ConstantReflector at all.

### LINKED ISSUE #688: BetterReflection does not support namespaced constants in default property and parameter values
Code like this fails:

```php
class Foo
{
    public const SOME = \Test\Bar\BAZ;
}
```

It tries to find constant with name `Test`.

There are several occurences of getting `Name::$parts[0]` in BR which obviously doesn't work for this case.

PoC fix (notice also the unskipped `todo` test): https://github.com/ondrejmirtes/BetterReflection/commit/7749ef8753a65801993274ed9215ae717490bd4d

kukulich:
Fixes https://github.com/Roave/BetterReflection/issues/688
Fixes https://github.com/Roave/BetterReflection/issues/519

### DIFF
--- a/src/NodeCompiler/Exception/UnableToCompileNode.php
+++ b/src/NodeCompiler/Exception/UnableToCompileNode.php
@@ -10,7 +10,6 @@ use Roave\BetterReflection\NodeCompiler\CompilerContext;
 use Roave\BetterReflection\Reflection\ReflectionClass;
 
 use function assert;
-use function reset;
 use function sprintf;
 
 class UnableToCompileNode extends LogicException
@@ -53,9 +52,8 @@ class UnableToCompileNode extends LogicException
     public static function becauseOfNotFoundConstantReference(
         CompilerContext $fetchContext,
         Node\Expr\ConstFetch $constantFetch,
+        string $constantName,
     ): self {
-        $constantName = reset($constantFetch->name->parts);
-
         $exception = new self(sprintf(
             'Could not locate constant "%s" while evaluating expression in %s in file %s (line %d)',
             $constantName,
--- a/src/Reflection/ReflectionParameter.php
+++ b/src/Reflection/ReflectionParameter.php
@@ -11,7 +11,6 @@ use LogicException;
 use OutOfBoundsException;
 use phpDocumentor\Reflection\Type;
 use PhpParser\Node;
-use PhpParser\Node\Expr;
 use PhpParser\Node\NullableType;
 use PhpParser\Node\Param as ParamNode;
 use PhpParser\Node\Stmt\Namespace_;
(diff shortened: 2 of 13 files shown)


src/NodeCompiler/Exception/UnableToCompileNode.php
```diff
@@ -53,9 +52,8 @@ public static function becauseOfNotFoundClassConstantReference(
     public static function becauseOfNotFoundConstantReference(
         CompilerContext $fetchContext,
         Node\Expr\ConstFetch $constantFetch,
+        string $constantName,
```
Ocramius, line 55: Just realized that this class is not internal - BC break, but acceptable, heh

src/Reflection/ReflectionParameter.php
```diff
@@ -161,46 +162,91 @@ public static function createFromNode(
...
+                try {
+                    $constantReflection = $this->reflector->reflectConstant($namespacedConstantName);
+
+                    $this->defaultValueConstantName = $constantReflection->getName();
+                    $this->defaultValue             = $constantReflection->getValue();
+                    $this->isDefaultValueParsed     = true;
+
+                    return;
+                } catch (IdentifierNotFound) {
+                    // Try constant name without namespace
```
Ocramius, line 202: Fallbacks, fallbacks everywhere!

src/Reflection/ReflectionParameter.php
```diff
@@ -161,46 +162,91 @@ public static function createFromNode(
...
+                }
+            }
+
+            $constantReflection = $this->reflector->reflectConstant($constantName);
+
+            $this->defaultValueConstantName = $constantReflection->getName();
+            $this->defaultValue             = $constantReflection->getValue();
+            $this->isDefaultValueParsed     = true;
```
Ocramius, line 210: Not a fan, but it's a pragmatic approach to avoid introducing yet another value object.

Perhaps `private ?ParsedConstantValue $defaultValue = null;`, but no need to refactor now.

src/Reflection/ReflectionParameter.php
```diff
@@ -247,6 +293,15 @@ public function getDeclaringClass(): ?ReflectionClass
         return null;
     }
 
+    public function getImplementingClass(): ?ReflectionClass
```
Ocramius, line 296: Is it required to have `getImplementingClass()` on a `ReflectionParameter`? :thinking:

Ocramius (changes requested):

src/Reflection/ReflectionParameter.php
kukulich: It’s just a shortcut. We already have getDeclaringClass()

src/Reflection/ReflectionParameter.php
kukulich: Will probably do in another PR because it can make one thing more nice.

Ocramius: Alright, rolling with this as-is then :+1:

Ocramius (approved):
Thanks @kukulich!

PR #798, opened 2021-10-10
It's possible to filter PHPStorm stubs by PHP version

kukulich:
Implemented https://github.com/Roave/BetterReflection/issues/603


asgrim (approved):

asgrim:
Neat idea, falls back to "yes function|class is supported if there is no `@since` tag etc" which is a sensible default. :shipit:

PR #801, opened 2021-10-11
Improved "deprecated" support to work with internal classes and PHP version constraints

kukulich:
The "deprecated" support is provided by the `@deprecated` annotation. There was already support for it, so I've just extend it.

There's just one disadvantage - the doccomment is not the same that is in the stubs but I don't think it's an issue for internal stubs.

---

There's another possible solution. The support can be provided by special BR `Deprecated` attribute that will be added in `ReflectionSourceStubber` or `PhpStormSourceStubber`.

Advantages:
- The doccomment will be untouched
- The attribute can be easily filtered in future `getAttributes()` methods so we can ignore it where we want.
- It allows to support even deprecated parameters: https://github.com/JetBrains/phpstorm-stubs/blob/master/Core/Core.php#L343

Waiting for ideas/thoughts :)

### DIFF
--- a/src/Reflection/Deprecated/DeprecatedHelper.php
+++ b/src/Reflection/Deprecated/DeprecatedHelper.php
@@ -13,16 +13,19 @@ final class DeprecatedHelper
 {
     private static ?DocBlockFactory $docBlockFactory = null;
 
+    /**
+     * @psalm-pure
+     */
     public static function isDeprecated(string $docComment): bool
     {
         if ($docComment === '') {
             return false;
         }
 
+        /** @psalm-suppress ImpureStaticProperty, ImpureMethodCall */
         self::$docBlockFactory ??= DocBlockFactory::createInstance();
 
-        $docBlock = self::$docBlockFactory->create($docComment);
-
-        return $docBlock->hasTag('deprecated');
+        /** @psalm-suppress ImpureStaticProperty, ImpureMethodCall */
+        return self::$docBlockFactory->create($docComment)->hasTag('deprecated');
     }
 }
--- a/src/Reflection/Deprecated/DeprecatedHelper.php
+++ b/src/Reflection/Deprecated/DeprecatedHelper.php
@@ -1,0 +1,28 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Roave\BetterReflection\Reflection\Deprecated;
+
+use phpDocumentor\Reflection\DocBlockFactory;
+
+/**
+ * @internal
+ */
+final class DeprecatedHelper
+{
+    private static ?DocBlockFactory $docBlockFactory = null;
+
+    public static function isDeprecated(string $docComment): bool
+    {
+        if ($docComment === '') {
+            return false;
+        }
+
+        self::$docBlockFactory ??= DocBlockFactory::createInstance();
+
+        $docBlock = self::$docBlockFactory->create($docComment);
+
+        return $docBlock->hasTag('deprecated');
+    }
+}
--- a/src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
+++ b/src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
@@ -8,6 +8,7 @@ use Generator;
 use JetBrains\PHPStormStub\PhpStormStubsMap;
 use PhpParser\BuilderFactory;
 use PhpParser\BuilderHelpers;
+use PhpParser\Comment\Doc;
 use PhpParser\Node;
 use PhpParser\NodeTraverser;
 use PhpParser\NodeVisitor\NameResolver;
@@ -33,6 +34,7 @@ use function in_array;
 use function is_dir;
 use function is_string;
 use function preg_match;
+use function preg_replace;
 use function sprintf;
 use function str_replace;
 use function strtolower;
@@ -232,7 +234,7 @@ final class PhpStormStubsSourceStubber implements SourceStubber
                 continue;
             }
 
-            $classNode->stmts = $this->filterStmtsByPhpVersion($classNode->stmts);
+            $classNode->stmts = $this->modifyStmtsByPhpVersion($classNode->stmts);
 
             $this->classNodes[strtolower($className)] = $classNode;
         }
@@ -272,6 +274,8 @@ final class PhpStormStubsSourceStubber implements SourceStubber
     private function createStub(Node $node): string
     {
         if (! ($node instanceof Node\Expr\FuncCall)) {
+            $this->addDeprecatedDocComment($node);
+
             $nodeWithNamespaceName = $node instanceof Node\Stmt\Const_ ? $node->consts[0] : $node;
 
             $namespaceBuilder = $this->builderFactory->namespace($nodeWithNamespaceName->namespacedName->slice(0, -1));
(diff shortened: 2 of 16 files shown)


src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
```diff
@@ -434,20 +438,80 @@ private function getAbsoluteFilePath(string $filePath): string
...
+        if ($docComment === null) {
+            $docCommentText = '/** @deprecated */';
+        } else {
+            $docCommentText = preg_replace('~(\r?\n\s*)\*/~', '\1* @deprecated\1*/', $docComment->getText());
+        }
+
+        $node->setDocComment(new Doc($docCommentText));
+    }
+
+    private function isDeprecatedInPhpVersion(Node\Stmt\ClassLike|Node\Stmt\ClassConst|Node\Stmt\Property|Node\Stmt\ClassMethod|Node\Stmt\Function_ $node): bool
```
Ocramius, line 484: Elaborate :O

src/SourceLocator/SourceStubber/ReflectionSourceStubber.php
```diff
@@ -188,11 +188,21 @@ private function addDocComment(
         Class_|Interface_|Trait_|Method|Property|Function_ $node,
         CoreReflectionClass|CoreReflectionMethod|CoreReflectionProperty|CoreReflectionFunction $reflection,
     ): void {
-        if ($reflection->getDocComment() === false) {
+        if (
+            ($reflection instanceof CoreReflectionMethod || $reflection instanceof CoreReflectionFunction)
```
Ocramius, line 192: Is deprecation not existing at class level?

src/SourceLocator/SourceStubber/ReflectionSourceStubber.php
```diff
@@ -188,11 +188,21 @@ private function addDocComment(
...
+            && $reflection->isInternal()
+            && $reflection->isDeprecated()
+        ) {
+            $docComment = '/** @deprecated */';
+        } else {
+            $docComment = $reflection->getDocComment();
+        }
+
+        if ($docComment === false) {
```
Ocramius, line 201: Are we 100% sure that `$reflection->getDocComment()` never returns anything for core methods?

src/SourceLocator/SourceStubber/ReflectionSourceStubber.php
Ocramius: Properties also seem to be skipped here :thinking:

test/unit/Reflection/ReflectionConstantTest.php
```diff
@@ -295,4 +297,41 @@ public function testGetAstByDefine(): void
         self::assertInstanceOf(Node\Expr\FuncCall::class, $ast);
         self::assertSame('FOO', $ast->args[0]->value->value);
     }
+
+    public function deprecatedDocCommentProvider(): array
+    {
+        return [
+            [
+                '/** 
+                  * @deprecated since 8.0
```
Ocramius, line 310: I didn't even know this was a thing :O

Ocramius (changes requested):

src/SourceLocator/SourceStubber/ReflectionSourceStubber.php
kukulich: Yes, `isDeprecated()` method exists only for functions/methods in native reflection.

src/SourceLocator/SourceStubber/ReflectionSourceStubber.php
kukulich: Please see the description of this PR. There's another option :)

However I can do here the same as I did in PhpStormSourceStubber - I can modify the comment when exists.

Ocramius:
> here's just one disadvantage - the doccomment is not the same that is in the stubs but I don't think it's an issue for internal stubs.

Not a big deal: internal stubs are not really in use by anyone so far, AFAIK.

If we need parity on internal methods too, we can improve/iterate on that

src/SourceLocator/SourceStubber/ReflectionSourceStubber.php
Ocramius: > I can modify the comment when exists.

Would be fine by me, but IMO overkill.

This is already a good improvement, so we should roll with it for now

kukulich:
@Ocramius Is it something I should fix here? I'm not able to find anything...

src/Reflection/Deprecated/DeprecatedHelper.php
```diff
@@ -0,0 +1,28 @@
...
+use phpDocumentor\Reflection\DocBlockFactory;
+
+/**
+ * @internal
+ */
+final class DeprecatedHelper
+{
+    private static ?DocBlockFactory $docBlockFactory = null;
+
+    public static function isDeprecated(string $docComment): bool
```
asgrim, line 19: Hm - would be nice if we can make this `@psalm-pure` if it is `static`, although I'm not sure if `DocBlockFactory::createInstance` is pure itself

kukulich: I can remove `DocBlockFactory` completely. It can be easy regexp I suppose...

Ocramius: We can mark it `@psalm-pure` and then use an inline suppression after inspection of `DocblockFactory`.

At first glance, `DocblockFactory` seems quite safe to use, when isolated this way: https://github.com/phpDocumentor/ReflectionDocBlock/blob/0005eb9eaecc2a3a00b8ee34c06643a316ebb228/src/DocBlockFactory.php

kukulich: So should I suppress everything?

```
ERROR: ImpureStaticProperty - src/Reflection/Deprecated/DeprecatedHelper.php:25:9 - Cannot use a static property in a mutation-free context (see https://psalm.dev/221)
        self::$docBlockFactory ??= DocBlockFactory::createInstance();


ERROR: ImpureMethodCall - src/Reflection/Deprecated/DeprecatedHelper.php:25:53 - Cannot call an impure method from a pure context (see https://psalm.dev/203)
        self::$docBlockFactory ??= DocBlockFactory::createInstance();


ERROR: ImpureStaticProperty - src/Reflection/Deprecated/DeprecatedHelper.php:27:21 - Cannot use a static property in a mutation-free context (see https://psalm.dev/221)
        $docBlock = self::$docBlockFactory->create($docComment);


ERROR: ImpureMethodCall - src/Reflection/Deprecated/DeprecatedHelper.php:27:45 - Cannot call a non-mutation-free method phpDocumentor\Reflection\DocBlockFactory::create from a pure context (see https://psalm.dev/203)
        $docBlock = self::$docBlockFactory->create($docComment);


ERROR: ImpureMethodCall - src/Reflection/Deprecated/DeprecatedHelper.php:29:27 - Cannot call a non-mutation-free method phpDocumentor\Reflection\DocBlock::hasTag from a pure context (see https://psalm.dev/203)
        return $docBlock->hasTag('deprecated');
```

Ocramius: I'd say so - all of this is literally in-memory operations on a `string`

kukulich: https://github.com/Roave/BetterReflection/pull/801/commits/9221c327a3c47b4002a532078f632bd8903454b6

Ocramius:
Thanks @kukulich! :+1:

PR #802, opened 2021-10-11
`AutoloadSourceLocator` now supports class aliases

### LINKED ISSUE #616: AutoloadSourceLocator doesn't support class aliases
When an aliased class is already loaded in runtime, and AutoloadSourceLocator is asked for reflection via this alias, it won't be able to find the implementation in the reported filename (because the real class name is different):

```php
        if (class_exists($className, false) || interface_exists($className, false) || trait_exists($className, false)) {
            $filename = (new ReflectionClass($className))->getFileName();

            if (! is_string($filename)) {
                return null;
            }

            return $filename;
        }
```

I was able to implement it by giving also the real class name besides the filename, like this:

```php
		if (class_exists($className, false) || interface_exists($className, false) || trait_exists($className, false)) {
			$reflection = new ReflectionClass($className);
			$filename = $reflection->getFileName();

			if (!is_string($filename)) {
				return null;
			}

			if (!file_exists($filename)) {
				return null;
			}

			return [$filename, $reflection->getName()];
		}
```

```php
		$locateResult = $this->locateClassByName($identifier->getName());
		if ($locateResult === null) {
			return null;
		}
		[$potentiallyLocatedFile, $className] = $locateResult;

		return $this->findReflection($reflector, $potentiallyLocatedFile, new Identifier($className, $identifier->getType()));
```

But it's not possible when the locator extends AbstractSourceLocator.

kukulich:
Fixes https://github.com/Roave/BetterReflection/issues/616

### DIFF
--- a/src/SourceLocator/Located/AnonymousLocatedSource.php
+++ b/src/SourceLocator/Located/AnonymousLocatedSource.php
@@ -4,6 +4,9 @@ declare(strict_types=1);
 
 namespace Roave\BetterReflection\SourceLocator\Located;
 
+/**
+ * @internal
+ */
 class AnonymousLocatedSource extends LocatedSource
 {
     public function __construct(string $source, string $filename)
--- a/src/SourceLocator/Located/AnonymousLocatedSource.php
+++ b/src/SourceLocator/Located/AnonymousLocatedSource.php
@@ -1,0 +1,13 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Roave\BetterReflection\SourceLocator\Located;
+
+class AnonymousLocatedSource extends LocatedSource
+{
+    public function __construct(string $source, string $filename)
+    {
+        parent::__construct($source, null, $filename);
+    }
+}
--- a/src/SourceLocator/Located/LocatedSource.php
+++ b/src/SourceLocator/Located/LocatedSource.php
@@ -22,7 +22,7 @@ class LocatedSource
      * @throws InvalidArgumentException
      * @throws InvalidFileLocation
      */
-    public function __construct(private string $source, ?string $filename)
+    public function __construct(private string $source, private ?string $name, ?string $filename = null)
     {
         if ($filename !== null) {
             FileChecker::assertReadableFile($filename);
@@ -36,6 +36,11 @@ class LocatedSource
     public function getSource(): string
     {
         return $this->source;
+    }
+
+    public function getName(): ?string
+    {
+        return $this->name;
     }
 
     public function getFileName(): ?string
@@ -62,5 +67,10 @@ class LocatedSource
     public function isEvaled(): bool
     {
         return false;
+    }
+
+    public function getAliasName(): ?string
+    {
+        return null;
     }
 }
--- a/src/SourceLocator/Ast/Locator.php
+++ b/src/SourceLocator/Ast/Locator.php
@@ -44,6 +44,7 @@ class Locator
                 $identifier->getType(),
             ),
             $identifier,
+            $locatedSource->getName(),
         );
     }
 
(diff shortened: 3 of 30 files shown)


src/SourceLocator/Located/AutoloadLocatedSource.php
```diff
@@ -0,0 +1,21 @@
...
+ * @internal
+ */
+class AutoloadLocatedSource extends LocatedSource
+{
+    public function __construct(string $source, ?string $filename, private ?string $originalName)
+    {
+        parent::__construct($source, $filename);
+    }
+
+    public function getOriginalName(): ?string
```
Ocramius: We should probably document what `getOriginalName()` means

src/SourceLocator/Located/LocatedSource.php
```diff
@@ -63,4 +63,9 @@ public function isEvaled(): bool
     {
         return false;
     }
+
+    public function getOriginalName(): ?string
```
Ocramius: We should probably document what `getOriginalName()` means

src/SourceLocator/Located/LocatedSource.php
```diff
@@ -63,4 +63,9 @@ public function isEvaled(): bool
...
+    {
+        return null;
```
Ocramius, line 74: Is this supposed to be `null`? It's a bit unclear from a reader's perspective why the main `LocatedSource` (used in most places) would miss a `getOriginalName()` :thinking:

src/SourceLocator/Located/AutoloadLocatedSource.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Roave\BetterReflection\SourceLocator\Located;
+
+/**
+ * @internal
+ */
+class AutoloadLocatedSource extends LocatedSource
```
Ocramius: A class-level description of `AutoloadedLocatedSource` vs `LocatedSource` would go a long way :D

Ocramius (changes requested):

src/SourceLocator/Located/LocatedSource.php
kukulich: I can make it non-optional and rename it to eg. `getLocatedName()` ? It just means we will ignore `$identifier->getName()` here: https://github.com/Roave/BetterReflection/pull/802/files#diff-13cc5d496e14b9c4570202c9c30411d7fb5d57928344c388247f7b8754c4085dR84

Ocramius: IMO, if it is not an alias, `getName()` and `getOriginalName()` are supposed to be the same.

Ocramius: Perhaps we can make a `AliasLocatedSource {private $aliasName; private LocatedSource $actualSource; }` as a more semantically understandable construct: the main reason for my comments on this patch are around structural understanding/naming

asgrim (approved):

src/SourceLocator/Located/LocatedSource.php
kukulich: @Ocramius There's no `LocatedSource` interface so I suppose I cannot create `AliasLocatedSource {private $aliasName; private LocatedSource $actualSource; }` but I can probably create `AliasLocatedSource` instead of `AutoloadLocatedSource`. I'll try something.

kukulich:
@Ocramius New version forcepushed.

src/SourceLocator/Located/AnonymousLocatedSource.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Roave\BetterReflection\SourceLocator\Located;
+
+class AnonymousLocatedSource extends LocatedSource
```
Ocramius, line 10: Should still be `/** @internal */`

Ocramius (approved):
Looking much better now! Besides a missing `/** @internal */`, this can be :ship: -ed

src/SourceLocator/Located/AnonymousLocatedSource.php
kukulich: `LocatedSource` is `@internal` so I thought it's enough but done :) https://github.com/Roave/BetterReflection/pull/802/commits/b848b4f86857f5d7970707f6c4276ae27ee75647

Ocramius: All shall be noticed

lor-lord-of-the-rings

PR #803, opened 2021-10-12
Fixed `ReflectionParameter::isArray()` and `isCallable()` to handle union types

kukulich:
Improved version of https://github.com/Roave/BetterReflection/pull/792

Fixes #792

### DIFF
--- a/src/Reflection/ReflectionParameter.php
+++ b/src/Reflection/ReflectionParameter.php
@@ -30,7 +30,6 @@ use function in_array;
 use function is_array;
 use function is_object;
 use function is_string;
-use function preg_match;
 use function sprintf;
 use function strtolower;
 
@@ -409,7 +408,7 @@ class ReflectionParameter
      */
     public function isArray(): bool
     {
-        return preg_match('~^\??array$~i', (string) $this->getType()) === 1;
+        return $this->isType($this->getType(), 'array');
     }
 
     /**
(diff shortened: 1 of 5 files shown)


src/Reflection/ReflectionParameter.php
```diff
@@ -409,15 +408,53 @@ public function removeType(): void
...
+            foreach ($types as $type) {
+                if (strtolower($namedType->getName()) === $type) {
+                    return true;
+                }
+            }
+
+            return false;
+        };
+
+        if ($typeReflection === null) {
```
Ocramius: Minor optimization: can move this up before the closure declaration

src/Reflection/ReflectionParameter.php
```diff
@@ -409,15 +408,53 @@ public function removeType(): void
...
+            return false;
+        }
+
+        if ($typeReflection instanceof ReflectionUnionType) {
+            /** @var list<ReflectionNamedType> $unionTypes */
```
Ocramius: Shouldn't we have this declared on `ReflectionUnionType`, instead of having an `/** @var`?

src/Reflection/ReflectionParameter.php
```diff
@@ -409,15 +408,53 @@ public function removeType(): void
...
+            $unionTypes = $typeReflection->getTypes();
+
+            $allTypesAllowed = true;
+
+            foreach ($unionTypes as $unionType) {
+                if (! $isOneOfAllowedTypes($unionType, $type, 'null')) {
+                    $allTypesAllowed = false;
```
Ocramius: We should probably `return false` here

Ocramius (changes requested):

Ocramius (approved):
:ship: nice! :+1:

PR #804, opened 2021-10-12
Fixed `ReflectionParameter::allowsNull()` for union types

kukulich:


src/Reflection/ReflectionParameter.php
```diff
@@ -308,15 +307,18 @@ public function allowsNull(): bool
             return true;
         }
 
-        if ($this->node->type instanceof NullableType) {
-            return true;
-        }
+        $type = $this->getType();
```
Ocramius: Perhaps we can simplify this further:

```php
$type = $this->getType();

if (null === $type) {
    return true;
}

// ...
```

That removes a potentially pricey call, and removes `assert()`

Ocramius (approved):
Good simplification: up to you if I should merge as-is, or if the suggested minor improvement should be added first

src/Reflection/ReflectionParameter.php
kukulich: Wil do in 30 minutes.

kukulich: Done.

asgrim (approved):

PR #807, opened 2021-10-12
The resolving of constant name and constant value is now always done in `CompileNodeToValue`

kukulich:
The resolving of constant name and constant value is now always done in `CompileNodeToValue` - it means we don't need to resolve constant name for the first time in `ReflectionParameter` and for the second time in `CompileNodeToValue` again.

### DIFF
--- a/src/Reflection/ReflectionParameter.php
+++ b/src/Reflection/ReflectionParameter.php
@@ -152,20 +152,20 @@ class ReflectionParameter
     /**
      * @throws LogicException
      */
-    private function compileDefaultValueNode(): void
+    private function getCompiledDefaultValueNode(): CompiledValue
     {
         if (! $this->isDefaultValueAvailable()) {
             throw new LogicException('This parameter does not have a default value available');
         }
 
-        if ($this->compiledDefaultValue !== null) {
-            return;
+        if ($this->compiledDefaultValue === null) {
+            $this->compiledDefaultValue = (new CompileNodeToValue())->__invoke(
+                $this->node->default,
+                new CompilerContext($this->reflector, $this),
+            );
         }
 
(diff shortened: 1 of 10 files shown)


src/Reflection/ReflectionProperty.php
```diff
@@ -288,13 +291,17 @@ public function getDefaultValue(): string|int|float|bool|array|null
...
         }
 
-        return (new CompileNodeToValue())->__invoke(
-            $defaultValueNode,
-            new CompilerContext(
-                $this->reflector,
-                $this,
-            ),
-        );
+        if ($this->compiledDefaultValue === null) {
```
Ocramius, line 295: This simplification is much welcomed, and we cache some stuff too! :+1:

src/Reflection/ReflectionParameter.php
```diff
@@ -339,9 +248,10 @@ public function isDefaultValueAvailable(): bool
      */
     public function getDefaultValue(): string|int|float|bool|array|null
     {
-        $this->parseDefaultValueNode();
+        $this->compileDefaultValueNode();
 
-        return $this->defaultValue;
+        /** @psalm-suppress PossiblyNullPropertyFetch */
```
Ocramius: Could we adjust this to use an internal private getter method?

Ocramius (changes requested):
Overall looking very good! Improvements/simplifications totally worth it :) 

Some very minor feedback added

src/Reflection/ReflectionParameter.php
kukulich: https://github.com/Roave/BetterReflection/pull/807/commits/b8099f7077a5bea0cc0782e98fceba4d7ba1a3a0

Ocramius (approved):
LGTM :ship: 

Thanks @kukulich!

PR #808, opened 2021-10-12
Faster builds by avoiding `pcov` installation when not needed

kukulich:
Not sure about the infection but it looks it's a little faster.


Ocramius (approved):
LGTM 👍

Beware that I'm planning to rip out CI, and use a bit less home-grown stuff here in future

Ocramius:
Thanks @kukulich!

PR #810, opened 2021-10-13
Stricter/smaller types for some internal BetterReflection symbols

kukulich:


src/Reflection/ReflectionClass.php
```diff
@@ -68,7 +67,7 @@ class ReflectionClass implements Reflection
 
     private LocatedSource $locatedSource;
 
-    private ClassLikeNode $node;
```
Ocramius, line 71: Really like how we get rid of `ClassLikeNode` :+1:

Ocramius (approved):
LGTM :+1:

Ocramius:
Thanks @kukulich! Also, mutation score went up a bit :D

kukulich:
> Also, mutation score went up a bit

Yes, that was my intention :)

PR #812, opened 2021-10-13
PHP 8.1: `never` builtin type

kukulich:


Ocramius (approved):
Also here: simple/sweet change :ship:

PR #813, opened 2021-10-13
PHP 8.1: Implemented `ReflectionClass::isEnum()` and `ReflectionObject::isEnum()`

kukulich:


Ocramius (changes requested):
Overall, change is simple and looks good, but here, we should really have either a subtype or a dedicated type.

In fact, the original RFC also has reflection additions about this: https://wiki.php.net/rfc/enumerations#reflection

I think the simplistic approach of "slamming everything in a `ReflectionClass` is a bad fit for long-term maintenance here.

kukulich:
This is just really only the implementation of `isEnum` method. Other support for enums needs separate PR.

Ocramius:
Hmm, now that I look at it, I don't see `isEnum()` in the RFC? Am I missing something?

kukulich:
Not sure but it's there: https://github.com/php/php-src/blob/master/ext/reflection/php_reflection.c#L4794

I think I find it thanks to adapter test.

Ocramius:
Ooof, ugly, but OK then :shrug: 

We can indeed expand on ENUM inspection once we dig deeper into 8.1 support :+1:

Ocramius (approved):

Ocramius:
Thanks @kukulich!

PR #819, opened 2021-10-13
PHP 8.1: Implemented `new` in initializers

kukulich:


src/NodeCompiler/CompileNodeToValue.php
```diff
@@ -193,6 +199,22 @@ private function getClassConstantValue(Node\Expr\ClassConstFetch $node, ?string
...
+        /** @psalm-var class-string $className */
+        $className = $node->class->toString();
+
+        if (! class_exists($className)) {
+            throw Exception\UnableToCompileNode::becauseOfClassCannotBeLoaded($context, $node, $className);
+        }
+
+        $arguments = array_map(fn (Node\Arg $arg): mixed => $this($arg->value, $context)->value, $node->args);
+
+        return new $className(...$arguments);
```
Ocramius: This is very much a bummer: no way to avoid autoloading when inspecting default parameters with `new` in them :-\

Ocramius: In fact, I'm not sure if `BetterReflection` should stop operating here, and throw an exception :thinking:

Ocramius (changes requested):

src/NodeCompiler/CompileNodeToValue.php
kukulich: We can also add a method `isDefaultValueAnInitializer()` so everyone can call it as a check and then decide if call `getDefaultValue()`

Second method could be something like `getDefaultValueInitializer()` that could return some object with initializer name and arguments.

Ocramius: Will think about it - brain fried for today.

/cc @asgrim this breaks the initial premise of this library, for context.

ondrejmirtes: For PHPStan purposes I'd need the AST of the default value and that's sufficient to resolve the type of the value without triggering autoloading.

So the BR's Reflection* classes could offer the AST and the adapter could have getDefaultValue() with the risk of autoloading, which is neccesary for compatibility.

Ocramius: Seems viable. Another thing we could do is `getDefaultValue()` throwing an exception when/if the value would lead to autoloading, and having a dedicated method that allows initialization instead.

asgrim: Yeh, not sure why we need to initialise here, I'd really like to avoid it totally, since it undermines the whole "static analysis" thing. I'd rather ttake `$className` and run it through whatever configured Source Locators we can pass into here perhaps. I know it is added complexity, but it is complexity borne by these new language features, so is inherent I think

src/NodeCompiler/CompileNodeToValue.php
```diff
@@ -58,6 +58,10 @@ public function __invoke(Node $node, CompilerContext $context): CompiledValue
                 return $this->getClassConstantValue($node, $constantName, $context);
             }
 
+            if ($node instanceof Node\Expr\New_) {
+                throw Exception\UnableToCompileNode::becauseOfInitializer($context, $node);
```
Ocramius, line 62: Thanks for moving this to an exception for now :-)

src/NodeCompiler/CompileNodeToValue.php
kukulich: Now it throws exception. `getAst()` method is already there, so PHPStan should be ok too.

kukulich:
Ok, can I lower the MSI again? It looks it's already failing in master...

Ocramius:
@kukulich yes please - I was trying to see which mutants were introducing the problem, but nothing related to this specific patch.

Ocramius:
To me, it seems like coverage is ignoring `assert()` calls, perhaps we have assertions disabled (by accident) in `.ini` :shrug:

Ocramius (approved):
LGTM :+1:

kukulich:
I think the problem is that we use `assert` to make PHPStan/Psalm happy but there are places where the `assert` cannot never fail.

Ocramius:
@kukulich those should **in theory** be picked up by `roave/infection-static-analysis-plugin`, but since they are uncovered, the plugin never kicks in to squish those mutants.

Ocramius:
I see you're trying to improve this in #825 - meanwhile, merging here :)

PR #820, opened 2021-10-14
Removed some `@psalm` annotatons

kukulich:


src/Reflection/ReflectionProperty.php
```diff
@@ -468,9 +468,9 @@ public function removeType(): void
     }
 
     /**
-     * @throws ClassDoesNotExist
+     * @param class-string $className
      *
-     * @psalm-assert class-string $className
+     * @throws ClassDoesNotExist
```
asgrim, line 473: This was right; it is an assertion on a string that could be _any_ `string`, so this asserts it is of pseudo-subtype `class-string`

src/Reflection/ReflectionProperty.php
```diff
@@ -483,8 +483,6 @@ private function assertClassExist(string $className): void
      * @throws NoObjectProvided
      * @throws NotAnObject
      * @throws ObjectNotInstanceOfClass
-     *
-     * @psalm-assert object $object
```
asgrim: Why remove this assertion?

src/Reflection/ReflectionFunction.php
```diff
@@ -130,8 +130,6 @@ private function assertIsNoClosure(): void
 
     /**
      * @throws FunctionDoesNotExist
-     *
-     * @psalm-assert callable-string $functionName
```
asgrim, line 134: Why remove this assertion? It tells Psalm that the caller's value passed into `$functionName` is a `callable-string`?

src/SourceLocator/Type/Composer/Factory/MakeLocatorForComposerJsonAndInstalledJson.php
```diff
@@ -47,7 +47,7 @@ public function __invoke(string $installationPath, Locator $astLocator): SourceL
         }
 
         /**
-         * @psalm-var array{
+         * @var array{
```
asgrim: Hmm, does PhpStorm understand these complex array structures yet?

src/Reflection/ReflectionFunction.php
Ocramius, line 134: Unsure about this: is it just leading to redundant checks? 🤔

Ocramius (dismissed):

src/Reflection/ReflectionProperty.php
kukulich: It can be defined by `@var` if necessary but it looks `Psalm` and `PHPStan` are happy without the assert.

src/Reflection/ReflectionProperty.php
kukulich: It's useless. There's `is_object` check in the method.

src/Reflection/ReflectionProperty.php
asgrim: I don't think using `@var` would carry the type narrowing back to the caller though; that's the point of `@psalm-assert`, to let the caller know that the parameter they passed in is indeed of a specific type.

src/Reflection/ReflectionFunction.php
kukulich: No Psalm or PHPStan report without it so I think it's just useless...

asgrim: That doesn't mean it's useless, it just means no caller has tried to use `$functionName` as anything other than a `callable-string`

src/SourceLocator/Type/Composer/Factory/MakeLocatorForComposerJsonAndInstalledJson.php
kukulich: Array shapes are on many places already in `@var` annotations.

PHPStorm support: https://blog.jetbrains.com/phpstorm/2021/07/phpstorm-2021-2-release/#array_shapes_phpdoc_syntax_support

src/Reflection/ReflectionProperty.php
kukulich: I think `@var` or other annotations works the same way. It's already on some places:

https://github.com/Roave/BetterReflection/blob/5.0.x/src/Reflection/ReflectionClass.php#L197

src/Reflection/ReflectionFunction.php
asgrim: Have confirmed it carries the type back to the caller, so this is indeed redundant.

```
INFO: Trace - src/Reflection/ReflectionFunction.php:89:9 - $functionName: string (see https://psalm.dev/224)
        /** @psalm-trace $functionName */
        $functionName = $this->getName();


INFO: Trace - src/Reflection/ReflectionFunction.php:92:9 - $functionName: callable-string (see https://psalm.dev/224)
        /** @psalm-trace $functionName */
        $this->assertFunctionExist($functionName);
```

src/Reflection/ReflectionProperty.php
asgrim: This is just by chance, that we never call the function with anything other than a `class-string`, so the type has never been narrowed by the `psalm-assert`. This one is probably fine though, as a type error (such as passing a `string`) would be raised anyway:

```
ERROR: ArgumentTypeCoercion - src/Reflection/ReflectionProperty.php:394:37 - Argument 1 of Roave\BetterReflection\Reflection\ReflectionProperty::assertClassExist expects class-string, parent type string provided (see https://psalm.dev/193)
            $this->assertClassExist($implementingClassName);
```

src/Reflection/ReflectionProperty.php
asgrim: This one should probably actually stay - for some reason the caller's type is not narrowed (when I forced the type prior to the call with `/** @var mixed $object */`:

```
INFO: Trace - src/Reflection/ReflectionProperty.php:375:9 - $object: mixed (see https://psalm.dev/224)
        /** @psalm-trace $object */
        $instance = $this->assertObject($object);
```

With the `@psalm-assert` re-added, this changes that trace to:

```
INFO: Trace - src/Reflection/ReflectionProperty.php:375:9 - $object: object (see https://psalm.dev/224)
        /** @psalm-trace $object */
        $instance = $this->assertObject($object);
```

src/SourceLocator/Type/Composer/Factory/MakeLocatorForComposerJsonAndInstalledJson.php
asgrim: Yeah this isn't rendering properly yet for me (2021.2):

Screenshot from 2021-10-14 08-45-45

kukulich: https://github.com/Roave/BetterReflection/blob/5.0.x/src/SourceLocator/Type/Composer/Factory/MakeLocatorForComposerJson.php#L45

We already use it so I think it can be `@var` to be consistent?

src/Reflection/ReflectionProperty.php
kukulich: Returned back.

src/SourceLocator/Type/Composer/Factory/MakeLocatorForComposerJsonAndInstalledJson.php
asgrim: Yeh that doesn't render either, they should be `@psalm-var` until all typical use-cases are supported for plain `@var` IMO

Screenshot from 2021-10-14 09-00-56

src/Reflection/ReflectionProperty.php
asgrim: For completeness, the other alternative is to change the type definition to `object`, which would behave the same as that `assertClassExist` method; would simplify the implementation, but we'd have to check its not a BC break (as the exceptions may bubble up from caller locations; from the outside that would change from a `NoObjectProvided` to a `TypeError` for example, which would I think be a BC break.

src/SourceLocator/Type/Composer/Factory/MakeLocatorForComposerJsonAndInstalledJson.php
kukulich: https://github.com/Roave/BetterReflection/pull/820/commits/3f2b94746fba96ad41045a7f9c6abf666167bc19

asgrim (approved):

asgrim:
@kukulich Great discussions here, I've learnt some stuff, thank you! :)

src/Reflection/ReflectionProperty.php
kukulich: `mixed` comes from https://github.com/Roave/BetterReflection/blob/5.0.x/src/Reflection/Adapter/ReflectionProperty.php#L60

Can be probably modified the same way as in https://github.com/Roave/BetterReflection/blob/5.0.x/src/Reflection/Adapter/ReflectionProperty.php#L42

kukulich: No, it's not possible because `setValue()` can be called with one or two parameters.

asgrim: yeah, you'd have to check it's an `object` beforehand, I was just saying, that's an alternative approach (but indeed, probably slightly less desirable :+1:  it's fine as is, and will :shipit:  now

PR #822, opened 2021-10-14
Improved tentative return types from `ReflectionSourceStubber` to make sure return types are not declared when they are only "tentative" - see https://wiki.php.net/rfc/internal_method_return_types

kukulich:


Ocramius:
Is this the patch that is attempting to **expand** the docblock instead of replacing it?

kukulich:
Yes, this change is there too but it’s not the main change.

Ocramius:
@kukulich then I'm kinda missing what this is about :D 

"Improved X" is not really helping me much, heh - hoping for something that makes sense in release notes, and for my own understanding in future.

kukulich:
@Ocramius Oh, sorry :) Ok... I'm adding special annotation when method/function has tentative return type in native reflection. So when using `ReflectionSourceStubber` the tentative return types will be exact. It can be later reused for `PhpStormSourceStubber` when new release of PHPStorm stubs is released.

src/Reflection/ReflectionFunctionAbstract.php
```diff
@@ -349,21 +349,22 @@ public function getDocBlockReturnTypes(): array
...
+        return $this->createReturnType();
     }
 
     /**
      * Do we have a return type declaration
      */
     public function hasReturnType(): bool
     {
+        if ($this->hasTentativeReturnType()) {
+            return false;
```
Ocramius, line 365: If I understand it correctly, the stubs declare return types, but also tentative return types, and tentative return types are more indicative of actual state than the AST return types?

Ocramius (approved):
LGTM :+1:

PR #825, opened 2021-10-22
Enable zend.assertions=1 for code coverage

kukulich:


Ocramius:
Looks like disabled assertions was indeed the catch :D

Ocramius (approved):

Ocramius:
Thanks @kukulich!

PR #826, opened 2021-10-25
Updated dependencies

kukulich:


test/unit/Util/ConstantNodeCheckerTest.php
```diff
@@ -41,6 +41,24 @@ public function testInvalidArgumentsCount(): void
         ConstantNodeChecker::assertValidDefineFunctionCall($node);
     }
 
+    public function testInvalidWithVariadicPlaceholderAsName(): void
+    {
+        self::expectException(InvalidConstantNode::class);
```
Ocramius, line 46: Not important, but in future, put the assertions right before the method call that will throw

Ocramius (approved):
🚢

Ocramius:
Thanks @kukulich

PR #827, opened 2021-10-25
PHP 8.1: Enums

### LINKED ISSUE #789: Implement adapters for `ReflectionAttribute`, `ReflectionEnumPureCase`, `ReflectionEnumBackedCase` and `ReflectionEnumUnitcase` when supported by PHP
Ref: https://bugs.php.net/bug.php?id=81474

### LINKED ISSUE #806: PHP 8.1 support
- [x] Green build https://github.com/Roave/BetterReflection/pull/767
- [x] Readonly properties https://github.com/Roave/BetterReflection/pull/811
- [x] Final constants https://github.com/Roave/BetterReflection/pull/814
- [x] `never` type https://github.com/Roave/BetterReflection/pull/812
- [x] `ReflectionClass::isEnum()` and `ReflectionObject::isEnum()` https://github.com/Roave/BetterReflection/pull/813
- [x] `ReflectionFunction::isStatic()` https://github.com/Roave/BetterReflection/pull/815
- [x] ~~`getClosureUsedVariables()`~~
- [x] Intersection types https://github.com/Roave/BetterReflection/pull/818
- [x] Tentative return types https://github.com/Roave/BetterReflection/pull/817
- [x] New in initializers https://github.com/Roave/BetterReflection/pull/819
- [x] Enums https://github.com/Roave/BetterReflection/pull/827

kukulich:
Fixes #789
Fixes #806

### DIFF
--- a/src/Reflection/Adapter/ReflectionClass.php
+++ b/src/Reflection/Adapter/ReflectionClass.php
@@ -12,6 +12,8 @@ use ReflectionMethod as CoreReflectionMethod;
 use Roave\BetterReflection\Reflection\ReflectionAttribute as BetterReflectionAttribute;
 use Roave\BetterReflection\Reflection\ReflectionClass as BetterReflectionClass;
 use Roave\BetterReflection\Reflection\ReflectionClassConstant as BetterReflectionClassConstant;
+use Roave\BetterReflection\Reflection\ReflectionEnum as BetterReflectionEnum;
+use Roave\BetterReflection\Reflection\ReflectionEnumCase as BetterReflectionEnumCase;
 use Roave\BetterReflection\Reflection\ReflectionMethod as BetterReflectionMethod;
 use Roave\BetterReflection\Reflection\ReflectionProperty as BetterReflectionProperty;
 use Roave\BetterReflection\Util\FileHelper;
@@ -20,13 +22,14 @@ use function array_combine;
 use function array_filter;
 use function array_map;
 use function array_values;
+use function constant;
 use function func_num_args;
 use function sprintf;
 use function strtolower;
 
 final class ReflectionClass extends CoreReflectionClass
 {
-    public function __construct(private BetterReflectionClass $betterReflectionClass)
+    public function __construct(private BetterReflectionClass|BetterReflectionEnum $betterReflectionClass)
     {
         unset($this->name);
     }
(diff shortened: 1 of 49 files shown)


src/Reflection/ReflectionEnum.php
```diff
@@ -0,0 +1,102 @@
...
+use Roave\BetterReflection\Reflector\Reflector;
+use Roave\BetterReflection\SourceLocator\Located\LocatedSource;
+
+use function array_combine;
+use function array_filter;
+use function array_key_exists;
+use function array_map;
+use function assert;
+
+class ReflectionEnum extends ReflectionClass
```
Ocramius, line 23: Overall, not a fan of having inheritance here for enums (core has that), but it's not a blocker for 5.x for now.

I'd expect this to probably be broken up in 6.x

src/Reflection/ReflectionEnum.php
```diff
@@ -0,0 +1,102 @@
...
+
+    public function isBacked(): bool
+    {
+        return $this->node->scalarType !== null;
+    }
+
+    public function getBackingType(): ?ReflectionNamedType
+    {
+        if ($this->node->scalarType === null) {
+            return null;
```
Ocramius: We should probably not expose a nullable API in our core here, or we should perhaps get rid of `isBacked()` and `hasCase()`.

src/Reflection/ReflectionEnum.php
```diff
@@ -0,0 +1,102 @@
...
+        }
+
+        $backingType = ReflectionNamedType::createFromTypeAndReflector($this->node->scalarType);
```
Ocramius: `createFromTypeAndReflector` is missing a reflector here

Ocramius (changes requested):

src/Reflection/ReflectionEnum.php
kukulich: https://github.com/Roave/BetterReflection/pull/831

src/Reflection/ReflectionEnum.php
kukulich: `hasCase`/`getCase` - it's same as `hasProperty`/`getProperty` - I tried to use the same pattern.

`getBackingType` - I modified it to the same pattern as `getDefaultValue()`. It now throws exception when enum is not backed.

kukulich:
@Ocramius @asgrim 

I'm not sure... Should I implement https://3v4l.org/dN6BB as part of `BetterReflectionClass` or only in adapter?

Ocramius:
> or only in adapter?

Looks like an adapter-only concern, since it causes autoloading :-\

kukulich:
@Ocramius I think this return value does not need autoload because `enum(Foo::BOO)` is just string. I'm just not sure if BR enum/class should return enum cases in list of constants.

Ocramius:
```
  ["BOO"]=>
  enum(Foo::BOO)
```

Is that really just a `string`? Looks like the engine is doing more there :thinking:

kukulich:
@Ocramius Oh, shit. You're right https://3v4l.org/fQiIg :(

kukulich:
Btw my question is for `getReflectionConstants()` as well.

Ocramius:
> @Ocramius Oh, shit. You're right https://3v4l.org/fQiIg :(

Yeah, that's kinda what I was referring to - it's actually instantiating the ENUM, so runtime evaluation is quite needed :-\ 

I think it would be potentially acceptable to have that happen in the **adapters**, but not in core (similar to what you did with `new in initializers` in #819 )

Ocramius:
I think besides dependabot, this is the last one we need to push through for `5.0.0` :O

kukulich:
@Ocramius This is already ready for review :) `getConstants()` for enum cases is implemented only in adapters

Ocramius:
@asgrim since I'm off until Wednesday (literally on a trip), can you perhaps help getting this through in the next days? 

Also: let's please not tag `5.0.0` until we tried it with `roave/backward-compatibility-check` (should be few minor changes)

asgrim:
> @asgrim since I'm off until Wednesday (literally on a trip), can you perhaps help getting this through in the next days?
> 
> Also: let's please not tag `5.0.0` until we tried it with `roave/backward-compatibility-check` (should be few minor changes)

I'm also off until Weds :sweat_smile: I'll see if I can get to it but no promises, quite a beast this one @kukulich :grin:

kukulich:
@asgrim There are more tests than the code changes 😃

src/Reflection/Adapter/ReflectionClass.php
```diff
@@ -202,6 +232,16 @@ private function filterBetterReflectionClassConstants(?int $filter): array
...
         }
 
+        if (
+            $this->betterReflectionClass instanceof BetterReflectionEnum
+            && (
+                $filter === null
+                || $filter & ReflectionClassConstant::IS_PUBLIC
+            )
+        ) {
+            $reflectionConstants += $this->betterReflectionClass->getCases();
```
Ocramius, line 242: Not a fan of `+=`, but we'll make it work :D

Ocramius (approved):
Absolutely impeccable! :O 

Impressive work, @kukulich, thank you!

PR #828, opened 2021-10-25
Refactor tests to expect exceptions closer to the throw site

kukulich:


Ocramius (approved):
LGTM, thanks for cleaning up all these tests!

PR #829, opened 2021-10-25
Fixed missing `Stringable` interface in the list retrieved by `ReflectionClass::getImmediateInterfaces()`

kukulich:


src/Reflection/ReflectionClass.php
```diff
@@ -1204,7 +1204,8 @@ public function getImmediateInterfaces(): array
 
         $nodes = $this->node instanceof InterfaceNode ? $this->node->extends : $this->node->implements;
 
-        return array_combine(
+        /** @var array<class-string, self> $interfaces */
```
Ocramius, line 1207: Hmm, is this docblock needed at all? :thinking:

kukulich:
```
ERROR: ArgumentTypeCoercion - src/Reflection/ReflectionClass.php:1218:46 - Argument 1 of Roave\BetterReflection\Reflection\ReflectionClass::addStringableInterface expects array<class-string, Roave\BetterReflection\Reflection\ReflectionClass>, parent type array<string, Roave\BetterReflection\Reflection\ReflectionClass> provided (see https://psalm.dev/193)
        return $this->addStringableInterface($interfaces);
```

Ocramius: Ah, so `$interfaceName->toString()` is not a `class-string` :-\ 

Never mind then

Ocramius (approved):
LGTM :+1:

PR #830, opened 2021-10-25
Fixed `Reflection*#__toString()` for methods/functions with a declared return type 

kukulich:
https://3v4l.org/sqBlD


Ocramius (approved):
🚢 

Also, these edge cases are getting a bit extreme - is it worth even fixing them? Is anyone relying on the `__toString()` output (other than us for testing/reflection compatibility purposes)?

kukulich:
Not sure but it's easy to do the fixes now so 🤷‍♂️

PR #835, opened 2021-11-05
Improved detection of supported functions in `PhpStormStubsSourceStubber` based on `JetBrains\PhpStorm\Internal\PhpStormStubsElementAvailable` attribute to decide supported PHP versions

kukulich:


Ocramius (approved):
LGTM - was going to ask for purpose of the change, but the test makes it very much clear :-) 

:ship:

PR #836, opened 2021-11-06
Added `ReflectionNamedType::getClass()`

kukulich:


src/Reflection/ReflectionClass.php
```diff
@@ -282,7 +283,25 @@ private function getParentMethods(): array
         return array_merge(
             [],
             ...array_map(
-                static fn (ReflectionClass $ancestor): array => $ancestor->getMethods(),
```
Ocramius, line 285: I'm a bit puzzled about this: why did `$ancestor->getMethods()` not suffice? We're kinda exploding and re-assembling everything here? :thinking:

src/Reflection/ReflectionFunctionAbstract.php
```diff
@@ -396,7 +396,9 @@ private function createReturnType(): ReflectionNamedType|ReflectionUnionType|Ref
             return null;
         }
 
-        return ReflectionType::createFromNode($returnType);
+        assert($this instanceof ReflectionMethod || $this instanceof ReflectionFunction);
```
Ocramius, line 399: `$this` is a `ReflectionFunctionAbstract`, which already has `$this->reflector`

Why is an `assert()` needed?

src/Reflection/ReflectionMethod.php
```diff
@@ -30,6 +30,8 @@ class ReflectionMethod extends ReflectionFunctionAbstract
 
     private ReflectionClass $implementingClass;
 
+    private ReflectionClass $currentClass;
```
Ocramius, line 33: I am a bit confused here: why are there 3 classes?

 * `$declaringClass` - obviously the topmost level
 * `$implementingClass` - potentially the level at which the concrete method is introduced (if no override occurred)
 * `$currentClass` - the class where we currently are, ignoring inheritance?

We could need a line or two explaining that :|

src/Reflection/ReflectionNamedType.php
```diff
@@ -54,6 +64,47 @@ public function isBuiltin(): bool
         return array_key_exists(strtolower($this->name), self::BUILT_IN_TYPES);
     }
 
+    public function getClass(): ReflectionClass
+    {
+        if (
+            ! $this->owner instanceof ReflectionEnum
```
Ocramius: Can some of these conditionals be inverted to lead to early returns? For example, the `$this->isBuiltIn()` can be moved up, and `return $this->reflector->reflectClass($this->name);` too, assuming we can make these a happy path branch?

src/Reflection/ReflectionParameter.php
```diff
@@ -487,43 +476,25 @@ private function getClassName(): ?string
...
-            assert($declaringClass instanceof ReflectionClass);
-            $parentClass = $declaringClass->getParentClass();
-            assert($parentClass instanceof ReflectionClass);
-
-            return $parentClass->getName();
-        }
-
-        if ($namedType->isBuiltin()) {
+        try {
+            return $namedType->getClass();
```
Ocramius, line 494: Fancy simplification! :+1:

Ocramius (approved):
Overall, excellent work as usual, @kukulich!

My comments are only around understanding (requiring some documentation additions), as well as potential refactoring for readability

Ocramius:
Also, thank you so much for helping out with the progress on https://github.com/Roave/BackwardCompatibilityCheck/pull/324 through this patch!

I'm much slower than you at jumping at all this stuff :D

src/Reflection/ReflectionFunctionAbstract.php
kukulich: I'm trying to use `ReflectionMethod|ReflectionFunction` as parameter type hint when possible and Psalm/PHPStan are not happy with `ReflectionFunctionAbstract` so I need to tell them it's `ReflectionMethod` or `ReflectionFunction`.

The best solution would be to get rid of `ReflectionFunctionAbstract` completly.

Ocramius: Ah, makes sense, thanks! :) 


 > The best solution would be to get rid of `ReflectionFunctionAbstract` completly.

Possibly long-term - IMO irrelevant for the `5.0.0` cut-off

kukulich: We'll see how fast I can prepar PR :D

Ocramius: Please don't overload yourself, especially with refactoring exercises that use your wrists more than your mind: we can gladly defer more BC breaks to a `6.0.0`, as there's nothing preventing that

src/Reflection/ReflectionClass.php
kukulich: I need to pass `$this` as `currentClass` to parent methods so I can resolve `static` in `ReflectionMethod`. Yeah, it's not perfect but same pattern is used for trait methods as well. There may be better way how to solve it but it would need more refactoring of `getMethods()` etc.

src/Reflection/ReflectionMethod.php
kukulich: I've now added some comments to the methods...

src/Reflection/ReflectionNamedType.php
kukulich: Thanks. I was not happy with the previous version. Now it's much more beutiful.

src/Reflection/ReflectionNamedType.php
```diff
@@ -54,6 +61,48 @@ public function isBuiltin(): bool
...
     }
 
+    public function getClass(): ReflectionClass
+    {
+        if (! $this->isBuiltin()) {
+            return $this->reflector->reflectClass($this->name);
+        }
+
+        if (
+            $this->owner instanceof ReflectionEnum
```
Ocramius, line 72: Shouldn't an enum have a reflectable type? 

After all, it is still a class-alike symbol: https://3v4l.org/dsl4L#v8.1rc3

src/Reflection/ReflectionNamedType.php
```diff
@@ -54,6 +61,48 @@ public function isBuiltin(): bool
...
+            || $this->owner instanceof ReflectionFunction
+            || ($this->owner instanceof ReflectionParameter && $this->owner->getDeclaringFunction() instanceof ReflectionFunction)
+        ) {
+            throw new LogicException('This type does not have class type');
```
Ocramius: We should add `$this->name` to the exception

kukulich: Modified to:

```throw new LogicException(sprintf('The type %s cannot be resolved to class', $this->name));```

File: src/Reflection/ReflectionNamedType.php (hunk 6 above)
Comment (kukulich) [reply src/Reflection/ReflectionNamedType.php:72]: The owner is `ReflectionMethod` in your example.

`ReflectionEnum` is owner here:

```php
enum Foo: int {}
```

Backing type can currently contain only `int` or `string`.

Ocramius (approved):
LGTM 🚢 

Thanks @kukulich!

PR #838, opened 2021-11-07
`ReflectionMethod` and `ReflectionFunction` no longer extend `ReflectionFunctionAbstract`

kukulich:
`ReflectionFunctionAbstract` is trait now

This change is necessary because `ReflectionFunctionAbstract`, when used as a type hint, is too generic, and may refer to multiple things.

In fact, we often have to `assert($function instanceof ReflectionMethod)` or similar: by using PHP 8 union types, we can be much more specific about which kind of function we want, and `ReflectionFunctionAbstract` can be avoided.

### DIFF
--- a/src/Reflection/ReflectionFunctionAbstract.php
+++ b/src/Reflection/ReflectionFunctionAbstract.php
@@ -177,7 +177,7 @@ trait ReflectionFunctionAbstract
 
     public function setDocCommentFromString(string $string): void
     {
-        $this->getAst()->setDocComment(new Doc($string));
+        $this->node->setDocComment(new Doc($string));
     }
 
     public function getFileName(): ?string
@@ -446,10 +446,7 @@ trait ReflectionFunctionAbstract
     /**
      * Fetch the AST for this method or function.
      */
-    public function getAst(): Node\Stmt\ClassMethod|Node\Stmt\Function_|Node\Expr\Closure|Node\Expr\ArrowFunction
-    {
-        return $this->node;
-    }
+    abstract public function getAst(): Node\Stmt\ClassMethod|Node\Stmt\Function_|Node\Expr\Closure|Node\Expr\ArrowFunction;
 
     /**
      * @return list<ReflectionAttribute>
--- a/src/Reflection/ReflectionFunctionAbstract.php
+++ b/src/Reflection/ReflectionFunctionAbstract.php
@@ -40,13 +40,11 @@ use function is_array;
 use function is_string;
 use function strtolower;
 
-abstract class ReflectionFunctionAbstract
+trait ReflectionFunctionAbstract
 {
-    public const CLOSURE_NAME = '{closure}';
-
     private static ?Parser $parser;
 
-    protected function __construct(
+    private function __construct(
         private Reflector $reflector,
         private Node\Stmt\ClassMethod|Node\Stmt\Function_|Node\Expr\Closure|Node\Expr\ArrowFunction $node,
         private LocatedSource $locatedSource,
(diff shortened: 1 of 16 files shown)


kukulich:
@Ocramius What do you think? It may be ugly but it was very easy and effective :D

Ocramius (approved):
WOW, it **did** turn out to be much simpler!

src/Reflection/ReflectionFunctionAbstract.php
```diff
@@ -41,10 +41,8 @@
 use function is_string;
 use function strtolower;
 
-abstract class ReflectionFunctionAbstract
+trait ReflectionFunctionAbstract
```
Ocramius, line 43: Can we enforce implementations here, at type level?

kukulich: Added two more commits.

kukulich:
Master:

```
3403 mutations were generated:
    2870 mutants were killed
      17 mutants were not covered by tests
     258 covered mutants were not detected
       8 errors were encountered
       3 syntax errors were encountered
       5 time outs were encountered
     242 mutants required more time than configured

Metrics:
         Mutation Score Indicator (MSI): 91%
         Mutation Code Coverage: 99%
         Covered Code MSI: 91%
```

PR:

```
3392 mutations were generated:
    2837 mutants were killed
      17 mutants were not covered by tests
     280 covered mutants were not detected
       8 errors were encountered
       3 syntax errors were encountered
       5 time outs were encountered
     242 mutants required more time than configured

Metrics:
         Mutation Score Indicator (MSI): 90%
         Mutation Code Coverage: 99%
         Covered Code MSI: 91%
```

I think it's good but I was probably too optimistic when setting 91% as required MSI.

src/Reflection/ReflectionFunctionAbstract.php
Ocramius: I was more wondering if we had a way to use `@psalm-require-extends A|B` or such :thinking:

kukulich:
Ok, with little cheeting the build is green!

src/Reflection/ReflectionFunctionAbstract.php
kukulich: I tried to add 

```
/**
 * @psalm-require-extends ReflectionMethod|ReflectionFunction
 */
 ```
 
 and Psalm does not report anything. However Psalm does not report anything even without the annotation... Should I add the annotation?

Comment (Ocramius) [reply src/Reflection/ReflectionFunctionAbstract.php:43]: I guess we can ignore it then 🤷‍♀️

The idea was to prevent the trait from being used elsewhere, but without declaring it internal.

Comment (kukulich) [reply src/Reflection/ReflectionFunctionAbstract.php:43]: No, my bad... In this branch it reports this:

```
ERROR: ExtensionRequirementViolation - src/Reflection/ReflectionFunction.php:25:9 - Roave\BetterReflection\Reflection\ReflectionFunctionAbstract requires using class to extend Roave\BetterReflection\Reflection\ReflectionFunction|Roave\BetterReflection\Reflection\ReflectionMethod, but Roave\BetterReflection\Reflection\ReflectionFunction does not (see https://psalm.dev/239)
    use ReflectionFunctionAbstract;


ERROR: ExtensionRequirementViolation - src/Reflection/ReflectionMethod.php:29:9 - Roave\BetterReflection\Reflection\ReflectionFunctionAbstract requires using class to extend Roave\BetterReflection\Reflection\ReflectionFunction|Roave\BetterReflection\Reflection\ReflectionMethod, but Roave\BetterReflection\Reflection\ReflectionMethod does not (see https://psalm.dev/239)
    use ReflectionFunctionAbstract;
```

DanielBadura:
Just here to thank you @kukulich you are doing here really impressive work! 🚀

asgrim:
> Just here to thank you @kukulich you are doing here really impressive work! rocket

I second this :fire:

src/Reflection/ReflectionFunctionAbstract.php
Ocramius: Right, let's roll with it for now - we may remove this trait in a much further placed release, if needed :)

Ocramius (approved):
:ship: awesome!

PR #840, opened 2021-11-08
Use native `assert()` instead of `Webmozart\Assert\Assert`, which is not an explicit dependency

kukulich:
Webmozart\Assert is not even among dependencies.


Ocramius (approved):

Ocramius:
Good catch!

PR #844, opened 2021-11-10
Fixed `AutoloadSourceLocatorTest::testCanAutoloadPsr4ClassesInPotentiallyMultipleDirectories()` autoloading of PHPUnit assertion classes

kukulich:


test/unit/SourceLocator/Type/AutoloadSourceLocatorTest.php
```diff
@@ -385,7 +385,25 @@ public function testReturnsNullWithInternalFunctions(): void
 
     public function testCanAutoloadPsr4ClassesInPotentiallyMultipleDirectories(): void
     {
-        spl_autoload_register([$this, 'autoload']);
+        // Force autoload for these asserts because it's not possible to autoload them in the autoload callback
```
Ocramius, line 388: This stuff should be done in `bootstrap.php`, no?

kukulich: Yes, the composer autoloader is in `bootstrap.php` but it does not work here ¯\_(ツ)_/¯  I tried `$prepend = true` for this autoload but it does not work either.

kukulich: To reproduce: `vendor/bin/phpunit test/unit/SourceLocator/Type/AutoloadSourceLocatorTest.php --filter=testCanAutoloadPsr4ClassesInPotentiallyMultipleDirectories`

Ocramius (approved):
LGTM 👍

PR #847, opened 2021-11-13
Improved detection if parameter is optional

kukulich:
No more use of dynamic properties on `PhpParser\Node\Param`

### DIFF
--- a/src/Reflection/ReflectionParameter.php
+++ b/src/Reflection/ReflectionParameter.php
@@ -41,6 +41,8 @@ class ReflectionParameter
 
     private int $parameterIndex;
 
+    private bool $isOptional;
+
     private ?CompiledValue $compiledDefaultValue = null;
 
     private Reflector $reflector;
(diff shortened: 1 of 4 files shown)


src/Reflection/ReflectionParameter.php
```diff
@@ -497,6 +500,31 @@ private function getClassFromNamedType(ReflectionNamedType $namedType): ?Reflect
...
+        if ($this->node->variadic) {
+            return true;
+        }
+
+        if ($this->node->default === null) {
+            return false;
+        }
+
+        foreach ($this->function->getAst()->getParams() as $otherParameterIndex => $otherParameterNode) {
+            if ($otherParameterIndex <= $this->parameterIndex) {
```
Ocramius, line 514: Is this already covered by our existing tests, perhaps? 🤔

Ocramius (approved):

src/Reflection/ReflectionParameter.php
kukulich: Yes, it is: https://github.com/Roave/BetterReflection/blob/5.0.x/test/unit/Reflection/ReflectionParameterTest.php#L488

Ocramius:
Thanks @kukulich!

PR #848, opened 2021-11-13
Less ignored errors in Psalm

kukulich:


src/Reflection/ReflectionParameter.php
```diff
@@ -497,6 +494,31 @@ private function getClassFromNamedType(ReflectionNamedType $namedType): ?Reflect
         }
     }
 
+    private function detectIsOptional(): bool
```
Ocramius: This seems to be improved here, but not really tested 🤔

Ocramius (approved):
Nice cleanup!

src/Reflection/ReflectionParameter.php
kukulich: Rebased :)

Ocramius: Indeed, solved in #847

Ocramius:
I did a skim of the escaped mutations, and **a lot** of them should probably really be caught by static analysis: weird :-\ 

It may be that our `psalm.xml` is still too lenient, but we'll improve it over time.

Ocramius:
Thanks @kukulich!

Ocramius:
Ok, indeed, a lot of the mutations that are left seem to be covered by nullability scenarios, specifically part of our remaining `psalm.xml` suppressions:

```xml
    <issueHandlers>
        <NullableReturnStatement errorLevel="suppress" />
        <InvalidNullableReturnType errorLevel="suppress" />

        <MoreSpecificReturnType errorLevel="suppress" />
        <LessSpecificReturnStatement errorLevel="suppress" />

        <PossiblyNullArgument errorLevel="suppress" />
    </issueHandlers>
```

Getting rid of most of them should get us a much better mutation testing profile too 💪

PR #851, opened 2021-11-14
Don't ignore Psalm `MoreSpecificReturnType` and `LessSpecificReturnStatement` errors anymore

kukulich:


Ocramius (approved):
LGTM 👍

src/Reflector/DefaultReflector.php
```diff
@@ -116,7 +116,7 @@ public function reflectConstant(string $identifierName): ReflectionConstant
      */
     public function reflectAllConstants(): iterable
     {
-        /** @var array<int,ReflectionConstant> $allConstants */
+        /** @var list<ReflectionConstant> $allConstants */
```
Ocramius, line 119: Note: while adding `/** @var` blocks is acceptable to work around type inference limitations, we should perhaps try to improve type inference.

In this case, `locateIdentifiersByType` could potentially be improved to return different types based on the given `IdentifierType`, but that would require an overhaul of the templated types of `IdentifierType` itself.

PR #853, opened 2021-11-14
Infection should skip initial tests because we already provide coverage

kukulich:


Ocramius (approved):
Noice, didn't see that we were re-running coverage hee.

PR #855, opened 2021-11-14
Split PhpStormStubs caching visitor to separate class

kukulich:


Ocramius (approved):
Nice cleanup - indeed, that class was becoming a bit unmanageable :D

PR #858, opened 2021-11-16
Removed dead condition

kukulich:
Forgotten in https://github.com/Roave/betterReflection/commit/e0509af4b0a4ed42840532261c9477d938a55d92


Ocramius (approved):

kukulich:
Btw `IdentifierType::isMatchingReflector()` is not used anywhere now so I suggest to remove it completly, even it's another BC break.

Ocramius:
Let's kill it as part of 5.0.0?

Ocramius:
Thanks @kukulich!

PR #860, opened 2021-11-16
Small code improvements

kukulich:


src/SourceLocator/SourceStubber/ReflectionSourceStubber.php
```diff
@@ -76,21 +76,20 @@ public function __construct()
...
             || interface_exists($className, false)
             || trait_exists($className, false)
-            || $isEnum
             )
         ) {
             return null;
         }
 
+        /** phpcs:disable SlevomatCodingStandard.Namespaces.ReferenceUsedNamesOnly.ReferenceViaFullyQualifiedName */
+        $isEnum = function_exists('enum_exists') && \enum_exists($className, false);
```
Ocramius, line 90: do enums pass `class_exists()` checks?

src/SourceLocator/SourceStubber/ReflectionSourceStubber.php
```diff
@@ -76,21 +76,20 @@ public function __construct()
...
             class_exists($className, false)
             || interface_exists($className, false)
             || trait_exists($className, false)
-            || $isEnum
             )
         ) {
             return null;
         }
 
+        /** phpcs:disable SlevomatCodingStandard.Namespaces.ReferenceUsedNamesOnly.ReferenceViaFullyQualifiedName */
```
Ocramius, line 89: I suppose importing the function won't work due to PHP 8.0 and static analysis?

Ocramius (changes requested):

src/SourceLocator/SourceStubber/ReflectionSourceStubber.php
kukulich: Yes. I just figured it out. 🤷‍♂️ 

Here I need preserve the check to choose `ReflectionEnum` or `ReflectionClass`.

src/SourceLocator/SourceStubber/ReflectionSourceStubber.php
kukulich: It works but PHPStan does not like `use function` for missing functions so this is hack for PHPStan. Consulted with @ondrejmirtes

Ocramius (approved):
LGTM :ship:

Ocramius:
Thanks @kukulich!

PR #874, opened 2021-11-23
Always require PHP version for `PhpStormStubsSourceStubber` - `PHP_VERSION_ID` as default

kukulich:
It's necessary for stubs like: https://github.com/JetBrains/phpstorm-stubs/blob/master/bcmath/bcmath.php#L198-L201

We have to know the version to choose the right parameter.

### DIFF
--- a/src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
+++ b/src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
@@ -33,6 +33,8 @@ use function sprintf;
 use function str_replace;
 use function strtolower;
 
+use const PHP_VERSION_ID;
+
 /**
  * @internal
  */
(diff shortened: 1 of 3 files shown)


src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
```diff
@@ -33,6 +33,8 @@
 use function str_replace;
 use function strtolower;
 
+use const PHP_VERSION_ID;
```
asgrim, line 36: If I understand correctly, this uses "the PHP version BR is running under", rather than "the PHP version of the code being analysed" as the default? What is the mechanism for overriding this? e.g. if we are using PHP 8.0 to analyse a codebase that runs on PHP 7.4?

kukulich: You can always specify the version by yourself: https://github.com/Roave/BetterReflection/pull/874/commits/419c1a72cbfaa72e0a02afd7c565c0c79470e0e5#diff-462651b2de3f83cc239bebc471d6800cebf5893e7258569982934b031c8c422eR169

kukulich:
Hmm, I don't know why the tests failed :(

kukulich:
Ok, tests failed because `enums` cannot be reflected - `UnitEnum` and `BackedEnum` stubs exist only in PHP 8.1.

Ocramius (approved):
LGTM - constructor used to define PHP version is also more than acceptable.

Ocramius:
Thanks for keeping an eye out for the upcoming stubs patch, @kukulich!

PR #880, opened 2021-11-28
Promoted property with a default value does not have a default value - the default value is on the parameter only

ondrejmirtes:
The desugaring is described in the RFC (https://wiki.php.net/rfc/constructor_promotion#desugaring):

> Notably, the property is declared without a default value (i.e. it starts out in an uninitialized state), and the default value is only specified on the constructor parameter. 

Runtime proof: https://3v4l.org/tu2AT

### DIFF
--- a/src/Reflection/ReflectionClass.php
+++ b/src/Reflection/ReflectionClass.php
@@ -765,7 +765,7 @@ class ReflectionClass implements Reflection
 
                     $propertyNode                     = new Node\Stmt\Property(
                         $parameterNode->flags,
-                        [new Node\Stmt\PropertyProperty($parameterNameNode->name, $parameterNode->default)],
+                        [new Node\Stmt\PropertyProperty($parameterNameNode->name)],
                         $parameterNode->getAttributes(),
                         $parameterNode->type,
                         $parameterNode->attrGroups,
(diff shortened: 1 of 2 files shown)


Ocramius:
Wait what?! O_O

Ocramius:
I think `BetterReflection` was doing the right thing here - perhaps to fix in adapters only?

ondrejmirtes:
I can see it being useful both ways.

1) To stay in the Better ReflectionClass - sometimes the distinction is significant. For example readonly property cannot have a default value, but if it's promoted, the constructor's argument can have a default value.

2) To move to the Adapter - To be able to tell "is this typed property ever in an uninitialized state", we'd have to keep the default value expression even for promoted properties.

Which one is more important is up to you, I'll update the PR if you want 😊

Ocramius:
Ah, wait, now I see what is going on:

 * the parameter **HAS** a default value
 * the property **DOES NOT** have a default value, but it's rather an assignment

Given this consideration, I think that this patch is good as-is, after rebase.

ondrejmirtes:
Rebased 🥳

asgrim (approved):

asgrim:
Thank you @ondrejmirtes :+1:

PR #886, opened 2021-11-30
Added some native parameter type hints

kukulich:


asgrim (approved):
I don't think any of these are BC breaks :+1:

Ocramius:
@asgrim note: `5.0.0` isn't released yet, so BC breaks are potentially fine

asgrim:
Thanks @kukulich

PR #896, opened 2021-12-03
We should add `Stringable` interface only when it can be reflected

kukulich:


src/Reflection/ReflectionClass.php
```diff
@@ -1112,7 +1112,12 @@ private function addStringableInterface(array $interfaces): array
 
         foreach ($this->node->getMethods() as $methodNode) {
             if ($methodNode->name->toLowerString() === '__tostring') {
-                $interfaces[$stringableClassName] = $this->reflectClassForNamedNode(new Node\Name(Stringable::class));
+                try {
+                    $interfaces[$stringableClassName] = $this->reflectClassForNamedNode(new Node\Name(Stringable::class));
+                } catch (IdentifierNotFound) {
+                    // Reflector does not know Stringable so we should not add it
```
Ocramius, line 1118: Ooof, this is fine, but it will go through all source locators :|

kukulich:
Shit. I've removed the branch by accident and now I cannot reopen the PR.

Ocramius:
Happens 🤷‍♀️

PR #898, opened 2021-12-03
We should add `Stringable` interface only when it can be reflected

kukulich:
I made a little ugly optimization.

### DIFF
--- a/src/Reflection/ReflectionClass.php
+++ b/src/Reflection/ReflectionClass.php
@@ -1112,7 +1112,16 @@ class ReflectionClass implements Reflection
 
         foreach ($this->node->getMethods() as $methodNode) {
             if ($methodNode->name->toLowerString() === '__tostring') {
-                $interfaces[$stringableClassName] = $this->reflectClassForNamedNode(new Node\Name(Stringable::class));
+                try {
+                    $stringableInterfaceReflection = $this->reflectClassForNamedNode(new Node\Name($stringableClassName));
+
+                    if ($stringableInterfaceReflection->isInternal()) {
+                        $interfaces[$stringableClassName] = $stringableInterfaceReflection;
+                    }
+                } catch (IdentifierNotFound) {
+                    // Stringable interface does not exist on target PHP version
+                }
+
                 break;
             }
         }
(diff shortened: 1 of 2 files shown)


src/Reflection/ReflectionClass.php
```diff
@@ -90,6 +90,8 @@ class ReflectionClass implements Reflection
 
     private ?ReflectionClass $cachedParentClass = null;
 
+    private static ReflectionClass|false|null $cachedStringableInterface = null;
```
Ocramius: A `static` is not viable here, because in projects like `roave/backwards-compatibility-check`, different/independent reflectors are used.

I don't have a solution for this: perhaps taking the performance impact, and optimizing reflector hit/miss operations (overall) could be a better way? 🤔

Ocramius (changes requested):

src/Reflection/ReflectionClass.php
kukulich: @Ocramius 

I've just found out that we already have `MemoizingSourceLocator`: https://github.com/Roave/BetterReflection/blob/5.0.x/src/SourceLocator/Type/MemoizingSourceLocator.php#L16

And it's default source locator for default reflector: https://github.com/Roave/BetterReflection/blob/5.0.x/src/BetterReflection.php#L46

src/Reflection/ReflectionClass.php
```diff
@@ -1112,7 +1112,16 @@ private function addStringableInterface(array $interfaces): array
...
         foreach ($this->node->getMethods() as $methodNode) {
             if ($methodNode->name->toLowerString() === '__tostring') {
-                $interfaces[$stringableClassName] = $this->reflectClassForNamedNode(new Node\Name(Stringable::class));
+                try {
+                    $stringableInterfaceReflection = $this->reflectClassForNamedNode(new Node\Name($stringableClassName));
+
+                    if ($stringableInterfaceReflection->isInternal()) {
+                        $interfaces[$stringableClassName] = $stringableInterfaceReflection;
+                    }
+                } catch (IdentifierNotFound) {
```
Ocramius, line 1121: Indeed, going back to this approach, while inefficient, is better than the static approach :|

Ocramius (approved):
LGTM :+1:

Ocramius:
Thanks @kukulich!

PR #900, opened 2021-12-04
Removed doc blocks parsing support

### LINKED ISSUE #899: Suggestion: Remove `TypesFinder`
- We did not improve the feature in last years
- There is support only for union types, no intersection types: https://github.com/Roave/BetterReflection/blob/5.0.x/src/TypesFinder/FindPropertyType.php#L57
- The library does not support generic or array shapes - we also have tests it does not work: https://github.com/Roave/BetterReflection/blob/5.0.x/test/unit/TypesFinder/FindPropertyTypeTest.php#L37
- Users can easily do it by themselves - methods `getAst()` and `getDocComment()` are public

So we don't support parsing of complex types and the support of parsing simple types is less useful with native types.

What do you think?

kukulich:
Fixes https://github.com/Roave/BetterReflection/issues/899


src/Reflection/Annotation/AnnotationHelper.php
```diff
@@ -13,38 +13,21 @@ final class AnnotationHelper
 {
     public const TENTATIVE_RETURN_TYPE_ANNOTATION = 'betterReflectionTentativeReturnType';
 
-    private static ?DocBlockFactory $docBlockFactory = null;
-
-    /**
```
Ocramius, line 18: Purity declarations should really just stay 🤔

src/Reflection/Annotation/AnnotationHelper.php
```diff
@@ -13,38 +13,21 @@ final class AnnotationHelper
...
             return false;
         }
 
-        /** @psalm-suppress ImpureMethodCall */
-        return self::getDocBlockFactory()->create($docComment)->hasTag('deprecated');
+        return preg_match('~\*\s*@deprecated(?=\s|\*)~', $docComment) === 1;
     }
 
-    /**
-     * @psalm-pure
```
Ocramius: Should remain

composer.lock
```diff
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "2c808292bc67bc7fa11a7bd66a10e71d",
+    "content-hash": "4a95e33fc49d25c30d4dacd6edfed638",
```
Ocramius: Wait... how did we **drop** 2 dependencies, yet `composer.lock` kept growing?! O_O

src/Reflection/Annotation/AnnotationHelper.php
```diff
@@ -13,38 +13,21 @@ final class AnnotationHelper
...
             return false;
         }
 
-        /** @psalm-suppress ImpureMethodCall */
-        return self::getDocBlockFactory()->create($docComment)->hasTag(self::TENTATIVE_RETURN_TYPE_ANNOTATION);
-    }
-
-    private static function getDocBlockFactory(): DocBlockFactory
-    {
-        self::$docBlockFactory ??= DocBlockFactory::createInstance();
```
Ocramius, line 46: As previously discussed in #899, I'm fully on board with dropping type parsing: what's the alternative these days, if we need to endorse it to end-users? 

Related: https://github.com/CuyZ/Valinor/issues/6
Related: https://github.com/phpstan/phpstan-src/tree/447cd102d946a2d9883ac82793195275f54296a9/src/Type
Related: https://github.com/vimeo/psalm/tree/f1d47cc662500589e200a978e3af85488a5236cf/src/Psalm/Type
Related: https://github.com/phpstan/phpdoc-parser
Related: https://github.com/phpstan/phpdoc-parser/blob/master/tests/PHPStan/Parser/TypeParserTest.php#L64

Overall, I'd like to at least explain (in the docs) that this library "ends" at providing docblocks, and that parsing them should be deferred to third-party libraries (since types are ever evolving there)

src/Reflection/ReflectionParameter.php
```diff
@@ -303,37 +296,6 @@ public function allowsNull(): bool
         return $type->allowsNull();
     }
 
-    /**
-     * Get the DocBlock type hints as an array of strings.
-     *
-     * @return list<string>
-     */
-    public function getDocBlockTypeStrings(): array
```
Ocramius, line 311: TIL: I use this in https://github.com/Roave/BackwardCompatibilityCheck/blob/11c7b6ff055c00cf9cebcd3f36995261de40385e/src/DetectChanges/BCBreak/PropertyBased/PropertyDocumentedTypeChanged.php#L42 :D 

I'll have to do some legwork to move on from there, I suppose...

Ocramius: (good for figuring out what the correct alternatives are)

Ocramius (changes requested):

composer.lock
kukulich: PHPUnit requires `phpspec/prophecy` and it requires `phpdocumentor/reflection-docblock` so the dependencies are still there... So the `composer.lock` is rearranged.

src/Reflection/ReflectionParameter.php
Ocramius: I removed this entire feature from `roave/backward-compatibility-check` ( https://github.com/Roave/BackwardCompatibilityCheck/pull/324/commits/f8c25b356ab752fc962faceae1f72e0566604480 )

src/Reflection/Annotation/AnnotationHelper.php
kukulich: Done.

src/Reflection/Annotation/AnnotationHelper.php
kukulich: Done.

Ocramius (approved):
Thanks @kukulich!

Ocramius:
Thanks @kukulich!

src/Reflection/Annotation/AnnotationHelper.php
ondrejmirtes: phpdoc-parser is really low-level, good comparison is nikic/PHP-Parser. In order to do something useful, you'd need parts of PHPStan that are hard to separate, like reflection layer, and type system.

ondrejmirtes: But I acknowledge that phpdoc-parser can sometimes be useful on its own, for example Slevomat CS uses it :) But it needs to interpret the AST on its own.

PR #902, opened 2021-12-05
Ensure that simple union types like `A|null` yield a nullable `ReflectionNamedType`

### LINKED ISSUE #901: Make `Foo|null` reflection consistent with php-src
Currently, when running reflection on an `A|null` type, BetterReflection produces a `Roave\BetterReflection\Reflection\ReflectionUnionType`: 

```php
var_dump(
    get_class(
        (new DefaultReflector(new StringSourceLocator(
            <<<'PHP'
<?php
interface A {}
final class AClass {
    private A|null $typed;
}
PHP
            ,
            (new BetterReflection())->astLocator()
        )))
            ->reflectClass('AClass')
            ->getProperty('typed')
            ->getType()
    )
);
```

produces

```
string(53) "Roave\BetterReflection\Reflection\ReflectionUnionType"
```

In PHP-SRC, this behavior is different: https://3v4l.org/gMA4T#v8.1rc3

```php
<?php

interface A {}
interface B {}

class Implementation
{
    function foo(A|null $param) { throw new Exception(); }
    function bar(A|B|null $param) { throw new Exception(); }
}

var_dump((new ReflectionParameter([Implementation::class, 'foo'], 0))->getType());
var_dump((new ReflectionParameter([Implementation::class, 'bar'], 0))->getType());
```

produces:

```
object(ReflectionNamedType)#2 (0) {
}
object(ReflectionUnionType)#1 (0) {
}
```

This means that a `UnionType` AST node composed of just `null` plus another type should be converted into a `ReflectionNamedType`, for the sake of compatibility with upstream.

This is ugly, but will (for now) avoid some bad issues in downstream handling (presently blocking https://github.com/Roave/BackwardCompatibilityCheck/pull/324 )

Ocramius:
Fixes #901

## Context

Currently, when running reflection on an `A|null` type, BetterReflection produces a `Roave\BetterReflection\Reflection\ReflectionUnionType`:

```php
var_dump(
    get_class(
        (new DefaultReflector(new StringSourceLocator(
            <<<'PHP'
<?php
interface A {}
final class AClass {
    private A|null $typed;
}
PHP
            ,
            (new BetterReflection())->astLocator()
        )))
            ->reflectClass('AClass')
            ->getProperty('typed')
            ->getType()
    )
);
```

produces

```
string(53) "Roave\BetterReflection\Reflection\ReflectionUnionType"
```

In PHP-SRC, this behavior is different: https://3v4l.org/gMA4T#v8.1rc3

```php
<?php

interface A {}
interface B {}

class Implementation
{
    function foo(A|null $param) { throw new Exception(); }
    function bar(A|B|null $param) { throw new Exception(); }
}

var_dump((new ReflectionParameter([Implementation::class, 'foo'], 0))->getType());
var_dump((new ReflectionParameter([Implementation::class, 'bar'], 0))->getType());
```

produces:

```
object(ReflectionNamedType)#2 (0) {
}
object(ReflectionUnionType)#1 (0) {
}
```

This means that a `UnionType` AST node composed of just `null` plus another type should be converted into
a `ReflectionNamedType`, for the sake of compatibility with upstream (this patch does that).

This is ugly, but will (for now) avoid some bad issues in downstream handling (presently blocking https://github.com/Roave/BackwardCompatibilityCheck/pull/324 )

### DIFF
--- a/src/Reflection/ReflectionType.php
+++ b/src/Reflection/ReflectionType.php
@@ -11,6 +11,10 @@ use PhpParser\Node\NullableType;
 use PhpParser\Node\UnionType;
 use Roave\BetterReflection\Reflector\Reflector;
 
+use function array_filter;
+use function array_values;
+use function count;
+
 abstract class ReflectionType
 {
     protected function __construct(
@@ -41,6 +45,15 @@ abstract class ReflectionType
 
         if ($type instanceof IntersectionType) {
             return new ReflectionIntersectionType($reflector, $owner, $type);
+        }
+
+        $nonNullTypes = array_values(array_filter(
+            $type->types,
+            static fn (Identifier|Name $type): bool => $type->toString() !== 'null',
+        ));
+
+        if (count($nonNullTypes) === 1) {
+            return self::createFromNode($reflector, $owner, $nonNullTypes[0], true);
         }
 
         return new ReflectionUnionType($reflector, $owner, $type, $allowsNull);
--- a/src/Reflection/ReflectionType.php
+++ b/src/Reflection/ReflectionType.php
@@ -11,16 +11,11 @@ use PhpParser\Node\NullableType;
 use PhpParser\Node\UnionType;
 use Roave\BetterReflection\Reflector\Reflector;
 
-use function array_filter;
-use function array_values;
-use function count;
-
 abstract class ReflectionType
 {
     protected function __construct(
         protected Reflector $reflector,
         protected ReflectionParameter|ReflectionMethod|ReflectionFunction|ReflectionEnum|ReflectionProperty $owner,
-        private bool $allowsNull,
     ) {
     }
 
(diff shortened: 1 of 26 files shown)


Ocramius:
This makes a few tests fail, because of obvious translation details:

```
There were 4 failures:

1) Roave\BetterReflectionTest\SourceLocator\SourceStubber\PhpStormStubsSourceStubberTest::testFunctionInPhpVersion with data set #15 ('dom_import_simplexml', 70000, true, 'DOMElement|null')
--- Expected
+++ Actual
@@ @@
-'DOMElement|null'
+'?DOMElement'

/home/runner/work/BetterReflection/BetterReflection/test/unit/SourceLocator/SourceStubber/PhpStormStubsSourceStubberTest.php:933

2) Roave\BetterReflectionTest\SourceLocator\SourceStubber\PhpStormStubsSourceStubberTest::testPropertyInPhpVersion with data set #4 ('PDOException', 'errorInfo', 80100, true, 'array|null')
PDOException::$errorInfo
--- Expected
+++ Actual
@@ @@
-'array|null'
+'?array'

/home/runner/work/BetterReflection/BetterReflection/test/unit/SourceLocator/SourceStubber/PhpStormStubsSourceStubberTest.php:889

3) Roave\BetterReflectionTest\SourceLocator\SourceStubber\PhpStormStubsSourceStubberTest::testPropertyInPhpVersion with data set #8 ('DOMNode', 'parentNode', 80100, true, 'DOMNode|null')
DOMNode::$parentNode
--- Expected
+++ Actual
@@ @@
-'DOMNode|null'
+'?DOMNode'

/home/runner/work/BetterReflection/BetterReflection/test/unit/SourceLocator/SourceStubber/PhpStormStubsSourceStubberTest.php:889

4) Roave\BetterReflectionTest\SourceLocator\SourceStubber\PhpStormStubsSourceStubberTest::testFunctionParameterInPhpVersion with data set #3 ('bcscale', 'scale', 80000, true, 'int|null', true)
bcscale::$scale
--- Expected
+++ Actual
@@ @@
-'int|null'
+'?int'
```

I fear this inconsistency between `php-src` pre-union-type and post-union-types will be problematic long-term...

Ocramius:
TBH, I really wish this was the complete opposite, and that nullable `ReflectionNamedType` didn't exist :|

Ocramius:
Ok, looking at this from the perspective of https://github.com/Roave/BackwardCompatibilityCheck/pull/324, and I think it would be simpler to never have a mix of union and nullable type: only ever have union types, then change this behavior in the `Adapter/` layer. 

@kukulich WDYT?

src/Reflection/ReflectionType.php
```diff
@@ -43,6 +47,15 @@ public static function createFromNode(
             return new ReflectionIntersectionType($reflector, $owner, $type);
         }
 
+        $nonNullTypes = array_values(array_filter(
```
Ocramius: This code is both inefficient and janky as heck: after some thinking, and after experimenting against https://github.com/Roave/BackwardCompatibilityCheck/pull/324, the whole `allowsNull()` is completely elided away, when `ReflectionUnionType::createFromNode()` produces `ReflectionUnionType` for nullable signatures :muscle:

kukulich:
@Ocramius Yes, I don't like `ReflectionNamedType` with `allowsNull` as well. I expect other tools like PHPStan has to deal with both variants anyway so it should be probably ok when BR returns only union types. Do I have right @ondrejmirtes ?

Ocramius:
I'm currently working on:

 * making this consistent in `roave/better-reflection` internals
 * making sure the `/Adapter` layer still operates as it would in php-src

Ocramius:
@kukulich I've now rewritten the patch:

 * `ReflectionUnionType` is kept in internals, when `?T` and `T|null` are met (no longer produces a `ReflectionNamedType`)
 * `ReflectionNamedType` is exposed at the adapter level, to make sure we stay compatible with the reflection API

kukulich:
LGTM.

Just one idea: Shouldn't we normalize/sort types in `ReflectionUnionType` to make sure we always have consistent output, eg. `null` as the last type of union? However, not probably important.

Ocramius:
@kukulich probably, but my batteries for this are depleted for today :D 

Core seems to be doing it: https://3v4l.org/jcq9S

Ocramius:
Thanks for the review, @kukulich!

drupol:
> LGTM.
> 
> Just one idea: Shouldn't we normalize/sort types in `ReflectionUnionType` to make sure we always have consistent output, eg. `null` as the last type of union? However, not probably important.

I know that phpcsfixer has a rule for that, here it is: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/doc/rules/phpdoc/phpdoc_types_order.rst

PR #905, opened 2021-12-06
Support `resource` as constant value, make `STDIN`, `STDOUT` and `STDERR` stubs actual `resource`s

kukulich:
I used a little hack. I converted the `define` calls to this:

 `define('STDIN', constant('STDIN'))`
 
And this format can be supported in stubs and then in both source stubbers.

### DIFF
--- a/src/NodeCompiler/CompileNodeToValue.php
+++ b/src/NodeCompiler/CompileNodeToValue.php
@@ -113,6 +113,17 @@ class CompileNodeToValue
                 return '';
             }
 
+            if (
+                $node instanceof Node\Expr\FuncCall
+                && $node->name instanceof Node\Name
+                && $node->name->toLowerString() === 'constant'
+                && $node->args[0] instanceof Node\Arg
+                && $node->args[0]->value instanceof Node\Scalar\String_
+                && defined($node->args[0]->value->value)
+            ) {
+                return constant($node->args[0]->value->value);
+            }
+
             throw Exception\UnableToCompileNode::forUnRecognizedExpressionInContext($node, $context);
         });
 
--- a/src/NodeCompiler/CompileNodeToValue.php
+++ b/src/NodeCompiler/CompileNodeToValue.php
@@ -51,7 +51,7 @@ class CompileNodeToValue
             $constantName = $this->resolveClassConstantName($node, $context);
         }
 
-        $constExprEvaluator = new ConstExprEvaluator(function (Node\Expr $node) use ($context, $constantName): string|int|float|bool|array|null {
+        $constExprEvaluator = new ConstExprEvaluator(function (Node\Expr $node) use ($context, $constantName): mixed {
             if ($node instanceof Node\Expr\ConstFetch) {
                 return $this->getConstantValue($node, $constantName, $context);
             }
--- a/src/SourceLocator/SourceStubber/PhpStormStubs/CachingVisitor.php
+++ b/src/SourceLocator/SourceStubber/PhpStormStubs/CachingVisitor.php
@@ -16,6 +16,7 @@ use function assert;
 use function constant;
 use function defined;
 use function in_array;
+use function is_resource;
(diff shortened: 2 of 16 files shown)


test/unit/NodeCompiler/CompileNodeToValueTest.php
```diff
@@ -201,6 +201,24 @@ public function testVariousNodeCompilations(string $phpCode, mixed $expectedValu
         self::assertSame($expectedValue, $compiledValue->value);
     }
 
+    public function testResource(): void
+    {
+        $node = $this->parseCode('STDIN');
+
+        $compiledValue = (new CompileNodeToValue())->__invoke($node, $this->getDummyContext());
+
+        self::assertIsResource($compiledValue->value);
```
Ocramius, line 210: Fancy solution :O

src/SourceLocator/SourceStubber/PhpStormStubs/CachingVisitor.php
```diff
@@ -86,18 +87,27 @@ public function enterNode(Node $node): ?int
...
-                return null;
-            }
-
             $argumentNameNode = $node->args[0];
             assert($argumentNameNode instanceof Node\Arg);
             $nameNode = $argumentNameNode->value;
             assert($nameNode instanceof Node\Scalar\String_);
             $constantName = $nameNode->value;
 
+            // Modify the value to constant('STDIN') call
```
Ocramius: This comment needs to be expanded a bit: it is clear to me that we're replacing `STDIN` with `constant('STDIN')`, so that we force BetterReflection to resolve that constant fetch, but that requires explanation for posterity too

src/NodeCompiler/CompileNodeToValue.php
```diff
@@ -113,6 +113,17 @@ public function __invoke(Node $node, CompilerContext $context): CompiledValue
                 return '';
             }
 
+            if (
+                $node instanceof Node\Expr\FuncCall
+                && $node->name instanceof Node\Name
+                && $node->name->toLowerString() === 'constant'
```
Ocramius, line 119: Perhaps a silly edge case, so feel free to ignore, but `constant()` may be imported from a namespace?

Ocramius (changes requested):

src/NodeCompiler/CompileNodeToValue.php
kukulich: It should return FQN: https://github.com/nikic/PHP-Parser/blob/master/lib/PhpParser/Node/Name.php#L113-L115

src/SourceLocator/SourceStubber/PhpStormStubs/CachingVisitor.php
kukulich: I've tried to improve it.

Ocramius (approved):
LGTM, thanks @kukulich!

Ocramius (changes requested):
A random CS issue appears!

```


FILE: ...src/SourceLocator/SourceStubber/PhpStormStubs/CachingVisitor.php
----------------------------------------------------------------------
FOUND 2 ERRORS AFFECTING 2 LINES
----------------------------------------------------------------------
 17 | ERROR | [x] Type define is not used in this file.
    |       |     (SlevomatCodingStandard.Namespaces.UnusedUses.UnusedUse)
 19 | ERROR | [x] Type fopen is not used in this file.
    |       |     (SlevomatCodingStandard.Namespaces.UnusedUses.UnusedUse)
----------------------------------------------------------------------
PHPCBF CAN FIX THE 2 MARKED SNIFF VIOLATIONS AUTOMATICALLY
----------------------------------------------------------------------
```

kukulich:
@Ocramius Already fixed and force pushed :)

Ocramius (approved):
Excellent, thanks! :ship:

PR #910, opened 2021-12-06
`PhpStormStubsSourceStubber` should add uses from current namespace to the stub

kukulich:
It enables the ability to resolve types in doc comments only with the given stub

### DIFF
--- a/src/SourceLocator/SourceStubber/PhpStormStubs/CachingVisitor.php
+++ b/src/SourceLocator/SourceStubber/PhpStormStubs/CachingVisitor.php
@@ -28,13 +28,15 @@ class CachingVisitor extends NodeVisitorAbstract
 {
     private const TRUE_FALSE_NULL = ['true', 'false', 'null'];
 
-    /** @var array<string, Node\Stmt\ClassLike> */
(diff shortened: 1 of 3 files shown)


src/SourceLocator/SourceStubber/PhpStormStubs/CachingVisitor.php
```diff
@@ -135,23 +143,35 @@ public function enterNode(Node $node): ?int
     }
 
     /**
-     * @return array<string, Node\Stmt\ClassLike>
+     * {@inheritDoc}
+     */
+    public function leaveNode(Node $node)
```
Ocramius, line 148: Should/can we declare `:void` here? :thinking:

Ocramius (approved):

src/SourceLocator/SourceStubber/PhpStormStubs/CachingVisitor.php
kukulich: https://github.com/nikic/PHP-Parser/blob/master/lib/PhpParser/NodeVisitor.php#L39-L58

So I think this is right.

Ocramius:
🚢  awesome!

PR #924, opened 2021-12-24
Simplified `continuous-integration.yml` with `ramsey/composer-install@v2`

kukulich:


Ocramius (approved):
LGTM :+1: 

@ramsey don't jinx my release >:D

PR #925, opened 2021-12-24
Updated locked dependencies

kukulich:


kukulich:
It looks there are already Xmas in GitHub and actions are failing randomly...

Ocramius:
I restarted the build

Ocramius:
Seems like one of ours tests gets killed due to OOM 🤔

kukulich:
It works on my Windows 🤷‍♂️

Ocramius:
`Killed` indicates that the process was killed by the parent process: we're probably hitting a memory limit (imposed by github actions)

kukulich:
@Ocramius is it possible to trigger build on 5.0 branch to test actual version?

Ocramius:
Trying on https://github.com/Roave/BetterReflection/actions/runs/1604155137

My suspicion is that something broke in the latest `8.1` release: let's see what a build from `5.0.x` says.

Ocramius:
One thing I'm noticing is that https://github.com/Roave/BetterReflection/runs/4626303892?check_suite_focus=true did not produce any readable output for multiple minutes...

Ocramius:
https://github.com/Roave/BetterReflection/runs/4626219388?check_suite_focus=true#step:8:23 indicates that github actions decided to SIGKILL the process, perhaps due to lack of output (and considering it "stuck")

Ocramius:
Maybe they changed some defaults at infrastructure level.

I think we may need to specify this: https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes

There's at least one test that hangs for a longer time, also on `5.0.x` - perhaps wasn't a problem a few days ago, because of more relaxed timeouts, but unsure :thinking: 

I'd say that a timeout of 30m is reasonable for our stuff, for now?

kukulich:
30 minutes should be enough but it looks the default is 360 minutes.

kukulich:
Btw there’s also new composer 2.2 in the broken builds.

Ocramius:
The problem is with the "no output for X minutes", which is something that travis-ci and gitlab have too: anything that doesn't produce output for a number of minutes is considered "dead" and killed.

We could revert to `composer:2.1` and see if that changes things? :thinking:

Ocramius:
Also: do we run `composer install` in our test suite? That could be it...

kukulich:
https://github.com/composer/composer/issues/10387

kukulich:
I’m only on mobile now so not able to try composer 2.1

Ocramius:
I can give it a shot later - need to finish some deadline-for-xmas work (every year, traditionally, businesses think that xmas is some sort of blocker)

Ocramius (approved):
Always feels good to get rid of `composer/package-versions-deprecated` :D 

Thanks @kukulich!

PR #926, opened 2021-12-24
Use composer 2.1 for tests temporarily

kukulich:


kukulich:
Created on my parents' computer :) It looks it works :)

Ocramius:
Dude, go do xmas stuff :D

Ocramius:
I'll lock to `composer:2.1` for the time being - will first inspect if/where we use `composer` in the test suite though

kukulich:
Ok, I'm not able to amend commits in GitHub codespaces so fun is over :)

Btw the problem is with PHPUnit process isolation (see https://github.com/composer/composer/commit/7d1ca1f56d81e7641e4c5b7716472c582e3841a6) so I've added composer 2.1 to all actions when it can be needed. I expect it's solution only for a couple of days.

kukulich:
Rebased with composer 2.1.

Ocramius (approved):
:ship: thanks @kukulich!

PR #929, opened 2021-12-25
Add ReflectionMethod::isMagic()

WinterSilence:


kukulich:
The check can be done easier: https://github.com/nikic/PHP-Parser/blob/f4b835f7d858d90b58c6d965a7d0f87d2ba57c43/lib/PhpParser/Node/Stmt/ClassMethod.php#L152

WinterSilence:
`DisallowedLateStaticBindingForConstant` ?! and how users will adding new features if owners not online long time?

Ocramius (changes requested):
I'm a bit dubious about this, since knowing that a method is "magic" is quite useless, on its own.

Besides that, please **write tests** when you implement something: it's not supposed to be an afterthought.

WinterSilence:
@Ocramius it's useful if you use package for static code analyze. packages like as php-cs-fixer use native PHP tokenizer - creating custom fixes for that's packages are real hardcore. It is much easier and faster to create similar fixes using AST / BetterReflection / ReflectionDocBlock.

Ocramius:
> it's useful if you use package for static code analyze

I don't see how: can't this be done on the caller-side?

When is a disambiguation between magic/non-magic method needed, compared to just checking if a method is a magic getter, magic setter, constructor, etc?

WinterSilence:
@Ocramius is* methods using as prefilter, for example, fast pre-filter  `isPublic() && !isStatic() && !isAbstract() && !isMagic()`  and after slow post-filter matches name by regexp

Ocramius:
Right, but why should it live **here** rather than in a caller?

WinterSilence:
@Ocramius why not if it's already defined in AST class? `isDestructor()` less usefull, but implemented.

Ocramius:
Mostly because the definition of "magic" is relative.

For example, `__construct` is considered a magic method, according to that definition, while it really isn't magic at all...

https://github.com/nikic/PHP-Parser/blob/f4b835f7d858d90b58c6d965a7d0f87d2ba57c43/lib/PhpParser/Node/Stmt/ClassMethod.php#L25-L26

WinterSilence:
@Ocramius
> it really isn't magic at all

constructor behavior not same to default function behavior - is maaagic (:

Ocramius (changes requested):
Decided to not roll with this: implementation/decision of "magicness" will remain in userland, not in this library.

WinterSilence:
@Ocramius please, add hint about this to documentation or class description.

PR #941, opened 2022-01-03
`mixed` `ReflectionNamedType` does allow `null` values

ondrejmirtes:
Matches PHP core behaviour: https://3v4l.org/FO5e4

### DIFF
--- a/src/Reflection/Adapter/ReflectionNamedType.php
+++ b/src/Reflection/Adapter/ReflectionNamedType.php
@@ -7,6 +7,8 @@ namespace Roave\BetterReflection\Reflection\Adapter;
 use ReflectionNamedType as CoreReflectionNamedType;
 use Roave\BetterReflection\Reflection\ReflectionNamedType as BetterReflectionNamedType;
 
+use function strtolower;
+
 /**
  * @psalm-suppress MissingImmutableAnnotation
  */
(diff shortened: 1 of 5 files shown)


Ocramius:
@ondrejmirtes could you rebase this onto `5.0.x`? I'd tag a patch release afterwards :+1:

ondrejmirtes:
Rebased :) And at least one more fix is coming, once I figure out something :)

src/Reflection/Adapter/ReflectionNamedType.php
```diff
@@ -23,7 +25,7 @@ public function getName(): string
 
     public function __toString(): string
     {
-        return ($this->allowsNull ? '?' : '')
+        return ($this->allowsNull && strtolower($this->betterReflectionType->getName()) !== 'mixed' ? '?' : '')
```
Ocramius, line 28: This is kinda missing a test, but meh :P

Ocramius (approved):
Thanks @ondrejmirtes!

src/Reflection/Adapter/ReflectionNamedType.php
ondrejmirtes: Some tests failed with not expecting `?mixed` without this change :)

Ocramius: All good then :D :muscle:

PR #942, opened 2022-01-03
Fix trait method visibility change leaking into other traits

ondrejmirtes:
… MemoizingSourceLocator

### DIFF
--- a/src/Reflection/ReflectionClass.php
+++ b/src/Reflection/ReflectionClass.php
@@ -244,6 +244,7 @@ class ReflectionClass implements Reflection
         $methodHash = $this->methodHash($method->getImplementingClass()->getName(), $method->getName());
 
         if (array_key_exists($methodHash, $traitModifiers)) {
+            $methodAst        = clone $methodAst;
             $methodAst->flags = ($methodAst->flags & ~ Node\Stmt\Class_::VISIBILITY_MODIFIER_MASK) | $traitModifiers[$methodHash];
         }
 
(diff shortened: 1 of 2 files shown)


Ocramius (changes requested):
Also here: `5.0.x` please :D

src/Reflection/ReflectionClass.php
```diff
@@ -244,6 +244,7 @@ private function createMethodsFromTrait(ReflectionMethod $method): array
         $methodHash = $this->methodHash($method->getImplementingClass()->getName(), $method->getName());
 
         if (array_key_exists($methodHash, $traitModifiers)) {
+            $methodAst        = clone $methodAst;
```
Ocramius, line 247: Oooooof, that's an ugly one :| 

I suppose AST should be marked `/** @psalm-immutable */` when possible, but too late for that...

test/unit/Reflection/ReflectionClassTest.php
```diff
@@ -2482,4 +2483,61 @@ public function testGetAttributesByInstance(): void
...
+            interface Foo
+            {
+                public function doFoo(): void;
+            }
+
+            class Bar implements Foo
+            {
+                public function doFoo(): void
+                {
+                    echo "Doing foo the generic way";
```
Ocramius: Is it possible to reduce this code even further? I suppose method body isn't that relevant, right?

ondrejmirtes: Yeah, it's irrelevant, I just copied it from the PHPStan bug report :) Gonna amend that.

Ocramius: Figured: just want to have the test as-on-point-as-possible, before merging 👍

ondrejmirtes: Done, verified it still fails without the `clone`.

Ocramius (approved):
LGTM, thanks @ondrejmirtes!

PR #945, opened 2022-01-04
Fix and update PHPStan to `1.3.x`

ondrejmirtes:
This PR solves the failures from https://github.com/Roave/BetterReflection/pull/939.

The problem here is quite tricky - PHPStan complains about some PHP classes that don't exist when running on PHP 8.1, because on PHP 8.0.x these really don't exist. PHPStan uses `config.platform.php` to determine the PHP version to test against.

At the same time it's funny that PHPStan doesn't complain about the same problems when running on PHP 8.0. That's because PHPStan itself loads these classes as stubs so that BetterReflection Adapter layer can work. And since PHPStan sees these classes as defined by the user, it considers them as existing.

I can't make PHPStan fail with the same errors when running on PHP 8.0, because I can't sensibly differentiate between these PHPStan-related stubs, and actual user-defined polyfills...

Here I fixed it with the same approach I use in phpstan-src - it tells PHPStan to ignore the setting from config.platform.php.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -9,11 +9,6 @@
         "nikic/php-parser": "^4.13.2",
         "roave/signature": "^1.5"
     },
-    "config": {
-        "platform": {
-            "php": "8.0.99"
-        }
-    },
     "authors": [
         {
             "name":     "James Titcumb",
--- a/composer.json
+++ b/composer.json
@@ -39,7 +39,7 @@
     ],
     "require-dev": {
         "doctrine/coding-standard": "^9.0.0",
-        "phpstan/phpstan": "^1.2.0",
+        "phpstan/phpstan": "^1.3.0",
         "phpunit/phpunit": "^9.5.11",
         "vimeo/psalm": "^4.17.0",
         "roave/infection-static-analysis-plugin": "^1.13.0"
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "facc95b7513df2926ab0f9c8253eccf8",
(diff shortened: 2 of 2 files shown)


test/phpstan-php-version.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+declare(strict_types = 1);
+
+return [
+    'parameters' => [
+        'phpVersion' => PHP_VERSION_ID,
```
Ocramius: We could also drop `config.platform.php` from `composer.json`: the reason it was added was to prevent invalid @dependabot upgrades, but we can get rid of it, if it has such an influence on other tools too...

Ocramius (changes requested):

test/phpstan-php-version.php
ondrejmirtes: Yeah, `"php": "~8.0.12 || ~8.1.0"` might be sufficient to keep valid dependencies.

ondrejmirtes: Interesting, the changed dropped `composer/package-versions-deprecated` and installed `ocramius/package-versions`, not sure why is that :) https://github.com/Roave/BetterReflection/pull/945/commits/8812e7bce75c6fdffa83b870299335669866a2a1

Ocramius: Massive improvement, I'd say 🤣

kukulich: Funny.

I've already removed `composer/package-versions-deprecated` in https://github.com/Roave/BetterReflection/commit/bae419ad6fff85c5767daf2839862bc27985bc3d

But it returned back in: https://github.com/Roave/BetterReflection/commit/24320f80c59846bacf7fc536276dab088cbe1eaf

Ocramius (approved):
Thanks @ondrejmirtes!

PR #947, opened 2022-01-04
Child class should have renamed trait method from parent

kukulich:
Should fix: https://phpstan.org/r/7d2d47f0-9bcf-4586-a3d7-f47e5741fc15


Ocramius (approved):

Ocramius:
Thanks @kukulich!

PR #956, opened 2022-01-05
Main branch is 5.1.x

kukulich:


infection.json.dist
```diff
@@ -8,7 +8,7 @@
     "logs": {
         "text": "php:\/\/stderr",
         "badge": {
-            "branch": "5.0.x"
+            "branch": "5.1.x"
```
Ocramius: We should use a regex here

.github/workflows/continuous-integration.yml
```diff
@@ -6,7 +6,7 @@ on:
   pull_request:
   push:
     branches:
-      - "5.0.x"
+      - "5.1.x"
```
Ocramius: Can we match all `\d+\.\d+\.x` branches?

Ocramius (changes requested):
Note: don't have a solution for automating the badge change yet...

Ocramius (approved):
LGTM! :+1:

Ocramius:
AH, never mind:

```
   Could not parse the JSON file "/home/runner/work/BetterReflection/BetterRef  
  lection/infection.json.dist": Parse error on line 11:                        
  ...          "branch": "/^\d\.\d\.x$/" 
```

JSON escaping requires a bit more work here :D

Ocramius:
Thanks @kukulich!

PR #959, opened 2022-01-08
Added `false` to types considered by `ReflectionNamedType#isBuiltIn()` logic

aurimasniekis:
`ReflectionNamedType::isBuiltIn()` was missing `false` type as built-in type.

Here it's proof with original reflections:

https://3v4l.org/OjBF8


Ocramius (approved):
LGTM, thanks @aurimasniekis!

PR #962, opened 2022-01-09
`static`, `self` and `parent` are not built-in in `ReflectionNamedType` adapter

### LINKED ISSUE #961: ReflectionNamedType::isBuiltIn incorrect behavior
I have noticed that [`ReflectionNamedType::isBuiltIn`](https://github.com/Roave/BetterReflection/blob/5.1.x/src/Reflection/ReflectionNamedType.php#L60) is working incorrectly, and it looks like it was broken after #750 merged.

Neither `self`, `parent`, `static` is built-in types in PHP according original PHP Reflections.

The last commit I can find which looked like it worked like should be this 3419b0f28b6d80f1af6bece384a125e5350bbb3d [`ReflectionNamedType::isBuiltIn`](https://github.com/Roave/BetterReflection/blob/3419b0f28b6d80f1af6bece384a125e5350bbb3d/src/Reflection/Adapter/ReflectionNamedType.php#L43)

I am not sure if removing them from the list is a good thing because I think those are used somewhere else. I also think maybe having a test to test the list of types against real Reflection to see if it's really is a built-in type would be nice to prevent this in the future.

kukulich:
Fixes #961

### DIFF
--- a/test/unit/Reflection/Adapter/ReflectionNamedTypeTest.php
+++ b/test/unit/Reflection/Adapter/ReflectionNamedTypeTest.php
@@ -4,11 +4,14 @@ declare(strict_types=1);
 
 namespace Roave\BetterReflectionTest\Reflection\Adapter;
 
+use PhpParser\Node;
 use PHPUnit\Framework\TestCase;
 use ReflectionClass as CoreReflectionClass;
 use ReflectionNamedType as CoreReflectionNamedType;
 use Roave\BetterReflection\Reflection\Adapter\ReflectionNamedType as ReflectionNamedTypeAdapter;
+use Roave\BetterReflection\Reflection\ReflectionMethod as BetterReflectionMethod;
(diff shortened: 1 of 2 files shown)


test/unit/Reflection/Adapter/ReflectionNamedTypeTest.php
```diff
@@ -86,4 +89,27 @@ public function testAdapterMethods(string $methodName, ?string $expectedExceptio
         $adapter = new ReflectionNamedTypeAdapter($reflectionStub, false);
         $adapter->{$methodName}(...$args);
     }
+
+    public function dataNotBuildin(): array
+    {
+        return [
+            ['self'],
+            ['static'],
+            ['parent'],
```
Ocramius, line 100: We should probably also test `PaReNt`, I suppose?

Ocramius (approved):

Ocramius:
Thanks @kukulich!

PR #965, opened 2022-01-10
Fixed located source for parent methods and methods from traits

### LINKED ISSUE #964: `ReflectionFunctionAbstract` has incorrect `locatedSource` value
When trying to get the location of the method which is inherited from the parent class fails with the `InvalidNodePosition` exception, but from a deeper look, it seems that the `$this->locatedSource->getSource()` method returns not the parent class source but the source of which implemented it.

From looking around the code seems, like it should return `$this->getDeclaringClass()->getLocatedSource()->getSource()` for `ReflectionMethod` and `$this->locatedSource->getSource()`  for `ReflectionFunction`

Reproducible code:
<details>
  <summary>Click to expand!</summary>

```php
<?php

declare(strict_types=1);

use Roave\BetterReflection\Reflection\ReflectionClass;
use Roave\BetterReflection\Reflection\ReflectionEnum;

require_once __DIR__ . '/vendor/autoload.php';

$classRef = ReflectionClass::createFromName(ReflectionEnum::class);
$methodRef = $classRef->getMethod('__toString');
$methodRef->getStartColumn();
```

</details>

kukulich:
Fixes #964

### DIFF
--- a/src/Reflection/ReflectionClass.php
+++ b/src/Reflection/ReflectionClass.php
@@ -251,7 +251,7 @@ class ReflectionClass implements Reflection
         $createMethod = fn (?string $aliasMethodName): ReflectionMethod => ReflectionMethod::createFromNode(
             $this->reflector,
             $methodAst,
-            $this->locatedSource,
+            $method->getLocatedSource(),
             $method->getDeclaringClass()->getDeclaringNamespaceAst(),
             $method->getDeclaringClass(),
             $this,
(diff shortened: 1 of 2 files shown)


Ocramius (approved):
Thanks @kukulich!

kukulich:
@Ocramius It looks you can release today too ;)

Ocramius:
Don't worry, I always have some PRs waiting for review :D 

Selection_878

PR #971, opened 2022-01-17
Fix missing properties for `UnitEnum` and `BackedEnum` interfaces

ondrejmirtes:
Phew, interfaces with properties...

This one actually fixes a couple of problems:

1) When reflecting UnitEnum and BackedEnum, which aren't actually enums, just plain interfaces, BackedEnum didn't know about the "name" property because ReflectionClass code didn't inherit properties from interfaces (understandably).
2) jetbrains/phpstorm-stubs 2021.3 have several bugs in enum stubs - the properties aren't readonly, and the BackedEnum value property has wrong type. Relevant PRs: https://github.com/JetBrains/phpstorm-stubs/pull/1316 + https://github.com/JetBrains/phpstorm-stubs/pull/1278

### DIFF
--- a/src/Reflection/ReflectionClass.php
+++ b/src/Reflection/ReflectionClass.php
@@ -804,11 +804,7 @@ class ReflectionClass implements Reflection
      */
     private function addEnumProperties(array $properties): array
     {
-        if (! $this->node instanceof EnumNode) {
-            return $properties;
-        }
-
-        $createProperty = function (string $name, string|Node\Identifier $type): ReflectionProperty {
+        $createProperty = function (string $name, string|Node\Identifier|Node\UnionType $type): ReflectionProperty {
             $propertyNode = new Node\Stmt\Property(
                 ClassNode::MODIFIER_PUBLIC | ClassNode::MODIFIER_READONLY,
                 [new Node\Stmt\PropertyProperty($name)],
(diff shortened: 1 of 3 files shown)


Ocramius:
May I add: what the fuck?

ondrejmirtes:
Original issue from here :) https://github.com/phpstan/phpstan/issues/6385

ondrejmirtes:
The CI fails because you need this commit :) https://github.com/ondrejmirtes/BetterReflection/commit/b5d9ca345c491cb125217f7880241ed4cb140628

It's about updating phpstorm-stubs so that in earlier PHP versions classes and interfaces don't implement/extend things from newer PHP versions.

I think I'm gonna leave that to @kukulich because it needs some more work :)

Ocramius:
@ondrejmirtes does #972 cover the missing details needed for this one?

Ocramius:
@kukulich some agreement needed for #972: should it land on `5.0.x` as a bugfix, or `5.1.x` as improvement?

I'm tempted to move this one to `5.1.x` and cut `5.1.0` once ready

ondrejmirtes:
AFAIK https://github.com/Roave/BetterReflection/pull/972 fixes PhpStormStubsSourceStubber which is currently buggy - you ask for some class reflection, ask for its parent class and it crashes because it extends something it doesn't exist yet. So it can land in 5.0.x.

kukulich:
I agree with @ondrejmirtes, eg. `SimpleXMLIterator` implements `Stringable` (PHP 8.0 only) so it cannot be reflected on PHP < 8.0, hence I think it's bug.

Ocramius:
If anyone can rebase this one, we can 🚢

ondrejmirtes:
Looks like it's going to be green :)

Ocramius (approved):
LGTM 👍

Ocramius:
Thanks @ondrejmirtes!

PR #972, opened 2022-01-17
Improved `PhpStormStubsSourceStubber` for PHP < 8.0

kukulich:
It's not nice... but there's no other option... Jetbrains may introduce some better solution in the future.

### DIFF
--- a/src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
+++ b/src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
@@ -4,8 +4,16 @@ declare(strict_types=1);
 
 namespace Roave\BetterReflection\SourceLocator\SourceStubber;
 
+use CompileError;
+use DatePeriod;
+use Error;
 use Generator;
+use Iterator;
+use IteratorAggregate;
 use JetBrains\PHPStormStub\PhpStormStubsMap;
+use JsonSerializable;
+use ParseError;
+use PDOStatement;
 use PhpParser\BuilderFactory;
 use PhpParser\BuilderHelpers;
 use PhpParser\Comment\Doc;
@@ -14,10 +22,13 @@ use PhpParser\NodeTraverser;
 use PhpParser\NodeVisitor\NameResolver;
 use PhpParser\Parser;
 use PhpParser\PrettyPrinter\Standard;
+use RecursiveIterator;
 use Roave\BetterReflection\Reflection\Annotation\AnnotationHelper;
 use Roave\BetterReflection\SourceLocator\FileChecker;
 use Roave\BetterReflection\SourceLocator\SourceStubber\Exception\CouldNotFindPhpStormStubs;
 use Roave\BetterReflection\SourceLocator\SourceStubber\PhpStormStubs\CachingVisitor;
+use SimpleXMLElement;
+use SplFixedArray;
 use Traversable;
 
 use function array_change_key_case;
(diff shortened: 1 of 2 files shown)


ondrejmirtes (approved):

src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
```diff
@@ -415,6 +445,57 @@ private function getAbsoluteFilePath(string $filePath): string
         return sprintf('%s/%s', $this->getStubsDirectory(), $filePath);
     }
 
+    /**
+     * @param list<Node\Name> $nameNodes
+     *
+     * @return list<Node\Name>
+     */
+    private function modifyExtendsImplementsByPhpVersion(string $className, array $nameNodes): array
```
Ocramius: This method does not modify any of the `$nameNodes`: its name is confusing, so we should probably call it something like `extendsOrImplementsReplacmentsByPhpVersion()` or such

src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
```diff
@@ -415,6 +445,57 @@ private function getAbsoluteFilePath(string $filePath): string
...
+    {
+        $modifiedNames = [];
+        foreach ($nameNodes as $nameNode) {
+            $name = $nameNode->toString();
+
+            if ($className === ParseError::class) {
```
Ocramius, line 461: Is there a way to write this in a more "linear" version? Most of these `if/else` blocks seem nested and related to each other :thinking: 

Perhaps `continue;` when it does **not** match any of these 3? :thinking:

src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
```diff
@@ -415,6 +445,57 @@ private function getAbsoluteFilePath(string $filePath): string
...
+                    continue;
+                }
+            } elseif (in_array($className, [DatePeriod::class, PDOStatement::class, true])) {
+                if ($name === IteratorAggregate::class && $this->phpVersion < 80000) {
+                    $modifiedNames[] = new Node\Name\FullyQualified(Traversable::class);
+                    continue;
+                }
+            }
+
+            $nodeData = $this->getClassNodeData($name);
```
Ocramius: Let's get rid of this variable

src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
```diff
@@ -415,6 +445,57 @@ private function getAbsoluteFilePath(string $filePath): string
...
+
+            if ($nodeData === null) {
+                continue;
+            }
+
+            $modifiedNames[] = $nameNode;
```
Ocramius, line 493: If I understand it correctly, except for `ParseError`, `CompileError`, `SplFixedArray`, `SimpleXMLElement`, `DatePeriod` and `PDOStatement`, `$modifiedNames === $namedNode`?

Can we somehow short-circuit that perhaps? Doesn't need to be, if the code gets even worse, but would be best to avoid

src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
```diff
@@ -415,6 +445,57 @@ private function getAbsoluteFilePath(string $filePath): string
...
+                }
+
+                if ($name === Iterator::class && $this->phpVersion >= 80000) {
+                    continue;
+                }
+            } elseif ($className === SimpleXMLElement::class) {
+                if ($name === RecursiveIterator::class && $this->phpVersion < 80000) {
+                    continue;
+                }
+            } elseif (in_array($className, [DatePeriod::class, PDOStatement::class, true])) {
```
Ocramius: IMO `===` in this case - removes `in_array()` and array instantiation (even from opcache interned values) each time

src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
```diff
@@ -415,6 +445,57 @@ private function getAbsoluteFilePath(string $filePath): string
         return sprintf('%s/%s', $this->getStubsDirectory(), $filePath);
     }
 
+    /**
+     * @param list<Node\Name> $nameNodes
```
Ocramius, line 451: Let's please add a paragraph explaining why we're doing this, and perhaps an issue ID, if it's something we can get rid of, with new PHPStorm stubs

Ocramius (changes requested):

src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
kukulich: There's no PHPStubs issue :(

src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
kukulich: No, there are some others that are fixed automatically by `if ($this->getClassNodeData($name) === null) {`, eg. `DOMDocument` implements `DOMParentNode` that is PHP 8+ only.

src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php

src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php

src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
kukulich: Added comment.

src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
kukulich: Changed. I hope it's ok.

src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
kukulich: I've really tried but I don't know how to improve it. The stubs are sometimes really garbage.

Ocramius: I totally trust you on that, thanks for putting ij the effort 👍

Ocramius (approved):
LGTM, thanks @kukulich and @ondrejmirtes!

Ocramius:
Ah, last thing: I suppose we need this on `5.0.x`?

kukulich:
Changed the base now. The code is already based on 5.0.x.

Ocramius:
Awesome, thanks!

PR #973, opened 2022-01-17
Improved mutation score

kukulich:


src/SourceLocator/Type/Composer/Psr/Psr0Mapping.php
```diff
@@ -54,7 +54,7 @@ public function resolvePossibleFilePaths(Identifier $identifier): array
         foreach ($this->mappings as $prefix => $paths) {
             if (strpos($className, $prefix) === 0) {
                 return array_map(
-                    static fn (string $path): string => rtrim($path, '/') . '/' . str_replace(['\\', '_'], '/', $className) . '.php',
```
Ocramius, line 57: The ending `/` in the path is valid in PSR-0 and PSR-4 mappings - should probably be tested.

src/SourceLocator/Type/Composer/Psr/Psr4Mapping.php
```diff
@@ -68,7 +68,7 @@ public function resolvePossibleFilePaths(Identifier $identifier): array
                     return [];
                 }
 
-                return array_map(static fn (string $path): string => rtrim($path, '/') . '/' . $subPath . '.php', $paths);
+                return array_map(static fn (string $path): string => $path . '/' . $subPath . '.php', $paths);
```
Ocramius, line 71: The ending `/` in the path is valid in PSR-0 and PSR-4 mappings - should probably be tested.

test/unit/Util/Autoload/ClassLoaderTest.php
```diff
@@ -72,19 +75,18 @@ public function testAddClassThrowsExceptionWhenClassAlreadyRegisteredInAutoload(
 
         $this->expectException(ClassAlreadyRegistered::class);
         $loader->addClass($reflection);
-
-        spl_autoload_unregister($loader);
```
Ocramius, line 76: Probably to be kept in the same process?

test/unit/Util/Autoload/ClassLoaderTest.php
```diff
@@ -72,19 +75,18 @@ public function testAddClassThrowsExceptionWhenClassAlreadyRegisteredInAutoload(
...
+     */
     public function testAddClassThrowsExceptionWhenClassAlreadyLoaded(): void
     {
         $loaderMethod = $this->createMock(LoaderMethodInterface::class);
         $loader       = new ClassLoader($loaderMethod);
 
         $this->expectException(ClassAlreadyLoaded::class);
         $loader->addClass(ReflectionClass::createFromName(stdClass::class));
-
-        spl_autoload_unregister($loader);
```
Ocramius, line 87: Probably to be kept in the same process?

test/unit/Util/Autoload/ClassLoaderTest.php
```diff
@@ -106,7 +108,5 @@ public function testAutoloadThrowsExceptionWhenClassIsNotLoadedCorrectlyAfterAtt
 
         $this->expectException(FailedToLoadClass::class);
         new AnotherTestClassForAutoloader();
-
-        spl_autoload_unregister($loader);
```
Ocramius, line 110: Probably to be kept?

test/unit/Util/Autoload/ClassLoaderTest.php
```diff
@@ -72,19 +75,18 @@ public function testAddClassThrowsExceptionWhenClassAlreadyRegisteredInAutoload(
 
         $this->expectException(ClassAlreadyRegistered::class);
         $loader->addClass($reflection);
-
-        spl_autoload_unregister($loader);
     }
 
+    /**
+     * @runInSeparateProcess
```
Ocramius: Probably to be kept in the same process?

test/unit/Util/Autoload/ClassLoaderTest.php
```diff
@@ -61,6 +61,9 @@ public function testAutoloadTriggersLoaderMethod(): void
         spl_autoload_unregister($loader);
     }
 
+    /**
+     * @runInSeparateProcess
```
Ocramius: Probably to be kept in the same process?

Ocramius: Note: `@runInSeparateProcess` is to be avoided, if possible: we almost never use global state, so we should probably avoid it whenever possible.

Ocramius (changes requested):

src/SourceLocator/Type/Composer/Psr/Psr0Mapping.php
kukulich: It's already removed in https://github.com/Roave/BetterReflection/blob/5.1.x/src/SourceLocator/Type/Composer/Psr/Psr0Mapping.php#L37 and tested as well.

src/SourceLocator/Type/Composer/Psr/Psr4Mapping.php
kukulich: It's already removed in https://github.com/Roave/BetterReflection/blob/5.1.x/src/SourceLocator/Type/Composer/Psr/Psr4Mapping.php#L44 and tested as well.

test/unit/Util/Autoload/ClassLoaderTest.php
kukulich: It's not called because of thrown exception.

test/unit/Util/Autoload/ClassLoaderTest.php
kukulich: The call is not called because of thrown exception.

test/unit/Util/Autoload/ClassLoaderTest.php
kukulich: I think that `ClassLoader` uses global state: https://github.com/Roave/BetterReflection/blob/5.1.x/src/Util/Autoload/ClassLoader.php#L26

Ocramius: Yes, hence why we unregister it? :thinking: 

Is it avoidable to run this in a separate process at all?

kukulich: Removed for now, it looks it works :)

Ocramius (approved):
Thanks @kukulich!

PR #975, opened 2022-01-18
Added `ReflectionType::getOwner()`

kukulich:
The method is not really necessary it should just kill some mutants: https://dashboard.stryker-mutator.io/reports/github.com/Roave/BetterReflection/5.1.x#mutant/Reflection/ReflectionUnionType.php

The constructor in `ReflectionType` is useless but I don't think we can remove it because it would be BC break...

### DIFF
--- a/src/Reflection/ReflectionType.php
+++ b/src/Reflection/ReflectionType.php
@@ -53,6 +53,14 @@ abstract class ReflectionType
     }
 
     /**
+     * @internal
+     */
+    public function getOwner(): ReflectionParameter|ReflectionMethod|ReflectionFunction|ReflectionEnum|ReflectionProperty
+    {
+        return $this->owner;
+    }
+
+    /**
      * Does the type allow null?
      */
     abstract public function allowsNull(): bool;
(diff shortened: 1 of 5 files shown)


src/Reflection/ReflectionType.php
```diff
@@ -52,6 +52,11 @@ public static function createFromNode(
         return new ReflectionUnionType($reflector, $owner, $type);
     }
 
+    public function getOwner(): ReflectionParameter|ReflectionMethod|ReflectionFunction|ReflectionEnum|ReflectionProperty
```
Ocramius, line 58: Should we perhaps mark this `@internal`, so we don't necessarily need to make this public API (yet)?

Ocramius (changes requested):

src/Reflection/ReflectionType.php
kukulich: 👍 Added

Ocramius (approved):

Ocramius:
Thanks @kukulich!

PR #979, opened 2022-01-19
Improved mutation score

kukulich:


kukulich:
@Ocramius Btw do we need `ClassLoader` ? It looks it used only in its own tests...

Ocramius:
> It looks it used only in its own tests...

I think it was used in the monkey-patching functionality, but no longer needed: perhaps deprecation + removal in `6.0.0`?

Ocramius (approved):
LGTM, nice refactoring + code removal 💪

Ocramius:
BTW: 

```
Metrics:
         Mutation Score Indicator (MSI): 97%
         Mutation Code Coverage: 99%
         Covered Code MSI: 97%
```

Perhaps time to raise that threshold again?

kukulich:
> Perhaps time to raise that threshold again?

When I'm done ;)

PR #982, opened 2022-01-19
Mark `Roave\BetterReflection\Util\Autoload` as deprecated

kukulich:
My bad... It looks there are a lot more unused classes... I think I've not seen these classes before :D


Ocramius (approved):

Ocramius:
Thanks @kukulich!

PR #983, opened 2022-01-19
Improved mutation score

kukulich:


src/Reflection/ReflectionClass.php
```diff
@@ -931,7 +931,7 @@ public function getDefaultProperties(): array
     {
         return array_map(
             static fn (ReflectionProperty $property) => $property->getDefaultValue(),
-            array_filter($this->getProperties(), static fn (ReflectionProperty $property): bool => $property->isDefault()),
+            $this->getProperties(),
```
Ocramius, line 936: Do `ReflectionClass` and `ReflectionObject` behave differently here?

src/Reflection/ReflectionProperty.php
```diff
@@ -197,12 +194,18 @@ public function isInitialized(?object $object = null): bool
...
+
+        // @phpstan-ignore-next-line
+        } catch (Error $e) {
+            if (str_contains($e->getMessage(), 'must not be accessed before initialization')) {
                 return false;
             }
 
+            // @codeCoverageIgnoreStart
+            // @infection-ignore-all
+            // Untestable code
```
Ocramius: This path is potentially testable with a class that does `unset($this->prop)`, then declares `__get()` and throws there

src/SourceLocator/Type/AnonymousClassObjectSourceLocator.php
```diff
@@ -101,7 +100,7 @@ public function __construct(private string $fileName, private int $startLine)
              */
             public function enterNode(Node $node)
             {
-                if (! ($node instanceof Node\Stmt\Class_) || $node->name !== null) {
+                if (! ($node instanceof Node\Stmt\Class_) || $node->name !== null || $node->getLine() !== $this->startLine) {
```
Ocramius, line 103: Would it be better to check start column too? Multiple classes on the same line can happen, especially in malicious software (which BetterReflection should be able to analyse)

Ocramius: Ah, I see that this code throws on multiple same-line classes 😩

src/SourceLocator/Type/ClosureSourceLocator.php
```diff
@@ -131,17 +133,15 @@ public function leaveNode(Node $node)
...
             {
-                $closureNodesDataOnSameLine = array_values(array_filter($this->closureNodes, fn (array $nodes): bool => $nodes['node']->getLine() === $this->startLine));
-
-                if (! $closureNodesDataOnSameLine) {
+                if ($this->closureNodes === []) {
                     throw NoClosureOnLine::create($this->fileName, $this->startLine);
                 }
 
-                if (isset($closureNodesDataOnSameLine[1])) {
+                if (isset($this->closureNodes[1])) {
```
Ocramius, line 140: With short closures becoming more popular, I feel like this area will just bite us in future. Refactoring done so far is good as-is though.

src/Util/ClassExistenceChecker.php
```diff
@@ -0,0 +1,42 @@
...
+    {
+        return interface_exists($name, false);
+    }
+
+    /**
+     * @psalm-suppress InvalidNullableReturnType
+     */
+    public static function traitExists(string $name): bool
+    {
+        /** @psalm-suppress NullableReturnStatement */
```
Ocramius, line 39: Nani?!??

src/Util/FileHelper.php
```diff
@@ -25,6 +25,10 @@ public static function normalizeSystemPath(string $originalPath): string
             [, $scheme, $path] = $matches;
         }
 
-        return ($scheme !== null ? $scheme . '://' : '') . str_replace(['/', '\\'], [DIRECTORY_SEPARATOR, DIRECTORY_SEPARATOR], $path);
+        if (DIRECTORY_SEPARATOR === '\\') {
```
Ocramius, line 28: How is this mutation-tested, since we don't run MT on windows? 🤔

Ocramius: Also, I wonder if we should just normalize to windows path names regardless?

test/unit/Fixture/InheritedClassProperties.php
```diff
@@ -22,7 +22,11 @@ abstract class Qux extends Baz
 
     public $f;
 
-    public function __construct(private $g)
+    public function mutantKillerBeforeConstructor()
```
Ocramius: Probably to be renamed `unrelatedMethodBeforeConstructor`

test/unit/Fixture/Methods.php
```diff
@@ -71,6 +71,7 @@ public function methodIsArrayParameters(
         ?array $nullableArrayParameter,
         null|array $unionArrayParameterNullFirst,
         array|null $unionArrayParameterNullLast,
+        array|NULL $unionArrayParameterNullUppercase,
```
Ocramius: I don't remember if our phpcs setup ignores assets 🤔

test/unit/SourceLocator/Type/ClosureSourceLocatorTest.php
```diff
@@ -187,4 +188,23 @@ public function testTwoClosuresSameLineFails(string $file, Closure $closure): vo
...
+        $reflection = $sourceLocator->locateIdentifier(
+            $reflector,
+            new Identifier(
+                ReflectionFunction::CLOSURE_NAME,
+                new IdentifierType(IdentifierType::IDENTIFIER_FUNCTION),
+            ),
+        );
+
+        self::assertInstanceOf(ReflectionFunction::class, $reflection);
+        self::assertSame('Roave\BetterReflectionTest\Fixture\ClassUsedAsClosureParameter', $reflection->getParameter('parameter')->getType()->__toString());
```
Ocramius: Isn't `getType` nullable? 🤔

test/unit/Util/ClassExistenceCheckerTest.php
```diff
@@ -0,0 +1,123 @@
...
+{
+    private ?Closure $loader = null;
+
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        require_once __DIR__ . '/../Fixture/TraitFixture.php';
+
+        // Should not be called
```
Ocramius: Let's have this call `self::fail()`, instead of actually doing an include

Ocramius (changes requested):

src/Reflection/ReflectionClass.php
kukulich: Yes in BR. `ReflectionClass` is created only from AST, so it cannot have runtime properties. Runtime properties are implemented only in `ReflectionObject`: https://github.com/Roave/BetterReflection/blob/5.1.x/src/Reflection/ReflectionObject.php#L69

So the original code cannot be covered by `ReflectionClass` tests because the support is only in `ReflectionObject`.

src/Util/ClassExistenceChecker.php
kukulich:
```
ERROR: InvalidNullableReturnType - src/Util/ClassExistenceChecker.php:36:55 - The declared return type 'bool' for Roave\BetterReflection\Util\ClassExistenceChecker::traitExists is not nullable, but 'bool|null' contains null (see https://psalm.dev/144)
    public static function traitExists(string $name): bool


ERROR: NullableReturnStatement - src/Util/ClassExistenceChecker.php:38:16 - The declared return type 'bool' for Roave\BetterReflection\Util\ClassExistenceChecker::traitExists is not nullable, but the function returns 'bool|null' (see https://psalm.dev/139)
        return trait_exists($name, false);
```

🤷‍♂️

test/unit/Fixture/Methods.php
kukulich: Yes: https://github.com/Roave/BetterReflection/blob/5.1.x/phpcs.xml.dist#L12

src/Util/FileHelper.php
kukulich: > How is this mutation-tested, since we don't run MT on windows?

You're right, it's not. I can probably add ignore annotation...

> Also, I wonder if we should just normalize to windows path names regardless?

It will report errors somewhere, where we check differences between BR and native reflection.

src/Reflection/ReflectionProperty.php
kukulich: Nice 👍

src/Reflection/ReflectionClass.php
Ocramius: Thanks for clarifying!

src/Util/FileHelper.php
Ocramius: > It will report errors somewhere, where we check differences between BR and native reflection.

Good point!

src/Util/ClassExistenceChecker.php
Ocramius: @orklah do you think we should report this in upstream `vimeo/psalm`?

Ocramius (approved):
Awesome, LGTM! 🚢 

Thanks @kukulich!

kukulich:
Some of the remaining mutants can probably be ignored, eg. mutants with `??` or `??=` operators.

Ocramius:
TBH, it always depends: I believe most escaped mutated nullability checks are hidden bugs/unused caches, so we'll need to evaluate case-by-case

kukulich:
Yes

- Some are optimization - it works with the change, it's just a little slower
- Some cannot be broken - eg. some of the `??`
- Some can be broken only with more changes - one mutant is not enough

src/Util/ClassExistenceChecker.php
orklah: :o trait_exists in Psalm is indeed documented as `?bool` since https://github.com/vimeo/psalm/pull/1541

Not sure where this was coming from. @voku do you have an idea?

In any case, yeah, please create an issue, we'll sort out the answer there!

PR #1004, opened 2022-02-11
Refactor - introduce `psalm-type` alias for composer `autoload` array shape

sasezaki:
please see https://github.com/composer/composer/pull/10516.
At moment, I'm on proposing introduce AutoloadMapping type to composer iteself.

I'm not sure installed.json's autoload mapping is same as composer.json schema. installed.json does not have schema(maybe)
https://github.com/composer/composer/tree/main/res
But it seems mostly same of composer.json's.

About phpstan, over 0.12.84 also supports "type". So we can use it.
https://github.com/phpstan/phpstan/discussions/4878

### DIFF
--- a/src/SourceLocator/Type/Composer/Factory/MakeLocatorForComposerJson.php
+++ b/src/SourceLocator/Type/Composer/Factory/MakeLocatorForComposerJson.php
@@ -26,6 +26,9 @@ use function is_file;
 use function json_decode;
 use function realpath;
 
+/**
+ * @psalm-type AutoloadMapping array{psr-0?: array<string, string|string[]>, psr-4?: array<string, string|string[]>, classmap?: list<string>, files?: list<string>, exclude-from-classmap?: list<string>}
+ */
 final class MakeLocatorForComposerJson
 {
     public function __invoke(string $installationPath, Locator $astLocator): SourceLocator
@@ -42,7 +45,7 @@ final class MakeLocatorForComposerJson
             throw MissingComposerJson::inProjectPath($installationPath);
         }
 
-        /** @psalm-var array{autoload: array{classmap: list<string>, files: list<string>, psr-4: array<string, list<string>|string>, psr-0: array<string, list<string>|string>}}|null $composer */
+        /** @psalm-var array{autoload: AutoloadMapping}|null $composer */
         $composer = json_decode((string) file_get_contents($composerJsonPath), true);
 
         if (! is_array($composer)) {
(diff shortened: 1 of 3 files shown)


src/SourceLocator/Type/Composer/Factory/MakeLocatorForComposerJson.php
```diff
@@ -26,6 +26,9 @@
 use function json_decode;
 use function realpath;
 
+/**
+ * @psalm-type AutoloadMapping array{psr-0?: array<string, string|string[]>, psr-4?: array<string, string|string[]>, classmap?: list<string>, files?: list<string>, exclude-from-classmap?: list<string>}
```
Ocramius: Let's please use multiline here 😁

Ocramius: I think the naming is a bit confusing: this isn't a "mapping", but rather the autoload definition structure itself

Ocramius (changes requested):
Overall LGTM, besides naming 👍

src/SourceLocator/Type/Composer/Factory/MakeLocatorForComposerJson.php
sasezaki: > I think the naming is a bit confusing: this isn't a "mapping", but rather the autoload definition structure itself

oh my bad, I couldn't have thought good naming.

type name would be conflicted global class name, so I did avoid "Autoload".  So, I am now thinking rename as "ComposerAutoload". (Anyone has suggestion ?)

sasezaki: @Ocramius 
just now, renamed AutoloadMapping to ComposerAudoload.

Ocramius (approved):
Thanks @sasezaki!

src/SourceLocator/Type/Composer/Factory/MakeLocatorForComposerJson.php
```diff
@@ -26,6 +26,15 @@
 use function json_decode;
 use function realpath;
 
+/**
+ * @psalm-type ComposerAudoload array{
```
Seldaek, line 30: Typo `ComposerAudoload` => `ComposerAutoload`.

Anyway the types from https://github.com/composer/composer/pull/10516 will ship with the next Composer 2.2 patch release if you want to start relying on that instead.

Ocramius: Good catch! Still, adding a dependency for just the type alias is a bit much, in this case

Seldaek: Yeah I thought maybe it could be a dev dependency as you only need it for static analysis, but I'm not sure if the APIs are exposed publicly here, if that's the case then yes please duplicate rather than depending on composer/composer, that always ends up horribly.

PR #1019, opened 2022-02-27
Mini optimization of `CompileNodeToValue`

kukulich:
We can use the already known class - there's no need to reflect it again.


Ocramius (approved):

Ocramius:
Thanks @kukulich - directly cutting a release 👍

PR #1050, opened 2022-03-21
Fix #1043 - make `CompileNodeToValue` filesystem-agnostic

### LINKED ISSUE #1043: Make `CompileNodeToValue` semi-`@psalm-immutable`?
While working on:

 * https://github.com/Roave/BackwardCompatibilityCheck/issues/400
 * https://github.com/Roave/BackwardCompatibilityCheck/pull/476

I discovered that `CompileNodeToValue` internally uses `realpath()` and `dirname()` to chop off parts of file names.

This is problematic for any kind of processing where the `LocatedSource` does not have a file path, such as in serialized `LocatedSource` trees, where the filesystem is no longer accessible.

In addition to that, `realpath()` and `dirname()` do cause runtime stat calls to be avoided.

We should most likely adapt `CompileNodeToValue` to operate without attempting to access the filesystem (on its own) while evaluating constant expressions: `@psalm-immutable` is not achievable due to the large traversing of other reflection symbols, but we should get to a `@psalm-external-mutation-free` or similar scenario.

Ocramius:
With this change, `CompileNodeToValue` will honor the source location given by the source locator, and will not use the local filesystem to identify `__DIR__` nor `__FILE__`.

Note: as it currently is, this patch is just a stub of the final wished result, as some cleanup is still necessary.

Also, this should remove some filesystem calls that could happen deep inside `roave/better-reflection`'s core, potentially speeding up the whole thing :D 

Fixes #1043

### DIFF
--- a/test/unit/NodeCompiler/CompileNodeToValueTest.php
+++ b/test/unit/NodeCompiler/CompileNodeToValueTest.php
@@ -4,13 +4,13 @@ declare(strict_types=1);
 
 namespace Roave\BetterReflectionTest\NodeCompiler;
 
+use BadMethodCallException;
 use PhpParser\Node;
 use PhpParser\Node\Expr\ConstFetch;
 use PhpParser\Node\Expr\Yield_;
 use PhpParser\Node\Name;
 use PhpParser\Parser;
 use PHPUnit\Framework\TestCase;
-use Roave\BackwardCompatibility\SourceLocator\LocatedSourceWithStrippedSourcesDirectory;
 use Roave\BetterReflection\Identifier\Identifier;
 use Roave\BetterReflection\Identifier\IdentifierType;
 use Roave\BetterReflection\NodeCompiler\CompileNodeToValue;
@@ -685,7 +685,7 @@ PHP;
 
     public function testCanRetrieveMagicConstantValueWhenUsingFakeSourceLocator(): void
     {
-        $astLocatorWithFakeFilePath = new class ($this->astLocator) extends Locator {
+        $astLocatorProducingLocatedSourcesWithFakeFilePath = new class ($this->astLocator) extends Locator {
             public function __construct(private Locator $next)
             {
             }
@@ -703,7 +703,6 @@ PHP;
                         {
                         }
 
-                        // @TODO test that all methods are covered (use reflection)
                         public function getSource(): string
                         {
                             return $this->next->getSource();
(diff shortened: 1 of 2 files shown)


test/unit/NodeCompiler/CompileNodeToValueTest.php
```diff
@@ -678,12 +683,108 @@ public function magicConstantsInNamespaceProvider(): array
         ];
     }
 
+    public function testCanRetrieveMagicConstantValueWhenUsingFakeSourceLocator(): void
+    {
+        $astLocatorWithFakeFilePath = new class ($this->astLocator) extends Locator {
```
Ocramius: Not a fan of this declaration - potentially moving it to a new file?

test/unit/NodeCompiler/CompileNodeToValueTest.php
```diff
@@ -678,12 +683,108 @@ public function magicConstantsInNamespaceProvider(): array
...
+        );
+    }
+
     /**
      * @dataProvider magicConstantsInNamespaceProvider
      */
     public function testMagicConstantsInNamespace(string $constantName, mixed $expectedValue): void
     {
-        $reflector = new DefaultReflector(new SingleFileSourceLocator(__DIR__ . '/../Fixture/MagicConstants.php', $this->astLocator));
+        $reflector = new DefaultReflector(new SingleFileSourceLocator(realpath(__DIR__ . '/../Fixture/MagicConstants.php'), $this->astLocator));
```
Ocramius, line 788: Minor breakage here, but the file path should still be valid (even if relative), so IMO this is an OK change for 5.3.0

asgrim (approved):
Seems decent :+1:

test/unit/NodeCompiler/CompileNodeToValueTest.php
```diff
@@ -678,12 +683,108 @@ public function magicConstantsInNamespaceProvider(): array
...
+
+        self::assertSame(
+            '/non/existing/path/to',
+            $reflector
+                ->reflectConstant('CURRENT_DIR')
+                ->getValue()
+        );
+
+        self::assertSame(
+            '/non/existing/path/to/sources.php',
```
Ocramius, line 775: Not sure whether this test will "just work" on windows.

kukulich (approved):
Tests are green on Windows so 👍

test/unit/NodeCompiler/CompileNodeToValueTest.php
Ocramius: Decided to keep it here, and fixed variable name.

test/unit/NodeCompiler/CompileNodeToValueTest.php
Ocramius: Confirmed by @kukulich that this "just works"

Ocramius:
Now in `5.3.0` :)

PR #1082, opened 2022-04-16
RFC: Reflector to reflect on all attributes

cspray:
A bare-minimum, proof-of-concept implementation for gathering all ReflectionAttributes from a given source. This PR updates the Reflector interface and DefaultReflector implementation to include a reflectAllAttributes() method that will return a list of ReflectionAttribute. Additionally, several type-hints on parameters and method returns have been updated to include Node\Attribute and ReflectionAttribute where appropriate.

This implementation is not meant to represent something fully working or ready to be merged in but to spark a conversation on the desired feature.

---

Hi! Thanks for the awesome library! I have a use case for statically analyzing Attributes in a PHP codebase. To facilitate this I added a `Reflector::reflectAllAttributes()` method that will return a list of ReflectionAttribute instances. Before I continued with this work I wanted to get the maintainers feedback to:

1. See if this feature is even desired by the team and would be open to merging in to the codebase.
2. Collaborate on possible concerns with the implementation and to solidify the changes to the Reflector API.

## Todo

1. Finish testing and implementing the `ReflectionAttribute::createFromNode` method to construct the correct reflection for the node that owns the attribute.
2. Introduce additional methods on the `Reflector` interface for narrowing the ReflectionAttributes returned.

### DIFF
--- a/src/Identifier/IdentifierType.php
+++ b/src/Identifier/IdentifierType.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Roave\BetterReflection\Identifier;
 
 use InvalidArgumentException;
+use Roave\BetterReflection\Reflection\ReflectionAttribute;
 use Roave\BetterReflection\Reflection\ReflectionClass;
 use Roave\BetterReflection\Reflection\ReflectionConstant;
 use Roave\BetterReflection\Reflection\ReflectionFunction;
@@ -12,23 +13,23 @@ use Roave\BetterReflection\Reflection\ReflectionFunction;
 use function array_key_exists;
 use function sprintf;
 
-class IdentifierType
-{
-    public const IDENTIFIER_CLASS    = ReflectionClass::class;
+class IdentifierType {
+    public const IDENTIFIER_CLASS = ReflectionClass::class;
     public const IDENTIFIER_FUNCTION = ReflectionFunction::class;
     public const IDENTIFIER_CONSTANT = ReflectionConstant::class;
+    public const IDENTIFIER_ATTRIBUTE = ReflectionAttribute::class;
 
     private const VALID_TYPES = [
-        self::IDENTIFIER_CLASS    => null,
+        self::IDENTIFIER_CLASS => null,
         self::IDENTIFIER_FUNCTION => null,
         self::IDENTIFIER_CONSTANT => null,
+        self::IDENTIFIER_ATTRIBUTE => null,
     ];
 
     private string $name;
 
-    public function __construct(string $type = self::IDENTIFIER_CLASS)
-    {
-        if (! array_key_exists($type, self::VALID_TYPES)) {
+    public function __construct(string $type = self::IDENTIFIER_CLASS) {
+        if (!array_key_exists($type, self::VALID_TYPES)) {
             throw new InvalidArgumentException(sprintf(
                 '%s is not a valid identifier type',
                 $type,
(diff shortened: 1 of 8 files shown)


src/Identifier/IdentifierType.php
```diff
@@ -5,30 +5,32 @@
...
 use Roave\BetterReflection\Reflection\ReflectionConstant;
 use Roave\BetterReflection\Reflection\ReflectionFunction;
 
 use function array_key_exists;
 use function sprintf;
 
-class IdentifierType
-{
-    public const IDENTIFIER_CLASS    = ReflectionClass::class;
+class IdentifierType {
```
asgrim, line 16: Please use `vendor/bin/phpcbf` to fix the unnecessary coding style changes

src/Reflector/Reflector.php
```diff
@@ -52,4 +53,11 @@ public function reflectConstant(string $identifierName): ReflectionConstant;
      * @return list<ReflectionConstant>
      */
     public function reflectAllConstants(): iterable;
+
+    /**
+     * Get all attributes available in the scope specified by the SourceLocator.
+     *
+     * @return List<ReflectionAttribute>
```
asgrim, line 60: Hmm, not sure if case matters to Psalm, but I think it should be `list<ReflectionAttribute>` for consistency either way

src/Reflector/Reflector.php
```diff
@@ -52,4 +53,11 @@ public function reflectConstant(string $identifierName): ReflectionConstant;
...
+     */
+    public function reflectAllAttributes(): iterable;
```
asgrim, line 62: Note: adding a method to an interface is a BC break

asgrim (commented):
Couple of questions:

1) What do you see the use case as? There's minimal testing here (noted in description), so it's difficult to understand how the feature can be used

2) Given a list of `ReflectionAttribute` instances, what can you actually do with this information? Attributes on a method, for example, are outwardly linked to the owner for example, so you cannot know where the attributes are known. Is this useful information? It might not be, but just querying as it seems that once you have a list of "all attributes in a source", you can't do much with the list, apart from know "which attributes are used" in a given source (maybe that's the aim of this, but it wasn't clear, see point 1 heh)

cspray:
Hi! Thanks for the feedback. When I originally introduced this I was under the wrong assumption that the ReflectionAttribute class included the information about the target as well. Now that I realize I was mistaken my expected use case is no longer valid. Sorry for the confusion!

asgrim:
@cspray

>  I was under the wrong assumption that the ReflectionAttribute class included the information about the target as well.

I mean - the `$owner` is indeed given to `ReflectionAttribute` at the moment, but it isn't really useful beyond a basic check of type of owner. If there is a valid use case, it _could_ be added, but ONLY if there is a valid use case for it :+1:

cspray:
So the use-case I had in mind is actually to replace the static analysis I had already written for https://github.com/cspray/annotated-container. The idea is that I can parse various Attributes to configure a dependency injection container. For example...

```php

#[Service]
interface Storage {

    public function get(string $key) : mixed;

    public function put(string $key, mixed $val) : void;

}

#[Service]
class FilesystemStorage {

    public function __construct(#[Inject('/path/to/storage/dir')] private string $dir) {}

    public function get(string $key) : mixed {
        // ...
    }

    public function put(string $key) : void {
        // ...
    }

}

class StorageConsumer {

    public function __construct(Storage $storage) {}

}
```

With this setup I could determine that when you created `StorageConsumer` with the container we should use the `FielsystemStorage` class and inject the string `/path/to/storage/dir` on object construction.

The static analysis that I have currently gets the job done but, as I'm sure you're aware, statically analyzing PHP has some pitfalls and gotchas. I would prefer to use a library that is more likely to handle those pitfalls and gotchas correctly.

Ocramius:
The approach to that is:

 1. Reflect all classes
 2. Iterate
 3. Filter those with the wished attribute(s)

cspray:
@Ocramius The concern there is that the owner of the Attribute might not be a class. It could be a property, a method, a parameter on the method, or a class constant. Not only would you have to iterate over every ReflectionClass but all of its properties and methods as well. In addition, this would not include places where a function is the owner of the Attribute.

Ocramius:
> The concern there is that the owner of the Attribute might not be a class.

Up to you in userland to check that: the attribute is a class itself, so you can reflect it.

> Not only would you have to iterate over every ReflectionClass but all of its properties and methods as well.

I only took your `#[Service]` example: reflecting a class or all its properties/methods is not that big difference.

> In addition, this would not include places where a function is the owner of the Attribute.

My suggestion is to leave functions alone: they are super messy to work with, and BetterReflection Isn't better than core PHP at dealing with them.

Remember that modern PHP is autoloader-based, so iterating over a whole codebase is normal, when trying to define any sort of map that isn't declared upfront in a single deterministic location.

PR #1123, opened 2022-06-09
Unit tests tested by PHPStan

### LINKED ISSUE #1122: `test` is not covered by static analysis
Discussed with @ondrejmirtes - `test/` is not covered by static analysis :scream: 

We should have both PHPStan and Psalm cover that ASAP, even if it means expanding the baseline by a lot.

kukulich:
Currently with generated baseline with 323 errors.

Fixes #1122

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -35,6 +35,7 @@
     "require-dev": {
         "doctrine/coding-standard": "^9.0.0",
         "phpstan/phpstan": "^1.7.11",
+        "phpstan/phpstan-phpunit": "^1.1.1",
         "phpunit/phpunit": "^9.5.20",
         "vimeo/psalm": "^4.23",
         "roave/infection-static-analysis-plugin": "^1.19.0"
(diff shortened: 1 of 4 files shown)


composer.json
```diff
@@ -34,7 +34,9 @@
     ],
     "require-dev": {
         "doctrine/coding-standard": "^9.0.0",
+        "phpstan/extension-installer": "^1.1.0",
```
Ocramius: Can we install the extension without the extension installer?

phpstan.neon
```diff
@@ -1,16 +1,26 @@
...
+
 parameters:
+    phpVersion: 80100
+
     level: 6
     checkGenericClassInNonGenericObjectType: false
 
     paths:
         - src
+        - test/unit
```
Ocramius, line 15: Let's also check `test/benchmark`

Ocramius (changes requested):

composer.json
kukulich: Yes, done.

phpstan.neon
kukulich: Done.

Ocramius (approved):

PR #1125, opened 2022-06-10
PHPStan checks without baseline

kukulich:


Ocramius (approved):

PR #1127, opened 2022-06-13
Calling `setAccessible()` has no effect since PHP 8.1 - all methods/properties are accessible by default

### LINKED ISSUE #1121: ReflectionMethod Adapter: Calling setAccessible() shouldn't be needed for non-public methods as of PHP 8.1
https://www.php.net/manual/en/reflectionmethod.setaccessible.php

> Note: As of PHP 8.1.0, calling this method has no effect; all methods are invokable by default. 

But the adapter here still requires it before calling `invoke()`: https://github.com/Roave/BetterReflection/blob/c834c79e3c116346c830f4921a7aaafe2e35ea58/src/Reflection/Adapter/ReflectionMethod.php#L238-L240

The options IMHO are:

1) Remove the `isAccessible()` check and be done with it.
2) Require it only when running on PHP 8 but not on PHP 8.1+.
3) Remove the `isAccessible()` check and set PHP 8.1 as a new minimum version.

Originally reported: https://github.com/phpstan/phpstan/issues/7429

kukulich:
Fixes https://github.com/Roave/BetterReflection/issues/1121

### DIFF
--- a/src/Reflection/Adapter/ReflectionClass.php
+++ b/src/Reflection/Adapter/ReflectionClass.php
@@ -403,10 +403,6 @@ final class ReflectionClass extends CoreReflectionClass
 
         $property = new ReflectionProperty($betterReflectionProperty);
 
-        if (! $property->isAccessible()) {
-            throw new CoreReflectionException(sprintf('Property "%s" is not accessible', $name));
-        }
-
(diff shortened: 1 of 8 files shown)


Ocramius (approved):
Awesome, thanks @kukulich!

PR #1131, opened 2022-06-13
Implemented all missing properties in Reflection adapters

### LINKED ISSUE #1120: Not all public properties in Adapter namespace work correctly
There's currently an `unset` + `__get` hack for `ReflectionClass::$name`: https://github.com/Roave/BetterReflection/blob/c834c79e3c116346c830f4921a7aaafe2e35ea58/src/Reflection/Adapter/ReflectionClass.php#L38 + https://github.com/Roave/BetterReflection/blob/c834c79e3c116346c830f4921a7aaafe2e35ea58/src/Reflection/Adapter/ReflectionClass.php#L46-L53

We need to do a similar hack for other public properties. There's about 12 of them judging from https://github.com/JetBrains/phpstorm-stubs/tree/master/Reflection.

Otherwise we can get errors like `Error: Typed property ReflectionParameter::$name must not be accessed before initialization`.

Originally reported: https://github.com/phpstan/phpstan/issues/7428

kukulich:
Fixes https://github.com/Roave/BetterReflection/issues/1120

### DIFF
--- a/src/Reflection/Adapter/ReflectionClassConstant.php
+++ b/src/Reflection/Adapter/ReflectionClassConstant.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Roave\BetterReflection\Reflection\Adapter;
 
+use OutOfBoundsException;
 use ReflectionClassConstant as CoreReflectionClassConstant;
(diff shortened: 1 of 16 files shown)


Ocramius (approved):
Thanks @kukulich!

Issue #1073, opened 2022-04-08
Prevent memory leaks by not holding onto AST nodes


ondrejmirtes:
Hi, I'm in the process of reducing the memory amount consumed by PHPStan, and I have some promising results.

Current stable version consumes ~850 MB when runing PHPStan on PHPStan.

After I removed the NodeConnectingVisitor (this isn't related to BetterReflection) that creates cycles between nodes by setting parent/previous/next attributes, I saved around 100 MB. So now we're at ~750 MB.

I continued to debug the problem by using [php-meminfo](https://github.com/BitOne/php-meminfo) but there was a lot of noise because of CachedParser (equivalent of MemoizingParser). So I removed parser caching to see what else holds onto AST nodes. I also removed static reflection capabilities altogether to see if there's something else in PHPStan that leaks. I consider the floor of what's possible to achieve at around ~200 MB.

When I returned static reflection capabilities, it went back to ~750 MB. And most of the memory is held in AST nodes held by various reflection classes like ReflectionClass etc.

So I refactored some classes to extract all of the needed information in constructor so that the AST node is not saved in a property: https://github.com/ondrejmirtes/BetterReflection/compare/cc581ecece6e69674181ec456ecc5d846ad312d1...35db84f1abff5b3416daca4804901b0750da7610 (the work isn't done, there's still more to do)

Right now PHPStan consumes around 340 MB without a CachingParser and around 430 MB with a CachingParser.

The best part is that there's very little performance impact time-wise so it looks really promising. I don't have production-ready code yet as there are some problems to work out but I hope this will make PHPStan much more resource-effective.

One downside is that I had to remove some features that need the AST after a Reflection class is created - it can be seen in the diff.

Is this something you'd like to do here too?

Ocramius:
AFAIK, BR nodes are acyclic: the `CachedParser` was introduced on purpose, to prevent double-parsing of data (which happens a lot when jumping around a codebase, since we don't cache reflected symbols either).

IMO, a better approach would be to make the codebase as `@immutable` as possible:

 * `ReflectionClass`
 * `ReflectionProperty`
 * `ReflectionType`
 * ...

Besides some rare edge cases in their API, we can likely make these completely immutable, by deferring creation of stuff that requires I/O to the reflector itself.

This may potentially change the reflection API radically though :thinking: 

I suppose the code removals you mentioned are mostly around `getAst()`?

ondrejmirtes:
> BR nodes are acyclic

[...]

Issue #981, opened 2022-01-19
6.0 cleanup


kukulich:
- [x] Remove `Roave\BetterReflection\Util\Autoload` https://github.com/Roave/BetterReflection/pull/1142
- [x] Make `ReflectionEnum::__construct()` private https://github.com/Roave/BetterReflection/pull/1139
- [x] Remove `ReflectionType::__construct()` and `ReflectionType::getOwner()` https://github.com/Roave/BetterReflection/pull/1140
- [x] Remove adapter `ReflectionProperty::isAccessible()` https://github.com/Roave/BetterReflection/pull/1141

Ocramius:
@kukulich btw, milestone + issues > checklists :D

Ocramius:
Awesome work, @kukulich! :tada:

PR #1138, opened 2022-06-26
`StringSourceLocator` is forced to be built with a `non-empty-string`

### LINKED ISSUE #932: Enforce `StringSourceLocator` to be built with a `non-empty-string`
Currently, a `StringSourceLocator` can be constructed with an empty string, but will crash at runtime when doing so:

https://github.com/Roave/BetterReflection/blob/f3966ba125e3b4db948a240e8b29bc28882623c1/src/SourceLocator/Type/StringSourceLocator.php#L23-L38

We should really get rid of the exception, and instead enforce `non-empty-string` as input.

kukulich:
Fixes #932

I've started with 6.0.0 cleanup.


Ocramius (approved):
LGTM. @asgrim this removes a runtime exception, so it could be seen as a BC break, are you fine with it?

src/SourceLocator/Exception/EmptyPhpSourceCode.php
```diff
@@ -1,11 +0,0 @@
-<?php
-
-declare(strict_types=1);
-
-namespace Roave\BetterReflection\SourceLocator\Exception;
-
-use InvalidArgumentException;
-
-class EmptyPhpSourceCode extends InvalidArgumentException
```
Ocramius, line 9: I think we deprecate and remove this in the next major

kukulich: So one PR for 5.7 with `@deprecated` and this one should target 6.0.0?

Ocramius: as you prefer on that: next major release branch is not yet open

kukulich: I would be happy to just open 6.0 branch so I can target PR for #932, #981 and #1073

Ocramius: Works for me: beware that automatic releases will stop creating 5.x branches, once the new major branch is here

kukulich: Is there a way how I can open the branch? Or will you please do it for me? :)

Ocramius: Oh, it already exists: https://github.com/Roave/BetterReflection/tree/6.0.x

Never mind 😁

Ocramius (approved):
Thanks @kukulich!

PR #1140, opened 2022-06-26
Reflection types cleanup

kukulich:


src/Reflection/ReflectionIntersectionType.php
```diff
@@ -18,15 +18,16 @@ class ReflectionIntersectionType extends ReflectionType
...
+        private ReflectionParameter|ReflectionMethod|ReflectionFunction|ReflectionEnum|ReflectionProperty $owner,
         IntersectionType $type,
     ) {
-        parent::__construct($reflector, $owner);
-
-        $this->types = array_values(array_map(static function (Node\Identifier|Node\Name $type) use ($reflector, $owner): ReflectionNamedType {
-            $type = ReflectionType::createFromNode($reflector, $owner, $type);
+        $this->types = array_values(array_map(function (Node\Identifier|Node\Name $type): ReflectionNamedType {
+            $type = ReflectionType::createFromNode($this->reflector, $this->owner, $type);
             assert($type instanceof ReflectionNamedType);
```
Ocramius, line 29: This will change with PHP 8.2

src/Reflection/ReflectionUnionType.php
```diff
@@ -18,15 +18,16 @@ class ReflectionUnionType extends ReflectionType
...
+        private ReflectionParameter|ReflectionMethod|ReflectionFunction|ReflectionEnum|ReflectionProperty $owner,
         UnionType $type,
     ) {
-        parent::__construct($reflector, $owner);
-
-        $this->types = array_values(array_map(static function (Node\Identifier|Node\Name $type) use ($reflector, $owner): ReflectionNamedType {
-            $type = ReflectionType::createFromNode($reflector, $owner, $type);
+        $this->types = array_values(array_map(function (Node\Identifier|Node\Name $type): ReflectionNamedType {
+            $type = ReflectionType::createFromNode($this->reflector, $this->owner, $type);
             assert($type instanceof ReflectionNamedType);
```
Ocramius, line 29: This will change with PHP 8.2

src/Reflection/ReflectionType.php
```diff
@@ -73,14 +67,6 @@ public static function createFromNode(
         return new ReflectionUnionType($reflector, $owner, new UnionType($types));
     }
 
-    /**
-     * @internal
-     */
-    public function getOwner(): ReflectionParameter|ReflectionMethod|ReflectionFunction|ReflectionEnum|ReflectionProperty
```
Ocramius, line 79: Hmm, removing this means no longer being able to run some checks in `roave/backwards-compatibility-check` 🤔

kukulich: I can also remove the `@internal` flag :)

Ocramius: I'm not saying we shouldn't do it, I'm saying that we need some sort of approach to how to deal with it downstream :D 

Practically, if we move reflection stuff to be more and more immutable, we should totally do this.

kukulich: Do you really use this method in `roave/backwards-compatibility-check`? I'm not able to find it. Didn't you mean https://github.com/Roave/BetterReflection/blob/5.7.x/src/Reflection/ReflectionNamedType.php#L65 ?

kukulich: History of the method: https://github.com/Roave/BetterReflection/pull/975

Ocramius: Hmm, perhaps we can try this with the other lib, as I may already have worked around it 🤷

Ocramius: Checked again: LGTM, let's roll with it :)

Ocramius (approved):
:ship: thanks @kukulich!

PR #1152, opened 2022-07-22
Added missing `true` type as `ReflectionNamedType#isBuiltIn()` value

staabm:
https://wiki.php.net/rfc/true-type

inspired by https://github.com/Roave/BetterReflection/commit/0f76eb13b72700f5f56099713bc6cf43f5a46973 I figured this might already be evertyhing required..?


Ocramius (approved):
🚢 thanks @staabm!

staabm:
Thank you

PR #1191, opened 2022-09-05
Basic support for PHP 8.2

kukulich:


Ocramius:
I could release this as last 5.x, unless more 8.2 additions can be done without BC issues? 🤔

kukulich:
I think the last 8.2 addidition will be https://wiki.php.net/rfc/constants_in_traits and I'm currently expecting it could be done without BC breaks. I plan to prepare it soon.

Ocramius:
Ok, we should then keep the milestone open until that is through, and then we can merge all BC breaking changes and prepare 6.0.0 💪

Ocramius (approved):
🚢 thanks @kukulich!

PR #1192, opened 2022-09-06
PHP 8.2: Support for `readonly` classes

kukulich:
It's easier to implement this first and then make green build on PHP 8.2.

### DIFF
--- a/src/Reflection/ReflectionClass.php
+++ b/src/Reflection/ReflectionClass.php
@@ -62,6 +62,14 @@ use function strtolower;
 
 class ReflectionClass implements Reflection
 {
+    /**
+     * We cannot use CoreReflectionClass::IS_READONLY because it does not exist in PHP < 8.2.
+     * Constant is public, so we can use it in tests.
+     *
+     * @internal
+     */
+    public const IS_READONLY = 65536;
+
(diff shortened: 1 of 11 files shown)


src/Reflection/ReflectionClass.php
```diff
@@ -62,6 +62,8 @@
 
 class ReflectionClass implements Reflection
 {
+    public const IS_READONLY = 65536;
```
Ocramius, line 71: Should we make this `public`? Seems like a mistake.

Ocramius (changes requested):

src/Reflection/ReflectionClass.php
kukulich: It's public because of tests: https://github.com/Roave/BetterReflection/pull/1192/files#diff-e0dd711174c42542396ea4773a4b7db9ba6025250e7d50950c152a6d43b5fe91R1154

It's same as https://github.com/Roave/BetterReflection/blob/5.10.x/src/Reflection/ReflectionClassConstant.php#L21

Maybe I can mark it as `@internal` ?

Ocramius: `@internal` with an explanation would suffice 👍

Ocramius (approved):
LGTM :ship:

PR #1196, opened 2022-09-08
PHP 8.2: Support for fetching properties of `enum`s in `const` expressions

kukulich:
https://wiki.php.net/rfc/fetch_property_in_const_expressions


Ocramius (approved):
Excellent! Thanks for implementing strict limitations here, @kukulich!

PR #1197, opened 2022-09-08
PHP 8.2: Support for standalone `null` type

kukulich:
https://wiki.php.net/rfc/null-false-standalone-types


Ocramius (approved):
Thanks @kukulich! 🚀

PR #1198, opened 2022-09-08
PHP 8.2: Support for disjunctive normal form types 

kukulich:
https://wiki.php.net/rfc/dnf_types

https://github.com/Roave/BetterReflection/pull/1197 needs to be merged first.

### DIFF
--- a/src/Reflection/Adapter/ReflectionIntersectionType.php
+++ b/src/Reflection/Adapter/ReflectionIntersectionType.php
@@ -7,10 +7,9 @@ namespace Roave\BetterReflection\Reflection\Adapter;
 use ReflectionIntersectionType as CoreReflectionIntersectionType;
 use Roave\BetterReflection\Reflection\ReflectionIntersectionType as BetterReflectionIntersectionType;
 use Roave\BetterReflection\Reflection\ReflectionNamedType as BetterReflectionNamedType;
-use Roave\BetterReflection\Reflection\ReflectionUnionType as BetterReflectionUnionType;
 
-use function array_filter;
 use function array_map;
+use function assert;
 
 class ReflectionIntersectionType extends CoreReflectionIntersectionType
 {
@@ -18,13 +17,15 @@ class ReflectionIntersectionType extends CoreReflectionIntersectionType
     {
     }
 
-    /** @return array<ReflectionNamedType> */
+    /** @return non-empty-list<ReflectionNamedType> */
     public function getTypes(): array
     {
-        return array_filter(
-            array_map(static fn (BetterReflectionNamedType|BetterReflectionUnionType|BetterReflectionIntersectionType $type): ReflectionNamedType|ReflectionUnionType|ReflectionIntersectionType|null => ReflectionType::fromTypeOrNull($type), $this->betterReflectionType->getTypes()),
-            static fn (ReflectionNamedType|ReflectionUnionType|ReflectionIntersectionType|null $type): bool => $type instanceof ReflectionNamedType,
-        );
+        return array_map(static function (BetterReflectionNamedType $type): ReflectionNamedType {
+            $adapterType = ReflectionType::fromType($type);
+            assert($adapterType instanceof ReflectionNamedType);
+
+            return $adapterType;
+        }, $this->betterReflectionType->getTypes());
     }
 
     public function __toString(): string
--- a/src/Reflection/Adapter/ReflectionType.php
+++ b/src/Reflection/Adapter/ReflectionType.php
@@ -7,6 +7,7 @@ namespace Roave\BetterReflection\Reflection\Adapter;
 use ReflectionType as CoreReflectionType;
 use Roave\BetterReflection\Reflection\ReflectionIntersectionType as BetterReflectionIntersectionType;
 use Roave\BetterReflection\Reflection\ReflectionNamedType as BetterReflectionNamedType;
+use Roave\BetterReflection\Reflection\ReflectionType as BetterReflectionType;
 use Roave\BetterReflection\Reflection\ReflectionUnionType as BetterReflectionUnionType;
 
 use function array_filter;
@@ -15,12 +16,14 @@ use function count;
 
 abstract class ReflectionType extends CoreReflectionType
 {
-    public static function fromTypeOrNull(BetterReflectionNamedType|BetterReflectionUnionType|BetterReflectionIntersectionType|null $betterReflectionType): ReflectionUnionType|ReflectionNamedType|ReflectionIntersectionType|null
+    public static function fromTypeOrNull(BetterReflectionUnionType|BetterReflectionNamedType|BetterReflectionIntersectionType|null $betterReflectionType): ReflectionUnionType|ReflectionNamedType|ReflectionIntersectionType|null
     {
-        if ($betterReflectionType === null) {
-            return null;
-        }
+        return $betterReflectionType !== null ? self::fromType($betterReflectionType) : null;
+    }
 
+    /** @internal */
+    public static function fromType(BetterReflectionNamedType|BetterReflectionUnionType|BetterReflectionIntersectionType $betterReflectionType): ReflectionUnionType|ReflectionNamedType|ReflectionIntersectionType
+    {
         if ($betterReflectionType instanceof BetterReflectionUnionType) {
             // php-src has this weird behavior where a union type composed of a single type `T`
             // together with `null` means that a `ReflectionNamedType` for `?T` is produced,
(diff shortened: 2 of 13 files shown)


src/Reflection/Adapter/ReflectionIntersectionType.php
```diff
@@ -18,13 +16,10 @@ public function __construct(private BetterReflectionIntersectionType $betterRefl
     {
     }
 
-    /** @return array<ReflectionNamedType> */
+    /** @return list<ReflectionUnionType|ReflectionIntersectionType|ReflectionNamedType> */
```
Ocramius: This becomes a BC break that requires a new major 🤔

src/Reflection/Adapter/ReflectionType.php
```diff
@@ -7,20 +7,18 @@
...
 
 use function array_filter;
 use function array_values;
+use function assert;
 use function count;
 
 abstract class ReflectionType extends CoreReflectionType
 {
-    public static function fromTypeOrNull(BetterReflectionNamedType|BetterReflectionUnionType|BetterReflectionIntersectionType|null $betterReflectionType): ReflectionUnionType|ReflectionNamedType|ReflectionIntersectionType|null
+    public static function fromType(BetterReflectionType $betterReflectionType): ReflectionUnionType|ReflectionNamedType|ReflectionIntersectionType
```
Ocramius: The union type would still be better here, compared to `BetterReflectionType`, which could be extended by third parties :thinking:

src/Reflection/Adapter/ReflectionIntersectionType.php
```diff
@@ -6,25 +6,26 @@
...
 
-    /** @return array<ReflectionNamedType> */
+    /** @return list<ReflectionNamedType> */
     public function getTypes(): array
     {
-        return array_filter(
-            array_map(static fn (BetterReflectionNamedType|BetterReflectionUnionType|BetterReflectionIntersectionType $type): ReflectionNamedType|ReflectionUnionType|ReflectionIntersectionType|null => ReflectionType::fromTypeOrNull($type), $this->betterReflectionType->getTypes()),
-            static fn (ReflectionNamedType|ReflectionUnionType|ReflectionIntersectionType|null $type): bool => $type instanceof ReflectionNamedType,
-        );
+        return array_map(static function (BetterReflectionType $type): ReflectionNamedType {
```
Ocramius: `BetterReflectionType` is not `final`: IMO, the union type should be preserved :thinking:

Ocramius (changes requested):
I'm a bit confused about rolling back union types referencing concrete types, and going back to base `ReflectionType` usage :thinking:

src/Reflection/Adapter/ReflectionType.php
kukulich: Fixed.

src/Reflection/Adapter/ReflectionIntersectionType.php
kukulich: Changed to `BetterReflectionNamedType` because `ReflectionIntersectionType::getTypes()` returns `list<ReflectionNamedType>`

src/Reflection/Adapter/ReflectionIntersectionType.php
kukulich: Not valid anymore.

src/Reflection/Adapter/ReflectionIntersectionType.php
Ocramius: I see :thinking: 

Need to reason about this a bit, 'bit slow after processing a few hundred mails, sorry.

src/Reflection/Adapter/ReflectionIntersectionType.php
```diff
@@ -7,24 +7,25 @@
...
+use function assert;
 
 class ReflectionIntersectionType extends CoreReflectionIntersectionType
 {
     public function __construct(private BetterReflectionIntersectionType $betterReflectionType)
     {
     }
 
-    /** @return array<ReflectionNamedType> */
+    /** @return list<ReflectionNamedType> */
```
Ocramius: Can we perhaps make this a `non-empty-list`? :thinking:

src/Reflection/Adapter/ReflectionType.php
```diff
@@ -7,20 +7,18 @@
...
 
 use function array_filter;
 use function array_values;
+use function assert;
 use function count;
 
 abstract class ReflectionType extends CoreReflectionType
 {
-    public static function fromTypeOrNull(BetterReflectionNamedType|BetterReflectionUnionType|BetterReflectionIntersectionType|null $betterReflectionType): ReflectionUnionType|ReflectionNamedType|ReflectionIntersectionType|null
+    public static function fromType(BetterReflectionNamedType|BetterReflectionUnionType|BetterReflectionIntersectionType|null $betterReflectionType): ReflectionUnionType|ReflectionNamedType|ReflectionIntersectionType
```
Ocramius: According to infection, this method is not used in a `public` scope? :thinking:

src/Reflection/Adapter/ReflectionUnionType.php
```diff
@@ -5,26 +5,27 @@
...
+use function assert;
 
 final class ReflectionUnionType extends CoreReflectionUnionType
 {
     public function __construct(private BetterReflectionUnionType $betterReflectionType)
     {
     }
 
-    /** @return array<ReflectionNamedType> */
+    /** @return list<ReflectionNamedType|ReflectionIntersectionType> */
```
Ocramius: Can we make this a `non-empty-list`?

src/Reflection/ReflectionParameter.php
```diff
@@ -442,6 +442,10 @@ public function getClass(): ReflectionClass|null
 
         if ($type instanceof ReflectionUnionType) {
             foreach ($type->getTypes() as $innerType) {
+                if (! $innerType instanceof ReflectionNamedType) {
+                    continue;
+                }
```
Ocramius, line 447: Looks quite broken: perhaps we should drop this method completely from `src/Reflection`, and only keep it on adapters? :thinking:

src/Reflection/ReflectionUnionType.php
```diff
@@ -4,35 +4,38 @@
...
 use function array_map;
 use function array_values;
 use function assert;
 use function implode;
+use function sprintf;
 
 class ReflectionUnionType extends ReflectionType
 {
-    /** @var list<ReflectionNamedType> */
+    /** @var list<ReflectionNamedType|ReflectionIntersectionType> */
```
Ocramius: We should probably have `non-empty-list` here

stubs/ReflectionIntersectionType.php
```diff
@@ -6,7 +6,7 @@
 
 class ReflectionIntersectionType extends ReflectionType
 {
-    /** @return ReflectionType[] */
+    /** @return list<ReflectionType> */
```
Ocramius: Hmm, if we use `non-empty-list`, this stub will need adjusting

Ocramius (changes requested):
Gave it another read - found some types that we can refine a bit :+1:

Ocramius:
Sorry for not merging yet: this is really messy PHP stuff, and I've been through LOADS of meetings today @_@

src/Reflection/Adapter/ReflectionType.php
kukulich: It's used in `ReflectionIntersectionType` and `ReflectionUnionType` adapters so marked as `@internal`

src/Reflection/Adapter/ReflectionIntersectionType.php
kukulich: All `non-empty-list` added.

src/Reflection/ReflectionParameter.php
kukulich: It's good idea, the method is marked as deprecated in PHP. However I prefer to do it in separate PR because the method is used in some unit tests.

kukulich:
> Sorry for not merging yet: this is really messy PHP stuff, and I've been through LOADS of meetings today @_@

No problem :)

src/Reflection/Adapter/ReflectionUnionType.php
Ocramius: This was marked as resolved, I think? Not feasible?

src/Reflection/Adapter/ReflectionType.php
```diff
@@ -7,20 +7,19 @@
...
 use function array_filter;
 use function array_values;
+use function assert;
 use function count;
 
 abstract class ReflectionType extends CoreReflectionType
 {
-    public static function fromTypeOrNull(BetterReflectionNamedType|BetterReflectionUnionType|BetterReflectionIntersectionType|null $betterReflectionType): ReflectionUnionType|ReflectionNamedType|ReflectionIntersectionType|null
+    /** @internal */
+    public static function fromType(BetterReflectionNamedType|BetterReflectionUnionType|BetterReflectionIntersectionType|null $betterReflectionType): ReflectionUnionType|ReflectionNamedType|ReflectionIntersectionType
```
Ocramius: How do we kill this mutant, before merging? 🤔

src/Reflection/Adapter/ReflectionUnionType.php
kukulich: I was just blind. Fixed.

src/Reflection/Adapter/ReflectionType.php
kukulich: Added tests but the mutant is still there 🤷‍♂️

Ocramius (approved):
LGTM! Mutation is sub-optimal, but I don't want to waste your time any further :)

src/Reflection/Adapter/ReflectionType.php
Ocramius: Let's roll with the current solution for now :+1:

PR #1205, opened 2022-09-10
PHP 8.2: Constants in traits

kukulich:
https://wiki.php.net/rfc/constants_in_traits


Ocramius (approved):
Ah, one more reason to discourage traits :D 

The complexity is mind-boggling as usual. Glad I voted NO on the RFC: https://wiki.php.net/rfc/constants_in_traits#vote

Thanks @kukulich!

PR #1213, opened 2022-09-14
Removed `ReflectionParameter::getAst()` to prevent memory leaks

kukulich:
Part of https://github.com/Roave/BetterReflection/issues/1073

Not sure about the solution of `getStartColumn()` and `getEndColumn()`. We need it because it's not possible to get columns for magically added methods:

https://github.com/Roave/BetterReflection/blob/5a6ccc032eab2aec577f1abde8d9de1a86bee518/src/Reflection/ReflectionClass.php#L513-L523

### DIFF
--- a/test/unit/Reflection/ReflectionParameterTest.php
+++ b/test/unit/Reflection/ReflectionParameterTest.php
@@ -7,13 +7,15 @@ namespace Roave\BetterReflectionTest\Reflection;
 use InvalidArgumentException;
 use LogicException;
 use OutOfBoundsException;
-use PhpParser\Node\Param;
+use PhpParser\Node;
 use PHPUnit\Framework\TestCase;
 use Roave\BetterReflection\Reflection\Exception\Uncloneable;
+use Roave\BetterReflection\Reflection\ReflectionFunction;
 use Roave\BetterReflection\Reflection\ReflectionParameter;
 use Roave\BetterReflection\Reflector\DefaultReflector;
 use Roave\BetterReflection\Reflector\Reflector;
 use Roave\BetterReflection\SourceLocator\Ast\Locator;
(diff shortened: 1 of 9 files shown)


Ocramius:
> Not sure about the solution of `getStartColumn()` and `getEndColumn()`. We need it because it's not possible to get columns for magically added methods:

Do you have a test that breaks here? 🤔

kukulich:
> Do you have a test that breaks here? 🤔

Added two more tests: https://github.com/Roave/BetterReflection/pull/1213/commits/617231323d89e85e96c5c5d38566d9f6d2ace047#diff-fbb6d99cc76c50fb4c3e48b8e4c69aa36448b68086d49b507068d08c44bab3faR578-R641

test/unit/Reflection/ReflectionParameterTest.php
```diff
@@ -571,18 +575,70 @@ public function testGetStartColumnAndEndColumn(string $php, int $startColumn, in
...
 
-        $ast = $parameter->getAst();
+        $classReflection     = $reflector->reflectClass(StringEnum::class);
+        $methodReflection    = $classReflection->getMethod('tryFrom');
+        $parameterReflection = $methodReflection->getParameter('value');
 
-        self::assertInstanceOf(Param::class, $ast);
-        self::assertSame('boo', $ast->var->name);
+        self::expectException(RuntimeException::class);
+        $parameterReflection->getEndColumn();
```
Ocramius, line 641: I think this is an acceptable behavior for now: the `RuntimeException` is not super nice, but perhaps we can document that `getEndColumn()` throws? 🤔

kukulich: Added.

Ocramius (approved):
Thanks @kukulich!

PR #1214, opened 2022-09-20
Removed some node references to prevent memory leaks

kukulich:
Another part of https://github.com/Roave/BetterReflection/issues/1073

### DIFF
--- a/src/Reflection/ReflectionAttribute.php
+++ b/src/Reflection/ReflectionAttribute.php
@@ -11,19 +11,35 @@ use Roave\BetterReflection\NodeCompiler\CompilerContext;
 use Roave\BetterReflection\Reflection\StringCast\ReflectionAttributeStringCast;
 use Roave\BetterReflection\Reflector\Reflector;
 
+use function array_map;
+
 class ReflectionAttribute
 {
+    /** @var non-empty-string */
+    private string $name;
+
+    /** @var array<int|string, Node\Expr> */
+    private array $arguments = [];
+
     public function __construct(
         private Reflector $reflector,
-        private Node\Attribute $node,
+        Node\Attribute $node,
         private ReflectionClass|ReflectionMethod|ReflectionFunction|ReflectionClassConstant|ReflectionEnumCase|ReflectionProperty|ReflectionParameter $owner,
         private bool $isRepeated,
     ) {
+        /** @var non-empty-string $name */
+        $name       = $node->name->toString();
+        $this->name = $name;
+
+        foreach ($node->args as $argNo => $arg) {
+            $this->arguments[$arg->name?->toString() ?? $argNo] = $arg->value;
+        }
     }
 
+    /** @return non-empty-string */
     public function getName(): string
     {
-        return $this->node->name->toString();
+        return $this->name;
     }
 
     public function getClass(): ReflectionClass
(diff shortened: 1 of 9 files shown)


src/Reflection/ReflectionAttribute.php
```diff
@@ -11,19 +11,31 @@
 use Roave\BetterReflection\Reflection\StringCast\ReflectionAttributeStringCast;
 use Roave\BetterReflection\Reflector\Reflector;
 
+use function array_map;
+
 class ReflectionAttribute
 {
+    private string $name;
```
Ocramius, line 19: Should we have `non-empty-string` on this?

src/Reflection/ReflectionAttribute.php
```diff
@@ -11,19 +11,31 @@
...
+
+    /** @var array<int|string, Node\Expr> */
```
Ocramius, line 21: Is mixing `int` and `string` arguments allowed?

Wondering if this could be `array<int, Node\Expr>|array<string, Node\Expr>` :thinking:

src/Reflection/ReflectionAttribute.php
```diff
@@ -11,19 +11,31 @@
...
         private bool $isRepeated,
     ) {
+        $this->name = $node->name->toString();
+
+        foreach ($node->args as $argNo => $arg) {
+            $this->arguments[$arg->name?->toString() ?? $argNo] = $arg->value;
+        }
     }
 
     public function getName(): string
```
Ocramius, line 40: Can we somehow declare `non-empty-string` from here?

src/Reflection/ReflectionConstant.php
```diff
@@ -25,19 +25,50 @@
...
+
+    private string $docComment;
+
+    private int $startLine;
+
+    private int $endLine;
+
+    private int $startColumn;
+
+    private int $endColumn;
```
Ocramius: We should declare types like `non-empty-string` and `positive-int` where possible, on these

src/Reflection/ReflectionEnumCase.php
```diff
@@ -22,13 +23,38 @@
...
+
+    private string $docComment;
+
+    private int $startLine;
+
+    private int $endLine;
+
+    private int $startColumn;
+
+    private int $endColumn;
```
Ocramius: We should declare stricter types like `positive-int` or `non-empty-string` for these

src/Reflection/ReflectionAttribute.php
kukulich: I think `array<int|string, Node\Expr>` is the most accurate for code like this:

```
#[Attr('arg1', 'arg2', arg3: self::class, arg4: [0, 1]])]
```

Ocramius: Works for me :+1:

src/Reflection/ReflectionAttribute.php
kukulich: Done.

src/Reflection/ReflectionAttribute.php
kukulich: Done.

src/Reflection/ReflectionConstant.php
kukulich: Done.

src/Reflection/ReflectionEnumCase.php
kukulich: Done.

kukulich:
I've also added strict types to `ReflectionParameter` (properties added in #1213) and added `getStartLine()` and `getEndLine()` methods - I really don't know why they were missing 😄

src/Reflection/ReflectionEnumCase.php
```diff
@@ -22,13 +23,52 @@
...
+
     private CompiledValue|null $compiledValue = null;
 
     private function __construct(
         private Reflector $reflector,
-        private EnumCase $node,
+        EnumCase $node,
         private ReflectionEnum $enum,
     ) {
+        /** @var non-empty-string $name */
```
Ocramius, line 55: Hmm, can our CS tooling replace some of these with assertions?

Good enough for now, just don't like having a leap of faith

Ocramius (approved):

src/Reflection/ReflectionEnumCase.php
kukulich: Probably in the next release, I know the author 😄

PR #1216, opened 2022-09-22
Removed `ReflectionProperty::getAst()` to prevent memory leaks

kukulich:
Another part of #1073

I made `ReflectionProperty` cloneable because it was the most easy way. I've found issue https://github.com/Roave/BetterReflection/issues/61 but I'm not sure if it's still relevant. Is it not enough that adapters are not cloneable ?

Btw there are already some classes that are cloneable, eg. `ReflectionConstant` or `ReflectionEnumCase`.

### DIFF
--- a/src/Reflection/ReflectionProperty.php
+++ b/src/Reflection/ReflectionProperty.php
@@ -8,7 +8,6 @@ use Closure;
 use Error;
 use OutOfBoundsException;
 use PhpParser\Node;
-use PhpParser\Node\NullableType;
 use PhpParser\Node\Stmt\Property as PropertyNode;
 use ReflectionException;
 use ReflectionProperty as CoreReflectionProperty;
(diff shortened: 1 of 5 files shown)


src/Reflection/ReflectionProperty.php
```diff
@@ -84,6 +155,14 @@ public static function createFromInstance(object $instance, string $propertyName
         return $property;
     }
 
+    public static function cloneWithImplementingClass(self $property, ReflectionClass $implementingClass): self
```
Ocramius: Could be `withImplementingClass()`.

Should probably be `@internal`

src/Reflection/ReflectionProperty.php
```diff
@@ -475,4 +564,23 @@ private function assertObject(mixed $object): object
 
         return $object;
     }
+
+    private function setModifiers(PropertyNode $node): void
```
Ocramius: Probably best to name this `computeModifiers()`, and avoid the state mutation

src/Reflection/ReflectionProperty.php
```diff
@@ -37,17 +37,88 @@
...
+    private int|null $startLine;
+
+    /** @var positive-int|null */
+    private int|null $endLine;
+
+    /** @var positive-int|null */
+    private int|null $startColumn;
+
+    /** @var positive-int|null */
+    private int|null $endColumn;
```
Ocramius, line 72: Can we mark some of these read-only? :thinking:

Ocramius (changes requested):

Ocramius:
> I made `ReflectionProperty` cloneable because it was the most easy way. I've found issue #61 but I'm not sure if it's still relevant. Is it not enough that adapters are not cloneable ?

Yeah, I think the restriction was in place due to shared mutable internal state: removal of the AST should fix that.

src/Reflection/ReflectionProperty.php
kukulich: Fixed.

src/Reflection/ReflectionProperty.php
kukulich: Fixed.

src/Reflection/ReflectionProperty.php
kukulich: With `@readonly` ?

Ocramius: Ah, forgot, we're still on PHP 8.0.

Thought `readonly` from PHP 8.1.

Ocramius (approved):
LGTM :ship: 

Thanks @kukulich!

PR #1218, opened 2022-09-22
Removed `ReflectionClassConstant::getAst()` to prevent memory leaks

kukulich:
Other part of #1073


Ocramius (approved):
Simple/clean. Thanks @kukulich!

PR #1219, opened 2022-09-22
`Reflection*::getNamespaceName()` returns null when no namespace - namespace as empty string is implemented only in adapters

kukulich:
I think the compatibility can be only in adapters.


Ocramius:
Can't add more BC breaks to the milestone - wow Github, really? :D

kukulich:
Mutation tests are green 🍾

Ocramius (approved):
:ship: thanks! :tada:

PR #1223, opened 2022-09-23
Added methods to get value expression

kukulich:


src/Reflection/ReflectionEnumCase.php
```diff
@@ -86,6 +86,16 @@ public function getName(): string
         return $this->name;
     }
 
+    /** @throws LogicException */
+    public function getValueExpression(): Node\Expr
+    {
+        if ($this->value === null) {
+            throw new LogicException('This enum case does not have a value');
+        }
```
Ocramius: Hmm, this API is a bit "meh".

How does a consumer prevent calling this upfront, if no value exists?

Ocramius (changes requested):

src/Reflection/ReflectionEnumCase.php
kukulich: I did the same as https://github.com/Roave/BetterReflection/blob/6.0.x/src/Reflection/ReflectionEnumCase.php#L100

Consumer should probably check `ReflectionEnum::isBacked()`

Ocramius: Let's add that to the docblock as documentation 👍

kukulich: Added.

Ocramius (approved):
LGTM :+1:

PR #1229, opened 2022-09-23
Removed `ReflectionClass::getAst()` to prevent memory leaks

### LINKED ISSUE #1073: Prevent memory leaks by not holding onto AST nodes

kukulich:
Last part of https://github.com/Roave/BetterReflection/issues/1073

Fixes #1073


Ocramius:
Should we roll 6.0.0 after this?

kukulich:
I think we should give some time to @ondrejmirtes to test it in PHPStan. He may find some problems that will need another BC break.

kukulich:
6.0 will be big release so may I rename `ReflectionClass::getImmediateReflectionConstants()` to `getImmediateConstants()`? To make it compatible with `getImmediateMethods()` etc.

Ocramius:
Works for me, let's just not push the goalpost too far, since "done" is better than perfect, and we're already near damn perfect :D

Ocramius (approved):
Excellent, thanks!

PR #1231, opened 2022-09-24
`ReflectionClass::getConstant()`, `getConstants()` and `getImmediateConstants()` more consistent with methods and properties methods

kukulich:
- Return `ReflectionClassConstant` instances
- Have filter parameter

### DIFF
--- a/src/Reflection/ReflectionClass.php
+++ b/src/Reflection/ReflectionClass.php
@@ -637,9 +637,16 @@ class ReflectionClass implements Reflection
      *
      * @return array<string, ReflectionClassConstant> indexed by name
      */
-    public function getImmediateConstants(): array
+    public function getImmediateConstants(int|null $filter = null): array
     {
-        return $this->immediateConstants;
+        if ($filter === null) {
+            return $this->immediateConstants;
+        }
+
+        return array_filter(
+            $this->immediateConstants,
+            static fn (ReflectionClassConstant $constant): bool => (bool) ($filter & $constant->getModifiers()),
+        );
     }
 
     /**
@@ -683,7 +690,7 @@ class ReflectionClass implements Reflection
      *
      * @return array<string, ReflectionClassConstant> indexed by name
      */
-    public function getConstants(): array
+    public function getConstants(int|null $filter = null): array
     {
         // Note: constants are not merged via their name as array index, since internal PHP constant
         //       sorting does not follow `\array_merge()` semantics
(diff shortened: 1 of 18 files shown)


src/Reflection/ReflectionClass.php
```diff
@@ -633,74 +633,38 @@ public function hasMethod(string $methodName): bool
...
-     * Get the value of the specified class constant.
-     *
-     * Returns null if not specified.
+     * up parent classes etc.), with keys as constant names and values as {@see ReflectionClassConstant} objects.
      *
-     * @return scalar|array<scalar>|null
+     * @return array<string, ReflectionClassConstant> indexed by name
      */
-    public function getConstant(string $name): string|int|float|bool|array|null
+    public function getImmediateConstants(int|null $filter = null): array
```
Ocramius, line 640: If we only stick to `int $filter`, we should document its bitmask type precisely

src/Reflection/ReflectionClass.php
```diff
@@ -633,74 +633,38 @@ public function hasMethod(string $methodName): bool
...
-    }
-
-    /**
-     * Get the value of the specified class constant.
-     *
-     * Returns null if not specified.
+     * up parent classes etc.), with keys as constant names and values as {@see ReflectionClassConstant} objects.
      *
-     * @return scalar|array<scalar>|null
+     * @return array<string, ReflectionClassConstant> indexed by name
```
Ocramius, line 638: `non-empty-string` keys?

src/Reflection/ReflectionClass.php
```diff
@@ -633,74 +633,38 @@ public function hasMethod(string $methodName): bool
...
-    }
-
-    /**
-     * Get an associative array of only the constants for this specific class (i.e. do not search
-     * up parent classes etc.), with keys as constant names and values as {@see ReflectionClassConstant} objects.
-     *
-     * @return array<string, ReflectionClassConstant> indexed by name
-     */
-    public function getImmediateReflectionConstants(): array
+    public function getConstant(string $name): ReflectionClassConstant|null
```
Ocramius, line 665: We should only accept `non-empty-string` as name

src/Reflection/ReflectionClass.php
kukulich: @Ocramius Can I prepare everything in next PR? I will change it for methods and properties as well.

Ocramius: Sure 👍

PR #1233, opened 2022-09-24
`ReflectionClass::getMethod()` returns null when method does not exist

kukulich:
To make it consistent with `getProperty()`, `getConstant()` and `getCase()`


Ocramius (approved):
👍

PR #1234, opened 2022-09-24
Method names, property names etc could be only `non-empty-string`

kukulich:


src/Reflection/ReflectionFunctionAbstract.php
```diff
@@ -182,7 +185,7 @@ public function getNumberOfRequiredParameters(): int
      */
     public function getParameters(): array
     {
-        return $this->parameters;
+        return array_values($this->parameters);
```
Ocramius, line 188: If these are indexed by name, is the order still guaranteed? 🤔

We should at least have a test covering a signature like `foo($c, $a, $b)` to ensure this doesn't change by accident

Ocramius (approved):

src/Reflection/ReflectionFunctionAbstract.php
kukulich: Ou, you're right.

kukulich: Special test added.

PR #1235, opened 2022-09-24
Filter parameter for methods, properties and constants is always `int-mask-of`

kukulich:


src/Reflection/Adapter/ReflectionEnum.php
```diff
@@ -135,13 +134,18 @@ public function getMethod(string $name): ReflectionMethod
...
      * @return list<ReflectionMethod>
      *
      * @psalm-suppress MethodSignatureMismatch
      */
     public function getMethods(int|null $filter = null): array
     {
-        return array_map(static fn (BetterReflectionMethod $method): ReflectionMethod => new ReflectionMethod($method), $this->betterReflectionEnum->getMethods($filter));
+        return array_map(
+            static fn (BetterReflectionMethod $method): ReflectionMethod => new ReflectionMethod($method),
+            $this->betterReflectionEnum->getMethods($filter ?? 0),
```
Ocramius, line 147: Oh, what is `?? 1` here? 🤔

Ocramius (commented):
Loads of new mutations due to `?? 0` 😬

test/unit/Reflection/ReflectionClassTest.php
```diff
@@ -2081,7 +2089,7 @@ public function testGetImmediateConstants(): void
         self::assertSame('ff', $reflectionConstants['F']->getValue());
     }
 
-    /** @return list<array{0: int, 1: int}> */
+    /** @return list<array{0: int-mask-of<CoreReflectionClassConstant::IS_*>, 1: int}> */
     public function getConstantsWithFilterDataProvider(): array
     {
         return [
```
Ocramius, line 2097: Hmm, PHPStan kinda disagreeing here?

kukulich: `ReflectionClassConstant::IS_FINAL` exists since 8.1 that's why the problem is on PHP 8.0. I'm not sure what's the best way how to solve it.

Ocramius: Can we perhaps declare our own constant on the adapter, then use `ReflectionClass::IS_*`?

I think the constants aren't `final`

src/Reflection/Adapter/ReflectionEnum.php
kukulich: Killed them all.

test/unit/Reflection/ReflectionClassTest.php
kukulich: Perfect👍 Done.

Ocramius (approved):
Thanks @kukulich!

PR #1237, opened 2022-09-25
Fixed adapters compatibility

kukulich:
Some are broken since https://github.com/Roave/BetterReflection/pull/1231

### DIFF
--- a/src/Reflection/Adapter/ReflectionClass.php
+++ b/src/Reflection/Adapter/ReflectionClass.php
@@ -164,7 +164,7 @@ final class ReflectionClass extends CoreReflectionClass
         $betterReflectionProperty = $this->betterReflectionClass->getProperty($name);
 
         if ($betterReflectionProperty === null) {
-            throw new CoreReflectionException(sprintf('Property "%s" does not exist', $name));
+            throw new CoreReflectionException(sprintf('Property %s::$%s does not exist', $this->betterReflectionClass->getName(), $name));
         }
 
         return new ReflectionProperty($betterReflectionProperty);
@@ -429,13 +429,13 @@ final class ReflectionClass extends CoreReflectionClass
                 return $default;
             }
 
-            throw new CoreReflectionException(sprintf('Property "%s" does not exist', $name));
+            throw new CoreReflectionException(sprintf('Property %s::$%s does not exist', $this->betterReflectionClass->getName(), $name));
         }
 
         $property = new ReflectionProperty($betterReflectionProperty);
 
         if (! $property->isStatic()) {
-            throw new CoreReflectionException(sprintf('Property "%s" is not static', $name));
+            throw new CoreReflectionException(sprintf('Property %s::$%s does not exist', $this->betterReflectionClass->getName(), $name));
         }
 
         return $property->getValue();
@@ -448,13 +448,13 @@ final class ReflectionClass extends CoreReflectionClass
         $betterReflectionProperty = $this->betterReflectionClass->getProperty($name);
 
         if ($betterReflectionProperty === null) {
-            throw new CoreReflectionException(sprintf('Property "%s" does not exist', $name));
+            throw new CoreReflectionException(sprintf('Class %s does not have a property named %s', $this->betterReflectionClass->getName(), $name));
         }
 
         $property = new ReflectionProperty($betterReflectionProperty);
 
         if (! $property->isStatic()) {
-            throw new CoreReflectionException(sprintf('Property "%s" is not static', $name));
+            throw new CoreReflectionException(sprintf('Class %s does not have a property named %s', $this->betterReflectionClass->getName(), $name));
         }
 
         $property->setValue($value);
--- a/src/Reflection/Adapter/ReflectionClass.php
+++ b/src/Reflection/Adapter/ReflectionClass.php
@@ -129,7 +129,7 @@ final class ReflectionClass extends CoreReflectionClass
         $method = $this->betterReflectionClass->getMethod($name);
 
         if ($method === null) {
-            throw new OutOfBoundsException(sprintf('Could not find method: %s', $name));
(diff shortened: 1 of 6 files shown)


src/Reflection/Adapter/ReflectionClass.php
```diff
@@ -49,7 +48,7 @@ public function __get(string $name): mixed
             return $this->betterReflectionClass->getName();
         }
 
-        throw new OutOfBoundsException(sprintf('Property %s::$%s does not exist.', self::class, $name));
+        return null;
```
Ocramius: We should probably keep the `OutOfBoundsException` for dynamic properties

src/Reflection/Adapter/ReflectionClassConstant.php
```diff
@@ -170,6 +169,6 @@ public function __get(string $name): mixed
             return $this->getDeclaringClass()->getName();
         }
 
-        throw new OutOfBoundsException(sprintf('Property %s::$%s does not exist.', self::class, $name));
+        return null;
```
Ocramius: We should probably keep the `OutOfBoundsException` for dynamic properties

src/Reflection/Adapter/ReflectionEnum.php
```diff
@@ -47,7 +46,7 @@ public function __get(string $name): mixed
             return $this->betterReflectionEnum->getName();
         }
 
-        throw new OutOfBoundsException(sprintf('Property %s::$%s does not exist.', self::class, $name));
+        return null;
```
Ocramius: We should probably keep the `OutOfBoundsException` for dynamic properties

src/Reflection/Adapter/ReflectionEnumBackedCase.php
```diff
@@ -123,6 +121,6 @@ public function __get(string $name): mixed
             return $this->betterReflectionEnumCase->getDeclaringClass()->getName();
         }
 
-        throw new OutOfBoundsException(sprintf('Property %s::$%s does not exist.', self::class, $name));
+        return null;
```
Ocramius: We should probably keep the `OutOfBoundsException` for dynamic properties

src/Reflection/Adapter/ReflectionEnumUnitCase.php
```diff
@@ -118,6 +116,6 @@ public function __get(string $name): mixed
             return $this->betterReflectionEnumCase->getDeclaringClass()->getName();
         }
 
-        throw new OutOfBoundsException(sprintf('Property %s::$%s does not exist.', self::class, $name));
+        return null;
```
Ocramius: We should probably keep the `OutOfBoundsException` for dynamic properties

src/Reflection/Adapter/ReflectionFunction.php
```diff
@@ -251,7 +249,7 @@ public function __get(string $name): mixed
             return $this->betterReflectionFunction->getName();
         }
 
-        throw new OutOfBoundsException(sprintf('Property %s::$%s does not exist.', self::class, $name));
+        return null;
```
Ocramius: We should probably keep the `OutOfBoundsException` for dynamic properties

src/Reflection/Adapter/ReflectionMethod.php
```diff
@@ -333,6 +331,6 @@ public function __get(string $name): mixed
             return $this->betterReflectionMethod->getImplementingClass()->getName();
         }
 
-        throw new OutOfBoundsException(sprintf('Property %s::$%s does not exist.', self::class, $name));
+        return null;
```
Ocramius: We should probably keep the `OutOfBoundsException` for dynamic properties

src/Reflection/Adapter/ReflectionObject.php
```diff
@@ -499,6 +503,6 @@ public function __get(string $name): mixed
             return $this->betterReflectionObject->getName();
         }
 
-        throw new OutOfBoundsException(sprintf('Property %s::$%s does not exist.', self::class, $name));
+        return null;
```
Ocramius: We should probably keep the `OutOfBoundsException` for dynamic properties

src/Reflection/Adapter/ReflectionParameter.php
```diff
@@ -251,6 +249,6 @@ public function __get(string $name): mixed
             return $this->betterReflectionParameter->getName();
         }
 
-        throw new OutOfBoundsException(sprintf('Property %s::$%s does not exist.', self::class, $name));
+        return null;
```
Ocramius: We should probably keep the `OutOfBoundsException` for dynamic properties

src/Reflection/Adapter/ReflectionProperty.php
```diff
@@ -186,6 +185,6 @@ public function __get(string $name): mixed
             return $this->betterReflectionProperty->getImplementingClass()->getName();
         }
 
-        throw new OutOfBoundsException(sprintf('Property %s::$%s does not exist.', self::class, $name));
+        return null;
```
Ocramius: We should probably keep the `OutOfBoundsException` for dynamic properties

Ocramius (changes requested):

src/Reflection/Adapter/ReflectionClass.php
kukulich: I've tried to be compatible with the core reflection: https://3v4l.org/NlhUD

Ocramius: Right, but here we just say `null` and all is good: core at least raises a notice :D 

This is something I tried replicating also in `ocramius/proxy-manager`, but on the long term, it's just a waste of time and a source of bugs: undefined properties are undefined, and we don't need to be compatible with core on it.

A hard crash or a notice are equivalent here: let's have a clear crash.

kukulich: Ok, the commit is away.

Ocramius (approved):
Thanks @kukulich!

PR #1240, opened 2022-09-27
Detection of circular references

### LINKED ISSUE #1165: Better detection of circular references
Hi,

indirectly, via [phpstan user report](https://github.com/phpstan/phpstan/issues/7787), I found out that there is valid PHP code that triggers endless recursion in Better Reflection.
E.g. circular class or interface extends statements like
```
class TestClass extends TestClass {}

class Foo extends Bar {}
class Bar extends Foo {}
```

This code will parse just fine, but lead to errors like `Class "TestClass" not found`, see https://3v4l.org/YIQNl

There are currently 2 cases for that specific example that I found that are problematic
1. `ReflectionClass::getParentClass()` will never return null and consumers can recurse forever. This is kind of expected IMO and fixing it screams BC break or am I wrong?
2. Everything that internally uses `ReflectionClass::getInheritanceClassHierarchy()`, which in term uses getParentClass to find the root parent, leads to endless recursion. E.g. `ReflectionClass::getInterfaces()`, IMO this could be already "fixed" without a BC break? But if getParentClass would be able to deal with it then nothing needs to be done here additionally I assume.

The second one would be easy to prevent in Better Reflection directly, I did not dig deep enough yet to have an idea for the first one. And by preventing I mean something as throwing a `CircularReferenceException`, or do you have any better ideas?

Ondrej also mentioned potential other problematic things via https://github.com/phpstan/phpstan/issues/7787#issuecomment-1214383974 and https://github.com/phpstan/phpstan/issues/7787#issuecomment-1214386506

What do you think? Looking forward to your opinions. Also, I'm happy to work on anything related to this.

kukulich:
Fixes #1165

### DIFF
--- a/src/Reflection/ReflectionClass.php
+++ b/src/Reflection/ReflectionClass.php
@@ -21,6 +21,7 @@ use Roave\BetterReflection\Reflection\Adapter\ReflectionClassConstant as Reflect
 use Roave\BetterReflection\Reflection\Adapter\ReflectionProperty as ReflectionPropertyAdapter;
 use Roave\BetterReflection\Reflection\Annotation\AnnotationHelper;
 use Roave\BetterReflection\Reflection\Attribute\ReflectionAttributeHelper;
+use Roave\BetterReflection\Reflection\Exception\CircularReference;
 use Roave\BetterReflection\Reflection\Exception\ClassDoesNotExist;
 use Roave\BetterReflection\Reflection\Exception\NoObjectProvided;
 use Roave\BetterReflection\Reflection\Exception\NotAClassReflection;
@@ -129,6 +130,9 @@ class ReflectionClass implements Reflection
     private array|null $cachedMethods = null;
 
     private ReflectionClass|null $cachedParentClass = null;
+
+    /** @var list<class-string>|null */
+    private array|null $cachedParentClassNames = null;
 
     /** @internal */
     protected function __construct(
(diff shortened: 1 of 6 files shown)


src/Reflection/ReflectionClass.php
```diff
@@ -1056,7 +1060,28 @@ public function getParentClass(): ReflectionClass|null
...
+                ) {
+                    throw CircularReference::fromClassName($parentClassName);
+                }
+
+                $parentClassNames[] = $parentClassName;
+                $parentClass        = $parentClass->getParentClass();
+            }
+
+            $this->cachedParentClassNames = $parentClassNames;
+        }
```
Ocramius: Can we implement this check without having a mutable `$this->cachedParentClassNames`? Thinking of a by-ref parameter in a `private` method :thinking:

herndlm:
thank you for working on this! You beat me to opening a PR :) Looks like this will make it into 6.0, nice

src/Reflection/ReflectionClass.php
kukulich: Is is such a problem? `ReflectionClass` already have some more caches...

Ocramius: No, it's not, just trying to push the design for less stateful changes: can roll with the current solution too.

In practice, at some point, my aim is to have everything `readonly` :)

Ocramius (approved):
LGTM, nice! :+1:

Ocramius:
Thanks @kukulich @herndlm!

staabm:
Nice. Maybe this is similar to the endless loop we see in https://github.com/phpstan/phpstan/issues/8054 ?

PR #1241, opened 2022-09-27
Detection of interface extends circular references

herndlm:
This was not handled yet in https://github.com/Roave/BetterReflection/pull/1240


Ocramius (approved):
Thanks @herndlm!

PR #907, opened 2021-12-06
Revert "Removed support for PHP4 style constructors"

kukulich:
This reverts commit 8c02766b0098fdaea0ca0773379289dd2c67d89d.

PR #1244, opened 2022-09-28
Compute `$isConstructor` in `ReflectionMethod` constructor

kukulich:


src/Reflection/ReflectionMethod.php
```diff
@@ -336,16 +339,24 @@ public function isStatic(): bool
...
         if ($declaringClass->inNamespace()) {
             return false;
         }
 
-        return strtolower($this->getName()) === strtolower($declaringClass->getShortName());
+        if ($declaringClass->isAnonymous()) {
+            return false;
+        }
+
+        return strtolower($name) === strtolower($declaringClass->getShortName());
```
Ocramius, line 359: I don't think this applies anymore: https://3v4l.org/sGLco

kukulich: There's was one revert already :) https://github.com/Roave/BetterReflection/pull/907

https://github.com/Roave/BetterReflection/issues/904#issuecomment-986588835

@ondrejmirtes Is it still relevant?

ondrejmirtes: You're right, it last worked in PHP 7. (And it never worked in namespaces.)

Ocramius: Let's kill it: we don't support PHP 7.4 anymore.

PR #1247, opened 2022-09-28
Interface can have only public properties

kukulich:
Interface really should not have any properties but we have `UnitEnum` and `BackedEnum` https://github.com/Roave/BetterReflection/pull/971

### DIFF
--- a/src/Reflection/ReflectionClass.php
+++ b/src/Reflection/ReflectionClass.php
@@ -902,7 +902,7 @@ class ReflectionClass implements Reflection
                     [],
                     $this->getParentClass()?->getProperties(ReflectionPropertyAdapter::IS_PUBLIC | ReflectionPropertyAdapter::IS_PROTECTED) ?? [],
                     ...array_map(
-                        static fn (ReflectionClass $ancestor): array => $ancestor->getProperties(ReflectionPropertyAdapter::IS_PUBLIC | ReflectionPropertyAdapter::IS_PROTECTED),
+                        static fn (ReflectionClass $ancestor): array => $ancestor->getProperties(),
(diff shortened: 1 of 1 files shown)


src/Reflection/ReflectionClass.php
```diff
@@ -902,7 +902,7 @@ public function getProperties(int $filter = 0): array
                     [],
                     $this->getParentClass()?->getProperties(ReflectionPropertyAdapter::IS_PUBLIC | ReflectionPropertyAdapter::IS_PROTECTED) ?? [],
                     ...array_map(
-                        static fn (ReflectionClass $ancestor): array => $ancestor->getProperties(ReflectionPropertyAdapter::IS_PUBLIC | ReflectionPropertyAdapter::IS_PROTECTED),
+                        static fn (ReflectionClass $ancestor): array => $ancestor->getProperties(),
```
Ocramius, line 905: Ok, but do we have an accompanying test? 

This sort of change should have raised mutations before :thinking:

kukulich: It should generate mutant here but it does not 🤷‍♂️ https://github.com/Roave/BetterReflection/actions/runs/3143133403/jobs/5107521310

The mutant appears in master https://github.com/Roave/BetterReflection/actions/runs/3144134270/jobs/5109743535

And now should be killed: https://github.com/Roave/BetterReflection/actions/runs/3144233387/jobs/5109970208

Ocramius (approved):
LGTM :ship:

Ocramius:
BTW:

 > 0 covered mutants were not detected

I wonder if we should raise the threshold for preventing **any** mutant from making it through now :thinking:

PR #1248, opened 2022-09-28
Increased minimum covered code mutations core to 100 %

kukulich:


Ocramius:
Nice! This simplifies code reviews from third parties :-)

PR #1250, opened 2022-09-29
PHP 8.2: Property in readonly class is readonly

kukulich:
Found by @ondrejmirtes

### DIFF
--- a/src/Reflection/ReflectionProperty.php
+++ b/src/Reflection/ReflectionProperty.php
@@ -207,7 +207,7 @@ class ReflectionProperty
 
     public function isReadOnly(): bool
     {
-        return $this->node->isReadonly();
+        return $this->node->isReadonly() || $this->getDeclaringClass()->isReadOnly();
     }
 
     public function getDeclaringClass(): ReflectionClass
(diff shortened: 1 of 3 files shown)


kukulich:
I'm not sure if we should release 5.11.1.

Ocramius:
If this is a bugfix, then yes, we should target `5.x`, then merge up

src/Reflection/ReflectionProperty.php
```diff
@@ -207,7 +207,7 @@ public function isInitialized(object|null $object = null): bool
 
     public function isReadOnly(): bool
     {
-        return $this->node->isReadonly();
+        return $this->getDeclaringClass()->isReadOnly() || $this->node->isReadonly();
```
Ocramius: Minor optimization possible by reversing the operand order?

Ocramius (approved):

src/Reflection/ReflectionProperty.php
kukulich: Done.

Ocramius:
Github actions is having a hiccup here: unrelated to this patch, therefore :ship: 

Thanks @kukulich!

PR #1252, opened 2022-09-29
Optimise interface traversion in `ReflectionClass`

herndlm:
Previously it was using `getInterfaces` which determines the parents and their interfaces too. Now it only uses the currents class interfaces since parent constants or props are already explicitly collected.

This saves merely a couple of unnecessary calls, depending on how heavy inheritance is used. Just noticed it while trying to fully understand how constants, properties and methods are collected.


kukulich:
@herndlm Why not methods?

herndlm:
> @herndlm Why not methods?

oh, right. adapted. I initially thought that they do this already, but was mistaken :)

Ocramius (approved):
LGTM 👍

Thanks @herndlm and @kukulich!

Issue #1242, opened 2022-09-27
Detection of trait use circular references


herndlm:
Defining a trait, using itself and also using it in a class leads to endless recursion when methods, properties or constants are queried from the class.
Most likely other circular-trait-usages like a->b->a or so lead to the same.

I should be able to open a PR for this on the weekend the latest. I just created this issue because I was a bit worried since it looks like 6.0 is coming soon :) but I guess this shouldn't be a BC break anyways, would still nice to have it in though

herndlm:
Most likely I'll refactor the way trait hierarchies are collected and then add circular reference protection. Should I split the changes in 2 PRs?

kukulich:
@herndlm I think one PR is ok - maybe two commits. However if the refactoring is needed to implement the circular reference check then one commit should be ok too.

I made some refactoring yesterday too.

Ocramius:
Handled in #1253, #1254, #1256

PR #1253, opened 2022-09-29
Detection of trait use circular references

herndlm:

The solution is very influenced by (= mostly duplicated :P) the one that is used for interfaces. It is working fine and makes sense to me, but I'm not a 100% sure about the method names and how to make clear what each one does and how it is different to just `getTraits()`.

The new exception class is also influenced by similar older ones. Not sure if the same approach should still be used.

### DIFF
--- a/src/Reflection/ReflectionClass.php
+++ b/src/Reflection/ReflectionClass.php
@@ -28,6 +28,7 @@ use Roave\BetterReflection\Reflection\Exception\NoObjectProvided;
 use Roave\BetterReflection\Reflection\Exception\NotAClassReflection;
 use Roave\BetterReflection\Reflection\Exception\NotAnInterfaceReflection;
 use Roave\BetterReflection\Reflection\Exception\NotAnObject;
+use Roave\BetterReflection\Reflection\Exception\NotATraitReflection;
 use Roave\BetterReflection\Reflection\Exception\ObjectNotInstanceOfClass;
 use Roave\BetterReflection\Reflection\Exception\PropertyDoesNotExist;
 use Roave\BetterReflection\Reflection\StringCast\ReflectionClassStringCast;
@@ -384,7 +385,7 @@ class ReflectionClass implements Reflection
                         ),
                     );
                 },
-                $this->getTraits(),
+                $this->getFlattenedUsedTraits(),
             ),
         );
     }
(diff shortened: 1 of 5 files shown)


src/Reflection/ReflectionClass.php
```diff
@@ -1618,6 +1619,53 @@ private function getInterfacesHierarchy(array &$interfaceClassNames = []): array
...
+    }
+
+    /**
+     * This method allows us to retrieve all traits used by this trait. Do not use on class nodes!
+     *
+     * @param list<trait-string> $traitClassNames
+     *
+     * @return list<ReflectionClass>
+     */
+    private function getTraitsHierarchy(array &$traitClassNames = []): array
```
herndlm: ah, this name is also weird. it's not a hierarchy. maybe more like `getFlattenedTraits` or so? The only difference between this and the other new method is then only that one works only on traits and includes the current trait and the other works on everything and does not include the current trait 🤔

herndlm:
Thinking this through a second time let me realise that now it does do unnecessary calls if e.g. a class A uses trait T1 that uses trait T2. In the first iteration it would already get bot T1 and T2, but when going over T1 to e.g. get constants it finds T2 again and gets the constants again :/

Maybe I should just ensure that there are no trait circular references via helper method instead of doing this flatten thing too.

herndlm:
Not happy with this, I'll try another solution

kukulich:
@herndlm Btw I think we should refactor `getMethodsIndexedByLowercasedName()` (see #1255) because it's very inefective. 

```
        $classMethods     = $this->getImmediateMethods();
        $className        = $this->getName();
        $parentMethods    = $this->getParentMethods();
        $traitsMethods    = $this->getMethodsFromTraits();
        $interfaceMethods = $this->getMethodsFromInterfaces();
```

We should not call all of this methods because some calls may be useless.

herndlm:
> We should not call all of this methods because some calls may be useless.

do you have more in mind already? I see that those methods at least should only deal with the current class and not parents too.

Cleaning up there would definitely also help me add the circular ref protection. your linked PR already does the first step :)

kukulich:
@herndlm Eg. `$parentMethods = $this->getParentMethods()` and call `$method->withCurrentClass()` for all parent methods is useless when method already exist in current class.

PR #1254, opened 2022-09-29
Detection of trait use circular references [solution 2]

herndlm:

Simpler alternative to https://github.com/Roave/BetterReflection/pull/1253. This one does more trait recursion checks.

The only alternative, I can think of, that avoids any overhead would be to directly build such checks into `getConstants`, `getMethods` and `getProperties` or respective internal versions of those. I was just worried that it might add too much complexity just for this.. Well, maybe I'll open one more PR tomorrow xD


herndlm:
Not happy with this, I'll try another solution

PR #1255, opened 2022-09-30
Mark `ReflectionClass::getMethods()` compatible with other similar getters

kukulich:


Ocramius (approved):
Even simplifies the tests, nice :+1:

PR #1256, opened 2022-09-30
Detection of circular references in constant, method and property getters

herndlm:

3 times the charm :)

This one should be better. It should not add performance overhead. But it adds a bit of complexity of course and those private helper methods to "hide" the classNameStack reference from the API.

### DIFF
--- a/src/Reflection/ClassNameStack.php
+++ b/src/Reflection/ClassNameStack.php
@@ -1,0 +1,40 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Roave\BetterReflection\Reflection;
+
+use Roave\BetterReflection\Reflection\Exception\CircularReference;
+
+use function array_key_exists;
+
+/**
+ * @internal
+ *
+ * @psalm-immutable
+ */
+final class ClassNameStack
+{
+    /** @param array<class-string, null> $classNames */
+    private function __construct(private array $classNames)
+    {
+    }
+
+    public static function createEmpty(): self
+    {
+        return new self([]);
+    }
+
+    /** @param class-string $className */
+    public function push(string $className): self
+    {
+        if (array_key_exists($className, $this->classNames)) {
+            throw CircularReference::fromClassName($className);
+        }
+
+        $classNames             = $this->classNames;
+        $classNames[$className] = null;
+
+        return new self($classNames);
+    }
+}
--- a/src/Reflection/ReflectionClass.php
+++ b/src/Reflection/ReflectionClass.php
@@ -714,7 +714,7 @@ class ReflectionClass implements Reflection
      */
     public function getConstants(int $filter = 0): array
     {
-        return $this->getConstantsHelper($filter);
+        return $this->getConstantsConsideringAlreadyVisitedClasses($filter);
     }
 
     /**
@@ -723,7 +723,7 @@ class ReflectionClass implements Reflection
      *
      * @return array<non-empty-string, ReflectionClassConstant> indexed by name
      */
-    private function getConstantsHelper(int $filter = 0, array &$classNameStack = []): array
+    private function getConstantsConsideringAlreadyVisitedClasses(int $filter = 0, array &$classNameStack = []): array
     {
         $classNameStack = $this->pushClassNameStack($classNameStack, $this->getName());
 
@@ -732,18 +732,18 @@ class ReflectionClass implements Reflection
             //       sorting does not follow `\array_merge()` semantics
             $constants = array_merge(
                 array_values($this->getImmediateConstants()),
-                array_values($this->getParentClass()?->getConstantsHelper(ReflectionClassConstantAdapter::IS_PUBLIC | ReflectionClassConstantAdapter::IS_PROTECTED, $classNameStack) ?? []),
+                array_values($this->getParentClass()?->getConstantsConsideringAlreadyVisitedClasses(ReflectionClassConstantAdapter::IS_PUBLIC | ReflectionClassConstantAdapter::IS_PROTECTED, $classNameStack) ?? []),
                 ...array_map(
                     function (ReflectionClass $trait) use ($classNameStack): array {
                         return array_map(
                             fn (ReflectionClassConstant $classConstant): ReflectionClassConstant => $classConstant->withImplementingClass($this),
-                            $trait->getConstantsHelper(0, $classNameStack),
+                            $trait->getConstantsConsideringAlreadyVisitedClasses(0, $classNameStack),
                         );
                     },
                     $this->getTraits(),
                 ),
                 ...array_map(
-                    static fn (ReflectionClass $interface): array => array_values($interface->getConstantsHelper(0, $classNameStack)),
+                    static fn (ReflectionClass $interface): array => array_values($interface->getConstantsConsideringAlreadyVisitedClasses(0, $classNameStack)),
                     array_values($this->getCurrentClassImplementedInterfacesIndexedByName()),
                 ),
             );
@@ -927,7 +927,7 @@ class ReflectionClass implements Reflection
      */
     public function getProperties(int $filter = 0): array
     {
-        return $this->getPropertiesHelper($filter);
+        return $this->getPropertiesConsideringAlreadyVisitedClasses($filter);
     }
 
     /**
@@ -936,7 +936,7 @@ class ReflectionClass implements Reflection
      *
      * @return array<non-empty-string, ReflectionProperty>
      */
-    private function getPropertiesHelper(int $filter = 0, array &$classNameStack = []): array
+    private function getPropertiesConsideringAlreadyVisitedClasses(int $filter = 0, array &$classNameStack = []): array
     {
         $classNameStack = $this->pushClassNameStack($classNameStack, $this->getName());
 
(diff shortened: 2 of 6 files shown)


src/Reflection/ReflectionClass.php
```diff
@@ -360,27 +360,35 @@ private function createMethodsFromTrait(ReflectionMethod $method): array
...
     }
 
-    /** @return list<ReflectionMethod> */
-    private function getParentMethods(): array
+    /**
+     * @param list<class-string> $classNameStack
+     *
+     * @return list<ReflectionMethod>
+     */
+    private function getParentMethods(array &$classNameStack = []): array
```
Ocramius: holy guacamole by-ref stack passing: that is a lot of methods with by-ref added :O

src/Reflection/ReflectionClass.php
```diff
@@ -1588,39 +1630,52 @@ private function getCurrentClassImplementedInterfacesIndexedByName(): array
...
+    private function pushClassNameStack(array $classNameStack, string $className): array
+    {
+        if (in_array($className, $classNameStack, true)) {
+            throw CircularReference::fromClassName($className);
+        }
+
+        $classNameStack[] = $className;
+
+        return $classNameStack;
+    }
```
Ocramius: Minor improvement: can we make this `array<class-string, null>` and prevent usage of `in_array()` (which does a linear scan)?

src/Reflection/ReflectionClass.php
```diff
@@ -698,23 +714,36 @@ private function createImmediateConstants(ClassNode|InterfaceNode|TraitNode|Enum
...
+        return $this->getConstantsHelper($filter);
+    }
+
+    /**
+     * @param int-mask-of<ReflectionClassConstantAdapter::IS_*> $filter
+     * @param list<class-string>                                $classNameStack
+     *
+     * @return array<non-empty-string, ReflectionClassConstant> indexed by name
+     */
+    private function getConstantsHelper(int $filter = 0, array &$classNameStack = []): array
```
Ocramius: `Helper` doesn't say anything: can this be renamed? What does it **do**? :P

Ocramius: `doGetConstants()` would already be better than `*Helper` BTW

Ocramius: Perhaps `getConstantsConsideringAlreadyVisitedClasses()`?

Ocramius (changes requested):
Minor optimizations feasible, approach seems correct :+1:

src/Reflection/ReflectionClass.php
herndlm: yeah, it did hurt a bit typing out "helper" tbh :D

src/Reflection/ReflectionClass.php
herndlm: yeah :/ I could reduce them a bit at least

Ocramius: Oh, sorry, it was more of a comment on the complexity: I think you may actually have broken it, by removing the by-ref in 04cfc28

herndlm: no worries, you're right. interestingly tests are still green (I just added more to be sure) but indeed those should be refs. I kind of removed them too quickly because that fixed psalm errors.. psalm seems to be "loosing" the type when adding e.g. `use (&$classNameStack)` to closures. I'm not used to psalm, can you quickly point me to something that explains this maybe?

Ocramius: Ok, so by-ref types need a `@psalm-param-out` for this sort of stuff, AFAIK.

An alternative to this design can be achieved by having a small mutable `@internal` data structure: that removes `&` from all this code :)

herndlm: hmm I need to think a bit more about this. It can get messy quite fast, in theory I'd need to convert multiple arrow to anonymous functions, just to use by-ref variables inside, if I'm not mistaken. Also in e.g. https://github.com/Roave/BetterReflection/pull/1241/files#diff-c58f24d134490b09a18aa65366bc9ab2d5bf2dbf493bea9f12b365807a5b5e22R1628 I guess.

ondrejmirtes: My 2 cents - I really dislike references for this purpose, a property is much more clear: https://github.com/phpstan/phpstan-src/commit/33771e5d8ee8b6dbae4f42ce3f8b424a81f68dc9

herndlm: the small mutable internal data structure sounds like a good combination of both worlds. but in the simplest form it would just be an array property I guess? :) Or should I create something like a simple immutable ClassNameStack VO maybe? Anyways, I'll play around with this a bit later, sounds like we can get rid of the references then completely I guess.

Ocramius: An array property suffices 👍

herndlm: I was not thinking this through enough. The property on `ClassReflection` won't be of much help, since the code is creating new ClassReflection instances when recursing which would reset the property, except I preserve this somehow. The use-case is a bit different to what Ondrej posted I think. 🤔
I pushed a commit that passes around a simple immutable object instead, but it feels just like working-around the by-ref thing, doesn't it? 😅 Sorry that this turned out more complex than initially thought.

test/unit/Reflection/ReflectionClassTest.php
```diff
@@ -2661,4 +2664,86 @@ class FooBar extends Bar
         $fooBarDoFooMethod = $fooBar->getMethod('doFoo');
         self::assertTrue($fooBarDoFooMethod->isPublic());
     }
+
```
Ocramius, line 2667: Possibly need to add `@covers Roave\BetterReflection\Reflection\ClassNameStack` here

src/Reflection/ClassNameStack.php
```diff
@@ -0,0 +1,40 @@
...
+use Roave\BetterReflection\Reflection\Exception\CircularReference;
+
+use function array_key_exists;
+
+/**
+ * @internal
+ *
+ * @psalm-immutable
+ */
+final class ClassNameStack
```
Ocramius: Let's call this `AlreadyVisitedClasses`

src/Reflection/ClassNameStack.php
```diff
@@ -0,0 +1,40 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Roave\BetterReflection\Reflection;
```
Ocramius: Let's put this in a `\Support` sub-namespace?

src/Reflection/ReflectionClass.php
Ocramius: It is working around the by-ref, but this design is much cleaner, and exactly what was suggested above :+1:

src/Reflection/ClassNameStack.php
```diff
@@ -0,0 +1,40 @@
...
+
+use Roave\BetterReflection\Reflection\Exception\CircularReference;
+
+use function array_key_exists;
+
+/**
+ * @internal
+ *
+ * @psalm-immutable
```
Ocramius: Let's remove `@psalm-immutable` and use mutability here: much less work for the garbage collector.

test/unit/Reflection/ReflectionClassTest.php
herndlm: can also create a dedicated test file if you want. I totally missed that

Ocramius: Either way works :) 

Anyway, please remove the immutability first: `push()` can be `: void` there, and that removes a lot of complexity in this specific case

src/Reflection/ClassNameStack.php
herndlm: adapted, making this mutable revealed something I did not expect: Code that gets the parent classes or all interfaces already deals with detecting circular references. That means I had to create new `AlreadyVisitedClasses` instances for those things when getting constants, methods or properties. Not really nice I think. It partly might also show that classes are traversed multiple times by design and things can still be optimised maybe

herndlm: Because of this I could remove the `AlreadyVisitedClasses` param from `getParentMethods()` and `getMethodsFromInterfaces()` again I guess hmm

src/Reflection/ReflectionClass.php
```diff
@@ -361,26 +362,26 @@ private function createMethodsFromTrait(ReflectionMethod $method): array
     }
 
     /** @return list<ReflectionMethod> */
-    private function getParentMethods(): array
+    private function getParentMethods(AlreadyVisitedClasses $classNameStack): array
```
herndlm: I'd like to quickly rename those parameters to `$alreadyVisitedClasses` if that's OK

src/Reflection/ReflectionClass.php
```diff
@@ -415,17 +416,19 @@ private function getMethodsFromInterfaces(): array
...
         }
 
         $classMethods     = $this->getImmediateMethods();
         $className        = $this->getName();
-        $parentMethods    = $this->getParentMethods();
-        $traitsMethods    = $this->getMethodsFromTraits();
-        $interfaceMethods = $this->getMethodsFromInterfaces();
+        $parentMethods    = $this->getParentMethods(AlreadyVisitedClasses::createEmpty());
+        $traitsMethods    = $this->getMethodsFromTraits($alreadyVisitedClasses);
+        $interfaceMethods = $this->getMethodsFromInterfaces(AlreadyVisitedClasses::createEmpty());
```
Ocramius, line 431: Wait... is it intentional to use empty instances for methods and interfaces, and not for traits? :thinking:

herndlm: Unfortunately yes. This is https://github.com/Roave/BetterReflection/pull/1256#discussion_r984661676

herndlm: I remember getting some false positive circular refs with enums I think. I can try finding out why, maybe it leads to some optimisations

Ocramius (approved):
Ok, :ship: for this meanwhile. 

Thanks @herndlm!

herndlm:
also thx for bearing with me here :)
you don't squash commits on merge here, do you? I need to be careful with what I commit, I'm kind of used to mostly squashing them on merge

Ocramius:
Nope, I almost never squash.

src/Reflection/ReflectionClass.php
herndlm: I found two popular cases where interfaces are visited multiple times at the moment via e.g. `getMethods()` which is why I had to start with an empty AlreadyVisitedClass again:

1. an enum like

```php
enum StringEnum: string implements InterfaceForEnum
{
    case ONE
        = 'one';
    case TWO
        = 'two';
}
```
In that case apparently `UnitEnum` is automatically added as interface by PHP. Additionally `BackedEnum` is too. `getMethodsIndexedByLowercasedName` traverses then through the interface hierarchy, to query each interface for methods. It will look at `BackedEnum` and call `getMethodsIndexedByLowercasedName` again which get it's implemented interfaces. Since `BackendEnum` extends `UnitEnum` we end up with that thing again which leads to the circular reference.

2. A class Foo that implements interface A which extends interface B.

```php
interface B {}
interface A extends B {}
class Foo implements A {}
```

A very similar case where the interface hierarchy for Foo contains both interfaces already. And when traversing through them to get e.g. methods it looks at A and sees B again as implemented interface and that causes the circular reference.

I'm not sure yet how, but maybe this can be improved. CC @kukulich

kukulich: @herndlm I'm currently trying some optimization here: https://github.com/kukulich/BetterReflection/tree/optimization

PR #1264, opened 2022-09-30
Stricter parent circular reference checks on constant, method and property getters

herndlm:
Just a follow-up to #1256. In https://github.com/Roave/BetterReflection/pull/1256#discussion_r984903690 I noticed that only traversing interfaces needs an empty `AlreadyVisitedClass` instance right now because of some extends/implements special behaviour.

### DIFF
--- a/src/Reflection/ReflectionClass.php
+++ b/src/Reflection/ReflectionClass.php
@@ -426,7 +426,7 @@ class ReflectionClass implements Reflection
 
         $classMethods     = $this->getImmediateMethods();
         $className        = $this->getName();
-        $parentMethods    = $this->getParentMethods(AlreadyVisitedClasses::createEmpty());
+        $parentMethods    = $this->getParentMethods($alreadyVisitedClasses);
         $traitsMethods    = $this->getMethodsFromTraits($alreadyVisitedClasses);
         $interfaceMethods = $this->getMethodsFromInterfaces(AlreadyVisitedClasses::createEmpty());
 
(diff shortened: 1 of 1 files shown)


kukulich:
@herndlm Do you have a test case?

herndlm:
> @herndlm Do you have a test case?

unfortunately not, I think a circular reference will never be triggered there because `getParentClasses()` already deals with it and that is well tested. I merely wanted to pass through `$alreadyVisitedClasses` everywhere where constants, methods and properties getters do recursion.
I think it would only trigger if some internals change or by (e.g. refactoring) mistake somehow parents are traversed multiple times.

I'd like to do the same thing for the interface path, but that is not possible yet because of what is described in the linked comment.

kukulich:
@herndlm Ok, I will try to find a solution in my `optimization` branch.

kukulich:
@herndlm Not sure if it is possible to solve it because this code is perfectly valid: https://3v4l.org/SJFs2

herndlm:
you mean what I described in the link, right? yeah, I was hoping to somehow reduce recursion iterations, but I also couldn't figure it out yet.

herndlm:
@kukulich to give you maybe input why I wanted to solve that: if you replace `$this->getCurrentClassImplementedInterfacesIndexedByName()` with `$this->getImmediateInterfaces()` in the constant, method and property getters everything except the circular reference detection works at the moment. To fix that I wanted to get rid of `AlreadyVisitedClasses::createEmpty()` for the interface paths in those 3 getters. But this is not possible because of the linked problem (false positive circular references).

maybe you can do something with this info in your optimization branch :)

UPDATE: sorry, I might be wrong, I only ran the `ReflectionClassTest` and not the whole testsuite. this might not work after all. maybe it's not worth further looking into

herndlm:
yeah I think this doesn't make much sense then either..

kukulich:
@herndlm https://github.com/Roave/BetterReflection/commit/85c412b54b3f15b0e055963c012515f0c57ae604

I get rid of some tests because I don't think we need to report circular reference if BR can survive it. That's job of static analysis.

herndlm:
Nice, I had already something very similar yesterday. Including the new method on the support class :)

PR #1265, opened 2022-09-30
`Stringable` interface does not implement `Stringable`

herndlm:
that would be another circular reference again :)

https://3v4l.org/QqBDN

### DIFF
--- a/src/Reflection/ReflectionClass.php
+++ b/src/Reflection/ReflectionClass.php
@@ -932,7 +932,7 @@ class ReflectionClass implements Reflection
                     [],
                     $this->getParentClass()?->getPropertiesConsideringAlreadyVisitedClasses(ReflectionPropertyAdapter::IS_PUBLIC | ReflectionPropertyAdapter::IS_PROTECTED, AlreadyVisitedClasses::createEmpty()) ?? [],
                     ...array_map(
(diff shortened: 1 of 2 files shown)


src/Reflection/ReflectionClass.php
```diff
@@ -1218,7 +1218,7 @@ private function addStringableInterface(array $interfaces): array
         /** @psalm-var class-string $stringableClassName */
         $stringableClassName = Stringable::class;
 
-        if (array_key_exists($stringableClassName, $interfaces)) {
+        if (array_key_exists($stringableClassName, $interfaces) || ($this->isInterface && $this->getName() === $stringableClassName)) {
```
herndlm, line 1221: do I need to check if this is an interface or can this be simplified?

herndlm:
Weird, the escaped mutant should not be caused by my changes 🤔

Ocramius:
It could be that the added conditionals make previous conditionals redundant here :)

kukulich:
Shit. I've merge it to wrong branch.

kukulich:
@Ocramius I'm not sure what's the best way how to solve the merge to wrong branch 😞 Should I open duplicate PR to right branch? Or just cherry-pick the commits?

ondrejmirtes:
I wouldn't worry about it too much - the next release can be 6.1.0 and no one can oppose that.

herndlm:
sorry, I didn't notice either that 6.1.x was the target. I think it was the default branch and I didn't check :/

Ocramius:
No big issue: we'll release `6.1.0` then.

PR #1266, opened 2022-10-01
Optimisations of getting methods, properties and constants in `ReflectionClass`

kukulich:
`foreach` is not so elegant as `array_map` but it should be less memory and time consuming - probably more in real application.

For example:

```php
class ParentClass
{
    public function method() {}
}

class CurrentClass extends ParentClass
{
    public function method() {}
}
```

**Before**

1. We get methods for `CurrentClass` and for `ParentClass` - we even clone the method in `ParentClass` with `withCurrentClass()`
2. Then we merge methods and throw the parent method clone away.

**After**

1. We get methods for `CurrentClass`
2. We don't get the parent method because we already have method with the same name.

### DIFF
--- a/src/Reflection/ReflectionClass.php
+++ b/src/Reflection/ReflectionClass.php
@@ -892,50 +892,63 @@ class ReflectionClass implements Reflection
      */
     public function getProperties(int $filter = 0): array
     {
-        return $this->getPropertiesConsideringAlreadyVisitedClasses($filter, AlreadyVisitedClasses::createEmpty());
-    }
-
-    /**
-     * @param int-mask-of<ReflectionPropertyAdapter::IS_*> $filter
-     *
-     * @return array<non-empty-string, ReflectionProperty>
-     */
-    private function getPropertiesConsideringAlreadyVisitedClasses(int $filter, AlreadyVisitedClasses $alreadyVisitedClasses): array
-    {
-        $alreadyVisitedClasses->push($this->getName());
-
-        if ($this->cachedProperties === null) {
-            // merging together properties from parent class, interfaces, traits, current class (in this precise order)
-            $this->cachedProperties = array_merge(
-                array_merge(
-                    [],
-                    $this->getParentClass()?->getPropertiesConsideringAlreadyVisitedClasses(ReflectionPropertyAdapter::IS_PUBLIC | ReflectionPropertyAdapter::IS_PROTECTED, $alreadyVisitedClasses) ?? [],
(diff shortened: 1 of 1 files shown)


Ocramius:
Can we somehow measure the impact of this? The code is substantially worsened here, so I'm not super keen to do this :D

ondrejmirtes:
I plan to benchmark it in PHPStan.

Ocramius:
Ok, just need some rough feedback here, and then we roll `6.1.0` :)

ondrejmirtes:
I was waiting for the 6.1.x situation to clear, now I can test this again 😊 I'll provide some measurements later today.

Ocramius:
This needs a rebase regardless :)

ondrejmirtes:
Alright, some PHPStan-on-PHPStan benchmarks (single-threaded):

On https://github.com/Roave/BetterReflection/commit/4f9aa70bbd5c786e5dbcb9308494faf0e70ae7db: 486 MB, 52s
On current 6.1.x (https://github.com/Roave/BetterReflection/commit/60791bbf6b721f25845b87a4d3bd790d555a57b1): 486 MB, 50s (consistently a bit faster)
This PR rebased on top of 6.1.x: still waiting for it to finish, it's definitely several times slower 🤐 

So this is unexpected. I'll test it on some small sample with Blackfire to see what gets slower or called more times.

src/Reflection/ReflectionClass.php
```diff
@@ -418,54 +377,69 @@ private function getMethodsFromInterfaces(AlreadyVisitedClasses $alreadyVisitedC
      */
     private function getMethodsIndexedByLowercasedName(AlreadyVisitedClasses $alreadyVisitedClasses): array
     {
-        $alreadyVisitedClasses->push($this->getName());
-
-        if ($this->cachedMethods !== null) {
-            return $this->cachedMethods;
-        }
```
herndlm: aren't you missing this? it currently looks like the cache is only written, but never read

kukulich: @herndlm Yes, it's already fixed :) Probably removed during rebase by accident.

kukulich:
The unplanned slowdown is fixed 🤞

ondrejmirtes:
The new version is working better, it's about 482 MB and 48s, so a little improvement 👍

Ocramius (approved):
Let's roll with this then, :ship: 

Thanks everyone, cutting `6.1.0` now :)

PR #1269, opened 2022-10-02
Fixed detection of interface circular reference

### LINKED ISSUE #1267: Not-circular interface detected as circular
Last minute changes to BetterReflection 6.0 before release around `AlreadyVisitedClasses` brought a bug that was familiar to me.

This situation is now detected as circular but in fact it works in PHP:

```php
<?php
interface CacheableDependencyInterface {
}
interface RefinableCacheableDependencyInterface extends CacheableDependencyInterface {
}
interface AccessibleInterface {
}
interface EntityInterface extends AccessibleInterface, CacheableDependencyInterface, RefinableCacheableDependencyInterface {
}
```

In the previous version that used references I accidentally introduced it to my fork and had to fix it too. This is how I introduced the bug: https://github.com/ondrejmirtes/BetterReflection/commit/258f60ea9a1209dfd2520c056fa3911d4496cf3a#diff-c58f24d134490b09a18aa65366bc9ab2d5bf2dbf493bea9f12b365807a5b5e22L1654-R1673

This is how I had to fix it: https://github.com/ondrejmirtes/BetterReflection/commit/22ae7d2c3ea34cf85703db4225e1d814a013b32d + https://github.com/ondrejmirtes/BetterReflection/commit/8c2480836646c23a3baa1c62aec7c1c256acb096

Because: It turns out it's hard to rewrite an arrow function to equivalent imperative code. Because an arrow function doesn't use outer scope variables by reference, the value is copied when array_map is called. And it's called only once so the callback always uses the initial variable value, it's not updated by reference.

The object version using AlreadyVisitedClasses in Roave/BetterReflection 6.0.0 has the same problem. The reference is now updated each time and breaks analysing legit code.

I'll submit a failing test PR too.

kukulich:
Fixes #1267

### DIFF
--- a/src/Reflection/ReflectionClass.php
+++ b/src/Reflection/ReflectionClass.php
@@ -1628,16 +1628,15 @@ class ReflectionClass implements Reflection
         $interfaceClassName = $this->getName();
         $alreadyVisitedClasses->push($interfaceClassName);
 
+        $alreadyVisitedClassesCopy = clone $alreadyVisitedClasses;
+
         /** @var array<class-string, self> $interfaces */
-        $interfaces = array_merge(
-            [$interfaceClassName => $this],
-            ...array_map(
-                fn (string $interfaceClassName): array => $this->reflector
-                    ->reflectClass($interfaceClassName)
-                    ->getInterfacesHierarchy($alreadyVisitedClasses),
-                $this->implementsClassNames,
-            ),
-        );
+        $interfaces = [$interfaceClassName => $this];
+        foreach ($this->getImmediateInterfaces() as $interface) {
+            foreach ($interface->getInterfacesHierarchy(clone $alreadyVisitedClassesCopy) as $extendedInterfaceName => $extendedInterface) {
+                $interfaces[$extendedInterfaceName] = $extendedInterface;
+            }
+        }
 
         return $this->addStringableInterface($interfaces);
     }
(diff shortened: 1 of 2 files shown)


herndlm:
Thank you @kukulich
I see you also fixed what I did in https://github.com/Roave/BetterReflection/pull/1265 too. Was that part of the problem as well?

kukulich:
@herndlm I was not able to solve it without your `Stringable` fix. I've just force pushed your PR with the mutant fix. I hope it will make your build green and I'll rebase then.

herndlm:
I see, thx. Yeah, no worries about my PR, I was just curious :)

src/Reflection/ReflectionClass.php
```diff
@@ -1628,16 +1628,15 @@ private function getInterfacesHierarchy(AlreadyVisitedClasses $alreadyVisitedCla
...
-                fn (string $interfaceClassName): array => $this->reflector
-                    ->reflectClass($interfaceClassName)
-                    ->getInterfacesHierarchy($alreadyVisitedClasses),
-                $this->implementsClassNames,
-            ),
-        );
+        $interfaces = [$interfaceClassName => $this];
+        foreach ($this->getImmediateInterfaces() as $interface) {
+            foreach ($interface->getInterfacesHierarchy(clone $alreadyVisitedClassesCopy) as $n => $i) {
+                $interfaces[$n] = $i;
```
Ocramius: `$name`, `$interface` :P

Ocramius (approved):
:+1:

PR #1286, opened 2022-10-14
Fixed reflection compatibility when method/property/constant/case name is empty

### LINKED ISSUE #1278: Reflection Adapter vs. non-empty-string
This PR https://github.com/Roave/BetterReflection/pull/1234 made Adapter reflection incompatible with the native PHP reflection. The native reflection returns `false` in `has*` methods if it's asked about nonexistent elements, and throws ReflectionException in `get*` methods: https://3v4l.org/YebV9

But the `assert($name !== '');` calls are unforgiving even for `has*` methods. The current situation isn't great, because these methods have `string` in parameter types but reject `''` at runtime. So currently these problems cannot be detected and avoided with static analysis.

I'd prefer if the Adapter namespace matched the behaviour of native reflection:

For `has*` methods, instead of the assert, do this:

```php
if ($s === '') {
    return false;
}
```

For `get*` methods, throw ReflectionException for empty strings.

Thank you.

kukulich:
Fixes #1278

### DIFF
--- a/src/Reflection/Adapter/ReflectionClass.php
+++ b/src/Reflection/Adapter/ReflectionClass.php
@@ -22,7 +22,6 @@ use ValueError;
 use function array_combine;
 use function array_map;
 use function array_values;
-use function assert;
 use function constant;
 use function func_num_args;
 use function sprintf;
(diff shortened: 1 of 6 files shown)


Ocramius (approved):
Looking good, thanks @kukulich!

PR #1306, opened 2022-11-03
Configure config.platform.php in composer.json

herndlm:
should fix Renovate PRs, see https://github.com/Roave/BetterReflection/pull/1305#issuecomment-1301884370

btw., I'm opening this to basically ask (getting more Marco consulting for free hehehe) :D this is normal Composer behaviour, right? I never had to support multiple PHP versions until recently and it surprised me that running `composer update` with e.g. PHP 8.1 creates a lockfile that is not compatible with the virtual `php` dependency in composer.json (like PHP 8 here) :/

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -63,6 +63,9 @@
         "allow-plugins": {
             "dealerdirect/phpcodesniffer-composer-installer": true,
             "infection/extension-installer": false
+        },
+        "platform": {
+            "php": "8.0.99"
         }
     }
 }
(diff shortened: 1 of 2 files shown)


Ocramius (approved):
Excellent, thanks @herndlm!

Ocramius:
> this is normal Composer behaviour, right?

Yes, composer will pick whatever is compatible with the current range 👍

PR #1315, opened 2022-12-03
Dropped PHP 8.0 support

kukulich:


infection.bootstrap.php
```diff
@@ -2,10 +2,4 @@
...
 declare(strict_types=1);
 
-require __DIR__ . '/stubs/UnitEnum.php';
-require __DIR__ . '/stubs/BackedEnum.php';
-
-require __DIR__ . '/stubs/ReflectionEnum.php';
-require __DIR__ . '/stubs/ReflectionEnumUnitCase.php';
-require __DIR__ . '/stubs/ReflectionEnumBackedCase.php';
-require __DIR__ . '/stubs/ReflectionIntersectionType.php';
+// Nothing currently
```
Ocramius: If unused, kill it 😁

psalm.bootstrap.php
```diff
@@ -2,10 +2,4 @@
...
 declare(strict_types=1);
 
-require __DIR__ . '/stubs/UnitEnum.php';
-require __DIR__ . '/stubs/BackedEnum.php';
-
-require __DIR__ . '/stubs/ReflectionEnum.php';
-require __DIR__ . '/stubs/ReflectionEnumUnitCase.php';
-require __DIR__ . '/stubs/ReflectionEnumBackedCase.php';
-require __DIR__ . '/stubs/ReflectionIntersectionType.php';
+// Nothing currently
```
Ocramius: If unused let's kill it

src/Reflection/Adapter/ReflectionEnumBackedCase.php
```diff
@@ -14,6 +14,7 @@
 use function array_map;
 use function sprintf;
 
+/** @psalm-suppress MissingImmutableAnnotation */
```
Ocramius, line 17: Can we attempt hunting for immutability? 🤔

kukulich: It's probably complicated, we already have https://github.com/Roave/BetterReflection/blob/6.5.x/src/Reflection/Adapter/ReflectionParameter.php#L26 and https://github.com/Roave/BetterReflection/blob/6.5.x/src/Reflection/Adapter/ReflectionNamedType.php#L12

infection.bootstrap.php
kukulich: Done.

psalm.bootstrap.php
kukulich: Done.

src/Reflection/Adapter/ReflectionEnumBackedCase.php
Ocramius: OK, let's roll with this as-is then, thanks :)

Ocramius (approved):
Thanks @kukulich! 

Dropping a release with this, mostly to set a clean baseline for anyone upgrading.

composer.json
```diff
@@ -3,7 +3,7 @@
     "description": "Better Reflection - an improved code reflection API",
     "license": "MIT",
     "require": {
-        "php": "~8.0.12 || ~8.1.0 || ~8.2.0",
```
Ocramius, line 6: Ah, just realized that we didn't update `composer.lock` :| 

Running Renovate with that now.

PR #1320, opened 2022-12-16
100 % mutation score

kukulich:


src/Reflection/ReflectionProperty.php
```diff
@@ -286,6 +286,7 @@ public function isInitialized(object|null $object = null): bool
                 return false;
             }
 
+            // @infection-ignore-all Throw_: False positive, it's tested by testIsInitializedThrowsError()
```
Ocramius: Related: https://github.com/sebastianbergmann/php-code-coverage/pull/969 ?

kukulich: Yes, it looks it's related. So this PR can probably wait for the release :)

Ocramius: Probably worth doing that, ya :+1:

Slamdunk: Butfix released, a rebase now should give 100% MSI

kukulich:
And again one new mutant remaining.

kukulich:
I've just added ignore for this last mutant. I don't know what I should change and how to test it. And I really want to merge this PR and unblock dependencies upgrades 😄 

```
Warning: Escaped Mutant for Mutator "FalseValue":

--- Original
+++ New
@@ @@
     public function stream_open($path, $mode, $options, &$opened_path) : bool
     {
         self::$autoloadLocatedFile = $path;
-        return false;
+        return true;
```

Ocramius (approved):
Thanks @kukulich !

PR #1325, opened 2023-02-03
Fixed `Expecting E_WARNING and E_USER_WARNING is deprecated and will no longer be possible in PHPUnit 10.`

kukulich:


Ocramius (approved):
Thanks @kukulich !

PR #1328, opened 2023-02-08
Fix typo in phpdoc of method that should be `@internal`

shyim:
I am annoyed that my phpstorm autocompletes always `@interal` instead of `@internal`

### DIFF
--- a/src/Reflection/ReflectionAttribute.php
+++ b/src/Reflection/ReflectionAttribute.php
@@ -39,7 +39,7 @@ class ReflectionAttribute
         }
     }
 
-    /** @interal */
+    /** @internal */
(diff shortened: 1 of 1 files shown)


Ocramius (approved):
Thanks @shyim!

shyim:
```
[BC] REMOVED: Method Roave\BetterReflection\Reflection\ReflectionAttribute#withOwner() was removed
[BC] CHANGED: Roave\BetterReflection\Reflection\ReflectionAttribute#withOwner() was marked "@internal"
```

No idea what to do with that 😅

Ocramius:
That one is correct, but can be ignored.

`roave/backward-compatibility-check` is complaining that a method got "removed", because it is now `@internal`, and the tool is technically correct, but nothing we can do about the red build, for now :)

mitelg:
that was the reason why PhpStorm does that? 😅
thanks @shyim I was also annoyed by that the last days 😁

PR #1336, opened 2023-02-27
Add AstHelper utility and corresponding tests

Wuffz:
I created a utility in response to the #1334 issue, hopefully this can be merged, if it needs changes please keep me posted

PR #1337, opened 2023-02-27
Add AstHelper utility and corresponding tests improved

Wuffz:
based on Add AstHelper utility and corresponding tests #1336, but with improved phpstan and psalm results.

### DIFF
--- a/src/Util/AstHelper.php
+++ b/src/Util/AstHelper.php
@@ -1,0 +1,113 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Roave\BetterReflection\Util;
+
+use Closure;
+use PhpParser\Node;
+
+use Roave\BetterReflection\BetterReflection;
+use Roave\BetterReflection\Reflection\Reflection;
+use Roave\BetterReflection\Reflection\ReflectionClass;
+use Roave\BetterReflection\Reflection\ReflectionFunction;
+
+use PhpParser\Node\Expr\Closure as ClosureNode;
+use PhpParser\Node\Expr\ClosureUse as ClosureUseNode;
+use PhpParser\Node\Expr\ArrowFunction as ArrowFunctionNode;
+
+use PhpParser\NodeTraverser;
+use PhpParser\NodeVisitor\FindingVisitor;
+use PhpParser\PrettyPrinter\Standard as StandardPrettyPrinter;
+use PhpParser\PrettyPrinterAbstract;
+use RuntimeException;
+
+final class AstHelper {
+    private BetterReflection $betterReflection;
+
+    public function __construct( ?BetterReflection $betterReflection = null ) {
+        if ( !$betterReflection  ) {
+            $this->betterReflection = new BetterReflection();
+        } else {
+            $this->betterReflection = $betterReflection;
+        }
+    }
+    protected function getBetterReflection(): BetterReflection {
+        return $this->betterReflection;
+    }
+    /**
+     * 
+     * @return Node[]|null $nodes
+     * 
+     */
+    public function getAstForReflection(ReflectionClass|ReflectionFunction $reflection): array|null {
+        $parser = $this->getBetterReflection()->phpParser();
+        $source = $reflection->getLocatedSource();
+        return $parser->parse($source->getSource());
+    }
+    /**
+     * @return Node[]|null $nodes
+     */
+    public function getAstForClass(ReflectionClass $function): array|null {
+        return $this->getAstForReflection($function);
+    }
+    /**
+     * @return Node[]|null $nodes
+     */
+    public function getAstForFunction(ReflectionFunction $function): array|null {
+        return $this->getAstForReflection($function);
+    }
+     /**
+     * @return Node[]|null $nodes
+     */
+    public function getAstForClosure(ReflectionFunction $reflection): array|null {
+        $parser = $this->getBetterReflection()->phpParser();
+        
+        // can this be done better/more efficient using the parser itself?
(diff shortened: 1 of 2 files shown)


Ocramius:
Hmm, this PR is only composed by merge commits: something is wrong here.

You don't need to worry much about that, if you are OK with squashing it.

src/Util/AstHelper.php
```diff
@@ -0,0 +1,113 @@
...
+use PhpParser\PrettyPrinterAbstract;
+use RuntimeException;
+
+final class AstHelper {
+    private BetterReflection $betterReflection;
+
+    public function __construct( ?BetterReflection $betterReflection = null ) {
+        if ( !$betterReflection  ) {
+            $this->betterReflection = new BetterReflection();
+        } else {
```
Ocramius, line 31: I suggest we always ask for dependencies, especially in new code.

We usually make these optional only to help where the API wasn't well defined upfront, and we needed to preserve BC.

src/Util/AstHelper.php
```diff
@@ -0,0 +1,113 @@
...
+            $this->betterReflection = $betterReflection;
+        }
+    }
+    protected function getBetterReflection(): BetterReflection {
+        return $this->betterReflection;
+    }
```
Ocramius, line 37: Not really needed

src/Util/AstHelper.php
```diff
@@ -0,0 +1,113 @@
...
+    /**
+     * 
+     * @return Node[]|null $nodes
```
Ocramius, line 40: Can we be more precise about the type of the returned array? Perhaps `list<Node>`?

When could this be `null`? Can we somehow get rid of that scenario?

src/Util/AstHelper.php
```diff
@@ -0,0 +1,113 @@
...
+        $traverser->traverse($ast);
+        $nodes = $visitor->getFoundNodes();
+
+        return $nodes;
+    }
+    /**
+     * @param Node[] $node
+     * @psalm-suppress MixedArgument,NoInterfaceProperties
+     */
+    public function getBodyCode(array $node, PrettyPrinterAbstract|null $printer = null): String|null {
```
Ocramius, line 97: Is this supposed to be `public`?

src/Util/AstHelper.php
```diff
@@ -0,0 +1,113 @@
...
+use PhpParser\Node\Expr\ClosureUse as ClosureUseNode;
+use PhpParser\Node\Expr\ArrowFunction as ArrowFunctionNode;
+
+use PhpParser\NodeTraverser;
+use PhpParser\NodeVisitor\FindingVisitor;
+use PhpParser\PrettyPrinter\Standard as StandardPrettyPrinter;
+use PhpParser\PrettyPrinterAbstract;
+use RuntimeException;
+
+final class AstHelper {
```
Ocramius, line 25: We should probably rename this to get rid of the `Helper` naming smell (after all, all code is a "helper" :) )

Perhaps this could be named `ExtractAst`? Usage would then be `ExtractAst#fromClass()`, `ExtractAst#fromFunction()`, etc...

src/Util/AstHelper.php
```diff
@@ -0,0 +1,113 @@
...
+    }
+    protected function getBetterReflection(): BetterReflection {
+        return $this->betterReflection;
+    }
+    /**
+     * 
+     * @return Node[]|null $nodes
+     * 
+     */
+    public function getAstForReflection(ReflectionClass|ReflectionFunction $reflection): array|null {
```
Ocramius, line 43: Can we really put all reflection symbols in the same bucket?

src/Util/AstHelper.php
```diff
@@ -0,0 +1,113 @@
...
+        $parser = $this->getBetterReflection()->phpParser();
+        $source = $reflection->getLocatedSource();
+        return $parser->parse($source->getSource());
```
Ocramius, line 46: What happens if the `LocatedSource` contains multiple symbols?

For example:

```php
function foo() {}
function bar() {}
```

We probably only want the AST of the requested symbol, no?

Ocramius (changes requested):
Added suggestions for now: review of tests  will be done by our CI pipeline, mostly :)

test/unit/Util/AstHelperTest.php
```diff
@@ -0,0 +1,103 @@
...
+use Roave\BetterReflectionTest\BetterReflectionSingleton;
+use Roave\BetterReflection\SourceLocator\Ast\Locator;
+use Roave\BetterReflection\Util\AstHelper;
+use PhpParser\Node\Stmt\Class_;
+use PhpParser\Node\Stmt\Function_;
+use Roave\BetterReflection\Reflection\ReflectionFunction;
+
+
+
+final class AstHelperTest extends TestCase
```
Ocramius, line 20: Let's add test coverage declarations here: we don't want coverage from this test to leak on other classes

Wuffz:
As I was working on my project/proof of concept. I realised I was using this mostly to get the body of a closure amd getting the ast.  While working on it I noticed that using nicic/php-parser was the only dependency I needed. If that project will ever get into production, this entire package would be bloat because I use 1 single tiny feature which was intentionally removed. So I think I am going to skip this for now. 

If anyone wants to proceed or needs it anyway, let me know :)

Ocramius:
Fair enough, just beware that the comments put here also apply to the code you keep internally 😁

I'll close this for now, but good reference material, should anyone need it in a future stab at this problem 👍

PR #1345, opened 2023-05-15
Remove slow `assertReadableFile()` check

staabm:
this line showed up in my profilling and I realized most call-sites already do file-IO on the given filename before creating a `LocatedSource`, e.g.

- https://github.com/Roave/BetterReflection/blob/6b5cd7f7f68110d08f89644416d4911aff8f1b3e/src/SourceLocator/Type/AnonymousClassObjectSourceLocator.php#L123
- https://github.com/Roave/BetterReflection/blob/6b5cd7f7f68110d08f89644416d4911aff8f1b3e/src/SourceLocator/Type/AutoloadSourceLocator.php#L90-L103
- https://github.com/Roave/BetterReflection/blob/6b5cd7f7f68110d08f89644416d4911aff8f1b3e/src/SourceLocator/Type/SingleFileSourceLocator.php#L46-L50

grafik

### DIFF
--- a/src/SourceLocator/Located/LocatedSource.php
+++ b/src/SourceLocator/Located/LocatedSource.php
@@ -6,7 +6,6 @@ namespace Roave\BetterReflection\SourceLocator\Located;
 
 use InvalidArgumentException;
 use Roave\BetterReflection\SourceLocator\Exception\InvalidFileLocation;
-use Roave\BetterReflection\SourceLocator\FileChecker;
 use Roave\BetterReflection\Util\FileHelper;
 
 use function assert;
@@ -31,8 +30,6 @@ class LocatedSource
     {
         if ($filename !== null) {
             assert($filename !== '');
-
-            FileChecker::assertReadableFile($filename);
 
             $filename = FileHelper::normalizeWindowsPath($filename);
         }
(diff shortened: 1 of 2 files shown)


src/SourceLocator/Located/LocatedSource.php
```diff
@@ -32,8 +31,6 @@ public function __construct(private string $source, private string|null $name, s
         if ($filename !== null) {
             assert($filename !== '');
 
-            FileChecker::assertReadableFile($filename);
```
Ocramius, line 35: Worth using realpath here?

staabm: I did not find a single place where we pass in filename, but did not already do IO on the path, like e.g. `file_get_contents`.
I think this check here is useless.

we could hide it behand a `if(zend-assertions-enabled)` switch or similar, if you prefer.

I don't get why `realpath()` would be different then what we have right now?

Ocramius: This API is public, so yah, some invariant checking would be expected.

Ocramius: Ah, never mind, it's `@internal`

I'll 🚢 as-is, thanks @staabm!

Ocramius (approved):

src/SourceLocator/Located/LocatedSource.php
staabm: Thank you 🙏

PR #1353, opened 2023-06-21
CompileNodeToValueTest: failing tests for unsupported default values (enum case, new)

janedbal:
I was trying to fix this issue, but it looks like it is not designed for non-scalar values: `ReflectionParameter::getDefaultValue(): string|int|float|bool|array|null`.

----------

The actual issue we are facing is that we [initialize custom attribute](https://www.php.net/manual/en/reflectionattribute.newinstance.php) within custom phpstan rule. When a default value of such attribute is enum case, phpstan collapses.

### DIFF
--- a/test/unit/NodeCompiler/CompileNodeToValueTest.php
+++ b/test/unit/NodeCompiler/CompileNodeToValueTest.php
@@ -403,6 +403,43 @@ PHP;
         self::assertSame('baz', $paramInfo->getDefaultValue());
     }
 
+    public function testNewInInitializers(): void
+    {
+        $phpCode = '<?php
+        class Bar {
+            public function method($param = new stdClass()) {}
+        }
+        ';
+
+        $reflector = new DefaultReflector(new StringSourceLocator($phpCode, $this->astLocator));
+
+        $reflector->reflectClass('Bar')->getMethod('method')->getParameter('param')->getDefaultValue(); // should be stdClass
+    }
+
+    public function testEnumCaseResolution(): void
+    {
+        $phpCode = '<?php
+        enum BackedFoo: string {
+            case Bar = \'baz\';
+
+            public function method($param = self::Bar) {}
+        }
+        enum Foo {
+            case Bar;
+
+            public function method($param = self::Bar) {}
+        }
+        ';
+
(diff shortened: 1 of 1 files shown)


Ocramius:
This is correct: `getDefaultValue()` cannot produce `object`, as it goes against the basic invariant of `roave/better-reflection` not loading any of the code being analyzed.

kukulich:
@janedbal it can be “fixed” in https://github.com/ondrejmirtes/BetterReflection

Ocramius:
Certainly, but the basic invariant here is "don't load code", so that will certainly not happen here.

Also for the fork, loading runtime code could even be categorized as a security issue, from my point of view: instead of loading the code, load the AST for it instead.

JanTvrdik:
If it cannot be fixed, which I understand, could it at least throw better exception in such cases?

Ocramius:
@JanTvrdik what's the current exception you get? We generally have quite specific exception types.

Is it this one, perhaps? 

https://github.com/Roave/BetterReflection/blob/ce6a9a27c746fbd45d5a6603ba24e7d4a1f3b0f7/src/NodeCompiler/CompileNodeToValue.php#LL132C35-L132C35

Another alternative that comes to mind is having `BetterReflection` operating in "unsafe mode": something like `new Reflector($yesIReallyWantToExposeMyselfToRCEAttacksWhileInspectingForeignCodebases = true)`

janedbal:
Maybe worth updating [Compatibility page](https://github.com/Roave/BetterReflection/blob/6.11.x/docs/compatibility.md) with this issue? [ReflectionAttribute](https://www.php.net/manual/en/class.reflectionattribute.php) is missing there completely.

JanTvrdik:
@Ocramius It's actually https://github.com/Roave/BetterReflection/blob/6.11.x/src/NodeCompiler/CompileNodeToValue.php#L250, because the "enum case fetch" is represented as `ClassConstFetch` in AST. So instead of "constant not found" error I would expect "enum values are unsupported" error.

Ocramius:
@janedbal please do send a patch :-) 

@JanTvrdik is it possible to have a test with a better message expectation, perhaps? That would certainly improve the DX a lot :)

PR #1354, opened 2023-06-22
compatibility docs: add ReflectionAttribute

janedbal:
See https://github.com/Roave/BetterReflection/pull/1353#issuecomment-1602343689


Ocramius (approved):
Thanks @janedbal!

PR #1359, opened 2023-07-31
Implement `ReflectionClass->getParentClassName()`

clxmstaab:
A class can declare a parent which we cannot locate via autoloading mechanics.
For some static analysis cases its still useful to know, whether a parent class is defined - even if we can't reflect on it for some reason.

I was not able to find a API which allows me to dermine whether a class has a parent declared but only BetterReflection cannot find its definition from whether there is no parent class declared at all.

Before this PR we had to use reflection hacks to get the private property out of the class, see  https://github.com/rectorphp/rector/issues/8093#issuecomment-1657057947

### DIFF
--- a/src/Reflection/Adapter/ReflectionClass.php
+++ b/src/Reflection/Adapter/ReflectionClass.php
@@ -452,6 +452,14 @@ final class ReflectionClass extends CoreReflectionClass
         throw new Exception\NotImplemented('Not implemented');
     }
 
+    /**
+     * @return class-string|null
+     */
+    public function getParentClassName(): ?string
+    {
+        return $this->betterReflectionClass->getParentClassName();
+    }
+
     /** @psalm-mutation-free */
     public function getParentClass(): ReflectionClass|false
     {
--- a/src/Reflection/Adapter/ReflectionClass.php
+++ b/src/Reflection/Adapter/ReflectionClass.php
@@ -452,14 +452,6 @@ final class ReflectionClass extends CoreReflectionClass
         throw new Exception\NotImplemented('Not implemented');
     }
 
-    /**
-     * @return class-string|null
-     */
-    public function getParentClassName(): ?string
-    {
-        return $this->betterReflectionClass->getParentClassName();
-    }
-
     /** @psalm-mutation-free */
     public function getParentClass(): ReflectionClass|false
     {
--- a/src/Reflection/ReflectionClass.php
+++ b/src/Reflection/ReflectionClass.php
@@ -319,13 +319,11 @@ class ReflectionClass implements Reflection
         return $this->name;
     }
 
-	/**
-	 * @return class-string|null
-	 */
-	public function getParentClassName(): ?string
-	{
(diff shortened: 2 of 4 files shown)


src/Reflection/ReflectionClass.php
```diff
@@ -101,7 +101,7 @@ class ReflectionClass implements Reflection
     private int $endColumn;
 
     /** @var class-string|null */
-    private string|null $parentClassName;
+    private string|null $parentClassName = null;
```
Ocramius: This is always initialized in the constructor: shouldn't need a default value.

Ocramius (changes requested):

src/Reflection/Adapter/ReflectionClass.php
```diff
@@ -452,6 +452,14 @@ public function newInstanceArgs(array|null $args = null): object
         throw new Exception\NotImplemented('Not implemented');
     }
 
+    /**
```
kukulich: Is it good idea to add method to adapter?

Ocramius: Oh, true, this should only be on the BetterReflection API, **not** on the adapter

src/Reflection/ReflectionClass.php
clxmstaab: I added it, because the tests failed with "must not be accessed before initialization"

https://github.com/Roave/BetterReflection/actions/runs/5713496659/job/15479000753

src/Reflection/Adapter/ReflectionClass.php
clxmstaab: ok, I wasn't able to differentiate which is what and thought I need to reflect the change there. even easier if this is not required :)

clxmstaab: I want to call this new method from [PHPStan ClassReflection](https://github.com/phpstan/phpstan-src/blob/1.11.x/src/Reflection/ClassReflection.php). I guess this means we need it on the adapter then?

Ocramius: From an architectural PoV, it does not fit here: the adapter is only for mirroring the API of ext-reflection.

Anything beyond that already couples the consumer to BetterReflection, rather than ext-reflection, and should not be here.

kukulich: @clxmstaab PHPStan uses BR fork https://github.com/ondrejmirtes/BetterReflection/, so it can be implemented there.

staabm (commented):
I think this should be good to go

(sorry for mixing commits from my personal and my corporate account..)

src/Reflection/Adapter/ReflectionClass.php
Ocramius: even then, remember that the dependency is to ext-reflection, and the adapters are not supposed to deviate from that API.

Ocramius (approved):
Thanks @staabm!

PR #1360, opened 2023-07-31
Fix typos

szepeviktor:
Found some misspelling with `typos`.
I've left "invokable" alone.


test/core/010.phpt
```diff
@@ -1,5 +1,5 @@
 --TEST--
-ReflectionMethod::__toString() tests (overriden method)
```
Ocramius: These tests are imported from PHP-SRC, and should not be touched

Ocramius (changes requested):

test/core/010.phpt
szepeviktor: All right!
Reverting ...

szepeviktor: Done 🍏

Ocramius (approved):
Thanks @szepeviktor!

Note: due to the various revert commits, I'm squash-merging this.

szepeviktor:
Glad to contribute.

PR #1361, opened 2023-08-07
Implement `ReflectionClass#getInterfaceClassNames()`

staabm:
same motivation as in https://github.com/Roave/BetterReflection/pull/1359, but this time for interfaces :-)

refs https://github.com/rectorphp/rector/issues/8093#issuecomment-1657057947

---

if we agree on this one, I should do another PR for `traits` as a followup

### DIFF
--- a/test/unit/Reflection/ReflectionClassTest.php
+++ b/test/unit/Reflection/ReflectionClassTest.php
@@ -65,6 +65,7 @@ use Roave\BetterReflectionTest\Fixture\ClassUsesTwoTraitsWithSameMethodNameOneIs
 use Roave\BetterReflectionTest\Fixture\ClassUsingTraitWithAbstractMethod;
 use Roave\BetterReflectionTest\Fixture\ClassWithAttributes;
 use Roave\BetterReflectionTest\Fixture\ClassWithCaseInsensitiveMethods;
+use Roave\BetterReflectionTest\Fixture\ClassWithMissingInterface;
 use Roave\BetterReflectionTest\Fixture\ClassWithMissingParent;
 use Roave\BetterReflectionTest\Fixture\ClassWithNonAbstractTraitMethodThatOverwritePreviousAbstractTraitMethod;
 use Roave\BetterReflectionTest\Fixture\DefaultProperties;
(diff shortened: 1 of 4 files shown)


test/unit/Reflection/ReflectionClassTest.php
```diff
@@ -425,6 +426,93 @@ public function testGetParentClassWithMissingParent(): void
...
+    {
+        $classInfo = (new DefaultReflector(new SingleFileSourceLocator(
+            __DIR__ . '/../Fixture/ClassWithMissingInterface.php',
+            $this->astLocator,
+        )))->reflectClass(ClassWithMissingInterface::class);
+
+        $this->expectException(IdentifierNotFound::class);
+
+        self::assertNotNull($classInfo->getInterfaces());
+    }
```
staabm, line 451: these tests show, that the pre-PR existing API always throws and therefore does not allow to get the interface names, when the class-symbols are not autoloadable.

test/unit/Reflection/ReflectionClassTest.php
```diff
@@ -425,6 +426,93 @@ public function testGetParentClassWithMissingParent(): void
...
+
+    /** @return list<array{string, class-string, list<class-string>}> */
+    public static function getInterfaceClassNamesDataProvider(): array
+    {
+        return [
+            [
+                __DIR__ . '/../Fixture/ClassWithMissingInterface.php',
+                ClassWithMissingInterface::class,
+                ['Roave\BetterReflectionTest\Fixture\InterfaceThatDoesNotExist'],
+            ],
```
staabm, line 476: this first test-case shows that we are now able to extract the interface names even if not autoloadable.

the remaining cases add coverage for a few more related cases like autoloadable classes, traits,..

Ocramius:
This is a new feature: should go to `6.12.x` :-)

Ocramius (approved):
LGTM: feel free to send more of these, BTW.

staabm:
oh sorry. did not relealize 6.12.x is already open

Ocramius:
Yeah, release automation does create a new branch each time :)

Ocramius:
Thanks @staabm!

staabm:
thank you. I am already working on the "trait-variant" of it

PR #1364, opened 2023-08-28
micro-optimization: use `spl_object_id()` instead of `spl_object_hash()`

staabm:
`spl_object_id()` exists since php 7.2, I think we can use it over `spl_object_id()`.

rector and wordpress [micro benchmarks suggests](https://github.com/rectorphp/rector-src/pull/4876) its faster

### DIFF
--- a/src/SourceLocator/Type/MemoizingSourceLocator.php
+++ b/src/SourceLocator/Type/MemoizingSourceLocator.php
@@ -52,7 +52,7 @@ final class MemoizingSourceLocator implements SourceLocator
 
     private function reflectorCacheKey(Reflector $reflector): string
     {
-        return sprintf('type:%s#oid:%s', $reflector::class, spl_object_id($reflector));
+        return sprintf('type:%s#oid:%d', $reflector::class, spl_object_id($reflector));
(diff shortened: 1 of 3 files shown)


samsonasik (approved):

src/SourceLocator/Type/MemoizingSourceLocator.php
```diff
@@ -52,7 +52,7 @@ public function locateIdentifiersByType(Reflector $reflector, IdentifierType $id
 
     private function reflectorCacheKey(Reflector $reflector): string
     {
-        return sprintf('type:%s#oid:%s', $reflector::class, spl_object_hash($reflector));
+        return sprintf('type:%s#oid:%s', $reflector::class, spl_object_id($reflector));
```
samsonasik: sprintf shoud use %d since spl_object_id returns int

staabm: thanks - fixed

samsonasik (approved):

Ocramius (approved):
Thanks @staabm!

PR #1368, opened 2023-09-08
Support for new PHP 8.3 reflection API

kukulich:


src/Reflection/Adapter/ReflectionMethod.php
```diff
@@ -34,6 +37,27 @@ public function __construct(private BetterReflectionMethod $betterReflectionMeth
         unset($this->class);
     }
 
+    public static function createFromMethodName(string $method): static
+    {
+        if (! str_contains($method, '::')) {
+            throw new CoreReflectionException('Argument #1 ($method) must be a valid method name');
+        }
+
+        [$className, $methodName] = explode('::', $method);
```
Ocramius: What happens if a value like `foo::bar::baz` is given? I think we should have `2` as third argument here

test/unit/SourceLocator/SourceStubber/PhpStormStubsSourceStubberTest.php
```diff
@@ -265,6 +271,17 @@ public static function internalFunctionsProvider(): array
             array_filter(
                 $functionNames,
                 static function (string $functionName): bool {
+                    if (
+                        in_array($functionName, [
+                        // Missing in JetBrains/phpstorm-stubs
```
Ocramius, line 276: Are the stubs to be updated in `composer.lock`?

Ocramius (changes requested):

test/unit/SourceLocator/SourceStubber/PhpStormStubsSourceStubberTest.php
kukulich: We use latest version currently. So we have to wait for new release.

src/Reflection/Adapter/ReflectionMethod.php
kukulich: Added 👍

Ocramius (approved):

Ocramius:
Thanks @kukulich!

Issue #1376, opened 2023-11-14
Default enum value for a parameter throw an exception


forrest79:
Hi, I figured out, that reflecting a class like this will fail on getting the default value for the `$x` parameter:

```php
enum TestEnum: int {
    case X = 1;
}

class TestClass {
    public function text(TestEnum $x = TestEnum::X) {
    }
}
```

I have got `Uncaught Roave\BetterReflection\NodeCompiler\Exception\UnableToCompileNode: Could not locate constant TestEnum::X while trying to evaluate constant expression in method TestClass::test() in file ...`.

As a quick workaround, I added something like this:

```php
if ($classReflection instanceof ReflectionEnum) {
	$case = $classReflection->getCase($constantName);

	if ($case === null) {
		throw Exception\UnableToCompileNode::becauseOfNotFoundClassConstantReference($context, $classReflection, $node);
	}

	return $case->getValue();
}
```

Into `src/NodeCompiler/CompileNodeToValue.php` method `getClassConstantValue()`. I wanted to return the correct enum instance, but I got another error, the enum is not supported as return value.

Ocramius:
This is correct, because BetterReflection would have to instantiate the object, therefore loading it.

Ocramius:
BTW, the exception message is obviously confusing: 

 > `Could not locate constant TestEnum::X while trying to evaluate constant expression in method TestClass::test() in file`


We could improve that :thinking:

forrest79:
Thanks for a rapid response. I solved this using `getDefaultValueExpression` (my bad :-()

Ocramius:
@forrest79 are we OK with just raising a better exception message here?

forrest79:
Absolutely :-) It was my fault, I was expecting, that Enum could be simply returned as a returned value.

Ocramius:
@forrest79 are we OK with just raising a better exception message here? Would you be able to help with that, perhaps?

forrest79:
Yes, I will be glad to help. My idea is just to detect, if the class is an Enum and throw a different exception?

Ocramius:
I'd say so, yep 👍

forrest79:
PR #1380

Ocramius:
Handled in #1380

PR #1380, opened 2023-11-22
Better exception message for when default value is an enum

forrest79:
This solves an issue #1376.

### DIFF
--- a/test/unit/NodeCompiler/CompileNodeToValueTest.php
+++ b/test/unit/NodeCompiler/CompileNodeToValueTest.php
@@ -741,6 +741,29 @@ PHP;
         $classInfo->getConstant('ONE_VALUE')->getValue();
     }
 
+    public function testEnumValueThrowsException(): void
+    {
+        $phpCode = <<<'PHP'
+        <?php
+
+        enum Foo: int {
+            case ONE = 1;
+        }
+        class Bat {
+            const ONE_VALUE = Foo::ONE;
+        }
+        PHP;
+
+        $reflector = new DefaultReflector(new AggregateSourceLocator([
+            new StringSourceLocator($phpCode, $this->astLocator),
+            new PhpInternalSourceLocator($this->astLocator, $this->sourceStubber),
+        ]));
+        $classInfo = $reflector->reflectClass('Bat');
+
+        $this->expectException(UnableToCompileNode::class);
+        $classInfo->getConstant('ONE_VALUE')->getValue();
+    }
+
     /** @return list<array{0: string, 1: mixed}> */
     public static function magicConstantsWithoutNamespaceProvider(): array
     {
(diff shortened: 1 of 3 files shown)


test/unit/NodeCompiler/CompileNodeToValueTest.php
```diff
@@ -741,6 +741,29 @@ class Bat {
...
+        }
+        PHP;
+
+        $reflector = new DefaultReflector(new AggregateSourceLocator([
+            new StringSourceLocator($phpCode, $this->astLocator),
+            new PhpInternalSourceLocator($this->astLocator, $this->sourceStubber),
+        ]));
+        $classInfo = $reflector->reflectClass('Bat');
+
+        $this->expectException(UnableToCompileNode::class);
```
Ocramius: We should check at least part of the exception message in this test: it seems like the mutation test reports no behavioral change (removes `throw` and is happy with that), which indicates that an assertion is to be added here

Ocramius (changes requested):

forrest79:
Sorry for a delay, I was working remotely for a couple of days. I added a message check. You can look at it.

test/unit/NodeCompiler/CompileNodeToValueTest.php
```diff
@@ -760,8 +761,13 @@ class Bat {
         ]));
         $classInfo = $reflector->reflectClass('Bat');
 
-        $this->expectException(UnableToCompileNode::class);
-        $classInfo->getConstant('ONE_VALUE')->getValue();
+        try {
+            $classInfo->getConstant('ONE_VALUE')->getValue();
+        } catch (Throwable $e) {
+            self::assertInstanceOf(UnableToCompileNode::class, $e);
+            self::assertStringContainsString('An enum expression Foo::ONE is not supported in class Bat in file ', $e->getMessage());
```
Ocramius: Can use `expectExceptionMessage`, rather than adding a try-catch 😁

Ocramius (changes requested):
CS issues fixable with some minimal assertion improvements (by removing try-catch)

test/unit/NodeCompiler/CompileNodeToValueTest.php
forrest79: Done

Ocramius:
I think the failures are unrelated to your patch, and instead depend on https://github.com/shivammathur/setup-php/issues/799

Ocramius (approved):
LGTM, thanks @forrest79!

I will re-run tests and merge once the upstream issue with the CI pipeline is resolved, then releasing :muscle:

Ocramius:
Needs https://github.com/shivammathur/php-ubuntu/pull/3

Ocramius (approved):
Excellent, thanks @forrest79!

PR #1383, opened 2023-12-02
Support for `final` modifier in traits

kukulich:


Ocramius:
So... It's already supported? 🤔

kukulich:
@Ocramius Thank you :) I've checked it again and find a bug :)

kukulich:
Ok, it looks the test was right because the behaviour is a little strange and it removes visibility modifiers https://3v4l.org/fS33Y#v8.3.0

kukulich:
https://github.com/php/php-src/issues/12854

Ocramius:
This is now in PHP 8.3.2+

https://github.com/php/php-src/commit/e679ab36b362c34b899672c36982b194a5e98ce3

Worth moving forward?

kukulich:
Yes, it's in my todo list :)

Ocramius (approved):
Thanks @kukulich! One day, traits will make sense in PHP: that's not today

PR #1388, opened 2024-01-15
Prevent creation of ReflectionMethod objects

staabm:
while profilling a PHPStan `createImmediateMethods` was showing up.

lets make it more lazy and don't create unnecessary objects

grafik


staabm:
hmm the code I see locally and the one in the repo is different.. seems like I am not using the latest vendor

PR #1389, opened 2024-01-15
Fixed `ReflectionSourceStubber` for `enum`s, so it no longer generates `implements` and `cases()`, `tryFrom()`, `from()` methods as part of the source

kukulich:


Ocramius (approved):
Thanks @kukulich!

PR #1391, opened 2024-01-28
CI tools moved to special composer.json

kukulich:


tools/composer.json
```diff
@@ -0,0 +1,30 @@
+{
```
Ocramius, line 1: Does renovate need adjustments here too? 🤔

Ocramius (approved):
> `PHP Fatal error:  Uncaught OutOfBoundsException: Package "vimeo/psalm" is not installed in /home/runner/work/BetterReflection/BetterReflection/vendor/composer/InstalledVersions.php:182`

Welp, I may be able to patch around this in upstream 😬

tools/composer.json
kukulich: I was hoping it will work automatically 🤞

Ocramius:
Seems like this bit needs some polyfilling or such :thinking: 

https://github.com/Roave/infection-static-analysis-plugin/blob/975adabcbbdbd6c550159d3203a7336fd82cedb7/bin/roave-infection-static-analysis-plugin#L54-L60

Ocramius:
Release of other tool going up right now 👍

composer.json
```diff
@@ -32,40 +32,17 @@
         }
 
     ],
-    "require-dev": {
-        "doctrine/coding-standard": "^12.0.0",
-        "phpstan/phpstan": "^1.10.60",
-        "phpstan/phpstan-phpunit": "^1.3.16",
-        "phpunit/phpunit": "^10.5.12",
-        "vimeo/psalm": "5.23.0",
-        "roave/infection-static-analysis-plugin": "^1.34.0"
```
Ocramius, line 41: This tool shares the same address space as the main library: moving it here may be a mistake (same for phpunit).

I think we should only move tools that operate completely independently, which currently includes phpcs, phpstan (not psalm, due to the coupling with the test suite mutations).

That said, I'm fine with trying this out anyway 👍

composer.json
```diff
@@ -32,40 +32,17 @@
...
-    },
     "autoload": {
         "psr-4": {
             "Roave\\BetterReflection\\": "src"
         }
     },
-    "autoload-dev": {
```
Ocramius, line 48: This will also mess with editors 🤔

I'm not convinced of moving everything, but I'm open to trying it, for now 🤔

Ocramius (commented):
I think we should tune this back a notch: we can move tools that don't share address space, but not tools that do share it 🤔

Let's see how this goes first, but caution is advised.

Happy to include sources as a "dependcency" from the tools directory though: that approach is smart too 👍

tools/composer.json
```diff
@@ -0,0 +1,30 @@
...
+    "config": {
+        "allow-plugins": {
+            "infection/extension-installer": true,
+            "dealerdirect/phpcodesniffer-composer-installer": true
+        },
+        "platform": {
+            "php": "8.1.99"
+        }
+    },
+    "autoload": {
```
Ocramius, line 23: Thinking of this further, we should probably add the parent package as a dependency, not as an autoload definition: that makes the parent package able to declare its own dependencies, and the tooling follows only

kukulich:
I’m sorry - there was “approved” so I merged it.

Ocramius:
That's fine: it doesn't currently affect downstream, and we can always go back: I just thought this was splitting out a couple tools, but I misread from my phone, so I was too quick 😬

Let's see how this works out meanwhile 👍

kukulich:
Just another idea - would it be possible to prepare phar for https://github.com/Roave/infection-static-analysis-plugin ? Dependencies will be packed in the Phar

Ocramius:
IMO no:

 1. Maintaining phar is a massive pain in the royal butt
 2. I prefer people to rely on an explicit `composer.lock` that they maintain, like we do here

Also, the tool sharing memory space with the program is a problem, when used from a phar: that's why I still firmly believe that using tools like phpunit from a phar is a mistake.

Ocramius:
BTW, it would be fine if address space wasn't shared (most static analysis tools), but not for this

kukulich:
Hmm, it looks the PR was not successful.

We are still not able to upgrade PhpParser: https://github.com/Roave/BetterReflection/pull/1387

PhpParser is used in tests so if we upgrade it we will have problem with Psalm and Infection dependencies again :(

PR #1396, opened 2024-03-08
Implement PHP Version dependent deprecations for constants

staabm:
with this PR we strip the `@deprecated` phpdoc tag on constant definitions, in case the current php version does not match the declared min version.

motivation: report deprecation errors for constants only when the phpversion matches the `@deprecated $version`  we find in the jetbrains/phpstorm-stubs in PHPStan

example: https://github.com/JetBrains/phpstorm-stubs/blob/9608c953230b08f07b703ecfe459cc58d5421437/filter/filter.php#L478

### DIFF
--- a/src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
+++ b/src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
@@ -617,6 +617,10 @@ final class PhpStormStubsSourceStubber implements SourceStubber
     private function addDeprecatedDocComment(Node\Stmt\ClassLike|Node\Stmt\ClassConst|Node\Stmt\Property|Node\Stmt\ClassMethod|Node\Stmt\Function_|Node\Stmt\Const_ $node): void
     {
         if ($node instanceof Node\Stmt\Const_) {
+            if (! $this->isDeprecatedByPhpDocInPhpVersion($node)) {
+                $this->removeAnnotationFromDocComment($node, 'deprecated');
+            }
+
             return;
         }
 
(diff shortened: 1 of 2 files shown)


src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
```diff
@@ -662,6 +666,27 @@ private function isCoreExtension(string $extension): bool
...
     }
 
+    private function isDeprecatedByPhpDocInPhpVersion(Node\Stmt\Const_ $node): bool
+    {
+        $docComment = $node->getDocComment();
+        if ($docComment === null) {
+            return false;
+        }
+
+        if (preg_match('#@deprecated (\d+)\.(\d+)(?:\.(\d+)?)$#m', $docComment->getText(), $matches) === 1) {
```
Ocramius: Possibly `\s+`, rather than a single space?

test/unit/SourceLocator/SourceStubber/PhpStormStubsSourceStubberTest.php
```diff
@@ -465,6 +465,33 @@ public function testStubForConstantThatExists(): void
...
     }
 
+    public function testStubForConstantThatIsDeprecated(): void
+    {
+        $stubData = $this->sourceStubber->generateConstantStub('FILTER_SANITIZE_STRING');
+
+        self::assertInstanceOf(StubData::class, $stubData);
+
+        self::assertStringMatchesFormat(
+            "%Adefine('FILTER_SANITIZE_STRING',%w%d);",
```
Ocramius, line 484: I'm OK with the patch per se, but this test will start failing when we switch PHP versions, which makes it a bit fragile.

I wonder if we could define a stub that we read just for the purposes of this test? :thinking: 

/cc @kukulich WDYT?

staabm: good idea. added a fake stub instead.

Ocramius (approved):
LGTM! Thanks @staabm!

PR #1397, opened 2024-03-08
CI: use pcov for type coverage

staabm:
just a test whether this speedsup the CI pipeline

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -137,7 +137,7 @@ jobs:
       - name: "Install PHP"
         uses: "shivammathur/setup-php@2.29.0"
         with:
-          coverage: "xdebug"
+          coverage: "pcov"
           php-version: "${{ matrix.php-version }}"
           ini-values: memory_limit=-1, zend.assertions=1
 


Ocramius:
The biggest offender when it comes to performance here is `roave/infection-static-analysis-plugin` operating (under the hood) with Psalm's analysis, which needs more tuning around file consistency checks, perhaps. 

It's still faster than running all the mutants, but it's also still very slow when considering analysis on a limited set of changed files.

staabm:
I need to dive into `roave/infection-static-analysis-plugin` to understand more what it is actually doing :-).

my naiv thinking of this PR was/is: xdebug makes everything super slow. maybe pcov does have less function call overhead etc and might speedup it. lets see.

Ocramius:
It's really tiny. For each escaped mutant process (generated mutation), it will run static analysis:

https://github.com/Roave/infection-static-analysis-plugin/blob/063ab39377a515ec115d6480367d41d0f1a894a7/src/Roave/InfectionStaticAnalysis/RunStaticAnalysisAgainstEscapedMutant.php#L36-L71

In there, the major-major performance bottleneck is:

```php
if ($this->runStaticAnalysis->isMutantStillValidAccordingToStaticAnalysis($mutantProcess->getMutant())) {
```

staabm:
do you think we could use `--filter` and run several jobs in parallel ?

https://infection.github.io/guide/command-line-options.html

Ocramius:
Possibly, but CI is not as frequent here, so 15m are not worth wasting time on, unless you are looking for very specific mutants.

PR #1398, opened 2024-03-09
Harmonize version parsing

staabm:
tiny cleanup of my previous PR, as I realized there is already a parsing util method in this class


Ocramius:
@staabm closed on purpose?

staabm:
Yes. Realized that we would need additional type conversions. Not worth it

PR #1402, opened 2024-03-11
Move `PHPUnit` back to standard `composer.json`

kukulich:


test/benchmark/PhpUnitTestCaseBench.php
```diff
@@ -5,7 +5,7 @@
 namespace Roave\BetterReflectionBenchmark;
 
 use PhpBench\Benchmark\Metadata\Annotations\Iterations;
-use PHPUnit\Framework\TestCase;
+use Psalm\Config;
```
Ocramius: Seems wrong: psalm is not installed, so this will likely bite us?

Ocramius: Alternatively, let's look at a different file, such as a stub?

kukulich: I can use some file from PhpBench itself but I didn't find any big file like PHPUnit `TestCase`. That's why I've used `Psalm\Config` because Psalm is installed in `tools/composer.json` too.

Ocramius: Yeah, perhaps something in phpbench itself would be good: any cross-version comparisons are foobar anyway, at this point :D

kukulich: Change to biggest file in PhpBench itself.

Ocramius: Thanks!

Ocramius (approved):
:ship: (waiting for CI, though)

PR #1404, opened 2024-03-11
PhpParser 5 & PhpUnit 11

kukulich:
1. There are some little BC breaks, see changes eg. in`ClassForSourceStubberExpected.php`

2. 

```php
/** Comment */
define('CONSTANT_NAME', true);
```

`Node\Expr\FuncCall` is now wrapped in `Node\Stmt\Expression` and the comment is part of the wrapper. I've made a little hack and I copy the comment to the `FuncCall` node. See eg. `FindReflectionsInTree`


phpunit.xml.dist
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <phpunit
     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-    xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/10.1/phpunit.xsd"
+    xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/11.0/phpunit.xsd"
```
Ocramius: We should reference the XSD in `vendor/`

src/BetterReflection.php
```diff
@@ -58,9 +57,7 @@ public function reflector(): Reflector
     public function phpParser(): Parser
     {
         return $this->phpParser
-            ?? $this->phpParser = (new ParserFactory())->create(ParserFactory::ONLY_PHP7, new Emulative([
-                'usedAttributes' => ['comments', 'startLine', 'endLine', 'startFilePos', 'endFilePos'],
-            ]));
+            ?? $this->phpParser = (new ParserFactory())->createForNewestSupportedVersion();
```
Ocramius, line 60: Nice simplification!

Ocramius:
Note: didn't read through the whole patch yet 😁

phpunit.xml.dist
kukulich: Done.

src/SourceLocator/Ast/Parser/MemoizingParser.php
```diff
@@ -18,9 +19,12 @@
 /** @internal */
 final class MemoizingParser implements Parser
 {
-    /** @var array<string, string> indexed by source hash */
+    /** @var array<string, array{string, Token[]}> indexed by source hash */
     private array $sourceHashToAst = [];
 
+    /** @var Token[] */
+    private array $lastTokens = [];
```
Ocramius, line 26: This hack is quite risky: I don't think we should expose this API, as it introduces temporal coupling.

In code like this:

```php
$parser->parse($thing);

someCode(); // mutating logic: "fun"

$tokens = $parser->getTokens();
```

This will fail spectacularly.

I'll check where this is used, but I think we should not do this.

src/SourceLocator/Ast/Parser/MemoizingParser.php
```diff
@@ -34,15 +38,25 @@ public function parse(string $code, ErrorHandler|null $errorHandler = null): arr
...
+        $this->lastTokens             = $tokens;
 
         return $ast;
     }
+
+    /** @return Token[] */
+    public function getTokens(): array
+    {
+        return $this->lastTokens;
+    }
```
Ocramius, line 61: Connected to above comment in same file: we should not expose a method with high temporal coupling

src/SourceLocator/Ast/Strategy/NodeToReflection.php
```diff
@@ -28,8 +26,7 @@ public function __invoke(
         Node\Stmt\Namespace_|null $namespace,
         int|null $positionInNode = null,
     ): ReflectionClass|ReflectionConstant|ReflectionFunction {
-        /** @psalm-suppress PossiblyNullPropertyFetch, PossiblyNullReference */
-        $namespaceName = $namespace?->name !== null ? implode('\\', $namespace->name->getParts()) : null;
+        $namespaceName = $namespace?->name?->name;
```
Ocramius, line 29: very?->fancy?->simplification

src/SourceLocator/Type/AutoloadSourceLocator.php
```diff
@@ -298,29 +297,36 @@ public function enterNode(Node $node): int|null
...
                     assert($argumentNameNode instanceof Node\Arg);
                     $nameNode = $argumentNameNode->value;
                     assert($nameNode instanceof Node\Scalar\String_);
 
                     if ($nameNode->value === $this->constantName) {
-                        $this->node = $node;
+                        $nodeDocComment = $node->getDocComment();
+                        if ($nodeDocComment !== null) {
+                            $functionCall->setDocComment($nodeDocComment);
+                        }
```
Ocramius: Suspicious of this change: why is the `AutoloadSourceLocator` making decisions about **changing** the AST?

I'm not questioning the fact that this fixes something, but rather the fact that this call is specifically part of a source locator :thinking:

src/SourceLocator/Ast/Parser/MemoizingParser.php
```diff
@@ -34,15 +38,25 @@ public function parse(string $code, ErrorHandler|null $errorHandler = null): arr
...
+        $this->sourceHashToAst[$hash] = [serialize($ast), $tokens];
+        $this->lastTokens             = $tokens;
 
         return $ast;
     }
+
+    /** @return Token[] */
+    public function getTokens(): array
+    {
+        return $this->lastTokens;
```
Ocramius, line 60: I think we should call `$this->wrappedParser->getTokens()` directly, and not memoize ourselves.

If this API is added to be compliant with the upstream interface, let's just re-use the memoization from upstream, and not redo it ourselves :)

Ocramius (commented):
@kukulich I don't see any explicit BC breaks, or am I missing something?

src/SourceLocator/Ast/Parser/MemoizingParser.php
kukulich: The method is part of the interface: https://github.com/nikic/PHP-Parser/blob/master/lib/PhpParser/Parser.php

So should we throw “not supported” error?

Ocramius: Will ponder on it today: perhaps nothing we should do at all

Ocramius:
Did a local test of performance improvements:

```
betterreflection on  6.27.x [$?] via 🐘 v8.2.16 via ❄️  impure (nix-shell) 
❯ ./tools/vendor/bin/phpbench run --warmup=3 --revs=10 --iterations=20 --retry-threshold=4 --tag=6.27.x
betterreflection on  HEAD (8f94dee) [$?] via 🐘 v8.2.16 via ❄️  impure (nix-shell) 
❯ ./vendor/bin/phpbench report --ref=6.27.x --report=aggregate
+----------------------+------------------------------+-----+------+-----+----------+---------+--------+
| benchmark            | subject                      | set | revs | its | mem_peak | mode    | rstdev |
+----------------------+------------------------------+-----+------+-----+----------+---------+--------+
| PhpUnitTestCaseBench | benchReflectClass            |     | 10   | 20  | 17.154mb | 2.300μs | ±1.90% |
| PhpUnitTestCaseBench | benchReflectMethodParameters |     | 10   | 20  | 17.154mb | 0.900μs | ±0.00% |
+----------------------+------------------------------+-----+------+-----+----------+---------+--------+
```

```
betterreflection on  HEAD (8f94dee) [$?] via 🐘 v8.2.16 via ❄️  impure (nix-shell) 
❯ ./vendor/bin/phpbench run --warmup=3 --revs=10 --iterations=20 --retry-threshold=4 --tag=1404
betterreflection on  HEAD (8f94dee) [$?] via 🐘 v8.2.16 via ❄️  impure (nix-shell) took 8s 
❯ ./vendor/bin/phpbench report --ref=1404 --report=aggregate
+----------------------+------------------------------+-----+------+-----+----------+---------+--------+
| benchmark            | subject                      | set | revs | its | mem_peak | mode    | rstdev |
+----------------------+------------------------------+-----+------+-----+----------+---------+--------+
| PhpUnitTestCaseBench | benchReflectClass            |     | 10   | 20  | 11.882mb | 2.300μs | ±2.21% |
| PhpUnitTestCaseBench | benchReflectMethodParameters |     | 10   | 20  | 11.882mb | 0.900μs | ±0.00% |
+----------------------+------------------------------+-----+------+-----+----------+---------+--------+
```

Time-wise, we're running the same stuff, but memory usage dropped sharply, probably due to PHP-Parser relying less on `array` structures internally :muscle:

src/SourceLocator/Ast/Parser/MemoizingParser.php
Ocramius: After some pondering, I think we can leave this as-is: forwarding to the wrapped parser wouldn't work anyway, when returning cached results.

src/SourceLocator/Ast/Parser/MemoizingParser.php
Ocramius: After some pondering, I think we can leave this as-is: forwarding to the wrapped parser wouldn't work anyway, when returning cached results.

src/SourceLocator/Ast/Parser/MemoizingParser.php
Ocramius: After some pondering, I think we can leave this as-is: forwarding to the wrapped parser wouldn't work anyway, when returning cached results.

src/SourceLocator/Type/AutoloadSourceLocator.php
Ocramius: @kukulich this is the last concern in this patch: what's going on with us manipulating nodes during `*Locator` activities?

kukulich: It looks I've copied too much yesterday. The "copy" hack is not neccessary in `AutoloadSourceLocator`. Removed now.

Ocramius (approved):

Ocramius:
This is going in a minor release :)  

People should pin their own `nikic/php-parser`, if they rely on it directly ;-)

Issue #1382, opened 2023-11-24
Bring Mutation Test score back to 100%


Ocramius:
Seen in #1381 and #1380 ( https://github.com/Roave/BetterReflection/actions/runs/6983487200/job/19004937424?pr=1381#step:5:11665 )

```
Error: ] The minimum required MSI percentage should be 100%, but actual is      
         99.73%. Improve your tests!        
```

```
Escaped mutants:
================

1) /home/runner/work/BetterReflection/BetterReflection/src/Reflection/ReflectionFunctionAbstract.php:112    [M] GreaterThan

--- Original
+++ New
@@ @@
         $startLine = null;
         if ($node->hasAttribute('startLine')) {
             $startLine = $node->getStartLine();
-            assert($startLine > 0);
+            assert($startLine >= 0);
         }
         $endLine = null;
         if ($node->hasAttribute('endLine')) {


2) /home/runner/work/BetterReflection/BetterReflection/src/Reflection/ReflectionFunctionAbstract.php:118    [M] GreaterThan

--- Original
+++ New
@@ @@
         $endLine = null;
         if ($node->hasAttribute('endLine')) {
             $endLine = $node->getEndLine();
-            assert($endLine > 0);
+            assert($endLine >= 0);
         }
         $this->startLine = $startLine;
         $this->endLine = $endLine;


3) /home/runner/work/BetterReflection/BetterReflection/src/Reflection/ReflectionFunctionAbstract.php:507    [M] InstanceOf_

--- Original
+++ New
@@ @@
         if ($returnType === null) {
             return null;
         }
-        assert($returnType instanceof Node\Identifier || $returnType instanceof Node\Name || $returnType instanceof Node\NullableType || $returnType instanceof Node\UnionType || $returnType instanceof Node\IntersectionType);
+        assert(true || $returnType instanceof Node\Name || $returnType instanceof Node\NullableType || $returnType instanceof Node\UnionType || $returnType instanceof Node\IntersectionType);
         return ReflectionType::createFromNode($this->reflector, $this, $returnType);
     }
     /** @return list<ReflectionAttribute> */


4) /home/runner/work/BetterReflection/BetterReflection/src/Reflection/ReflectionFunctionAbstract.php:507    [M] InstanceOf_

--- Original
+++ New
@@ @@
         if ($returnType === null) {
             return null;
         }
[...]

PR #1407, opened 2024-03-13
Kill startLine & endLine mutants

staabm:
lets see how https://github.com/nikic/PHP-Parser/pull/985 affects mutations

refs https://github.com/Roave/BetterReflection/issues/1382

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<files psalm-version="5.23.0@005e3184fb6de4350a873b9b8c4dc3cede9db762">
+<files psalm-version="5.23.1@8471a896ccea3526b26d082f4461eeea467f10a4">
   <file src="src/Reflection/Attribute/ReflectionAttributeHelper.php">
     <ImpureMethodCall>
       <code><![CDATA[toLowerString]]></code>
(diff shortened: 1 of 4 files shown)


staabm:
@kukulich it looks like it doesn't improve mutations. I guess phpdoc-types are not useful for that (and only native types are taken into account?)

Ocramius:
Phpdoc types **are** considered: need to likely checked in more depth 🤔

Thanks for trying though!

kukulich:
@staabm 

I think this can work:

```
$startLine = $node->getStartLine()
if ($startLine !== -1) {
}
```

No `assert` -> no mutation and no mutant :) Psalm/Phpstan should be happy too thanks to your change in PhpParser.

The current solution probably does not work because `if ($node->hasAttribute('startLine')) {` does not tell Psalm that `$node->getStartLine()` is only positive-int. That's why the `assert` is not reported by Psalm as useless.

staabm:
Thank you. I will give your idea a try tomorrow 

(Btw: the php-parser PR was merged)

staabm:
the psalm error is a bug, reported in https://github.com/vimeo/psalm/issues/10826

kukulich:
It looks it's very complicated to kill one mutant 😄

staabm:
> It looks it's very complicated to kill one mutant 😄

hehe - agree, its a lot of change. but I feel the code is easier to read after the PR.

kukulich:
> I feel the code is easier to read after the PR.

I agree :)

staabm:
@Ocramius mutants will be killed on the next php-parser release. I think its good to go.

psalm-baseline.xml
```diff
@@ -149,16 +149,12 @@
...
-      <code><![CDATA[hasAttribute]]></code>
-      <code><![CDATA[hasAttribute]]></code>
-      <code><![CDATA[hasAttribute]]></code>
-      <code><![CDATA[hasAttribute]]></code>
-      <code><![CDATA[hasAttribute]]></code>
-      <code><![CDATA[hasAttribute]]></code>
-      <code><![CDATA[hasAttribute]]></code>
       <code><![CDATA[traverse]]></code>
     </ImpureMethodCall>
+    <PropertyTypeCoercion>
```
staabm, line 154: these baselined errors are related to https://github.com/Roave/BetterReflection/pull/1407#issuecomment-1996573163

Ocramius (approved):
```diff
- 4164 mutations were generated:
+ 4158 mutations were generated:
-    3333 mutants were killed
+    3329 mutants were killed
       0 mutants were configured to be ignored
      14 mutants were not covered by tests
-      10 covered mutants were not detected
+       8 covered mutants were not detected
       6 errors were encountered
       0 syntax errors were encountered
      14 time outs were encountered
     787 mutants required more time than configured
```

Also here, seems like a very good improvement with code reduction!

Ocramius:
Thanks @staabm!

Ocramius:
Btw: notice the reduction in mutations: that really says a lot whenever we refactor :)

PR #1406, opened 2024-03-13
Added test-case for deprecated phpstorm stub with patch-version

staabm:
tests the ternary in https://github.com/Roave/BetterReflection/blob/ab47d0b335a33ac211d5ad8758a48ed6be8622ca/src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php#L677 which wasn't covered before

PR #1408, opened 2024-03-17
Fix php version dependent constant deprecation

staabm:
the previous test was running against PHP 8.1, which we don't run in CI.
thats the reason why it [did not catch the implementation bug](https://github.com/Roave/BetterReflection/pull/1406#discussion_r1523065225)

### DIFF
--- a/src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
+++ b/src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
@@ -686,7 +686,7 @@ final class PhpStormStubsSourceStubber implements SourceStubber
             return false;
         }
 
-        if (preg_match('#@deprecated\s+(\d+)\.(\d+)(?:\.(\d+)?)$#m', $docComment->getText(), $matches) === 1) {
+        if (preg_match('#@deprecated\s+(\d+)\.(\d+)(?:\.(\d+))?$#m', $docComment->getText(), $matches) === 1) {
             $major     = $matches[1];
             $minor     = $matches[2];
             $patch     = $matches[3] ?? 0;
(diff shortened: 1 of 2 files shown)


staabm:
I am afk. Will fix it later today

test/unit/SourceLocator/SourceStubber/PhpStormStubsSourceStubberTest.php
```diff
@@ -468,37 +468,26 @@ public function testStubForConstantThatExists(): void
...
-
-/**
- * ID of "string" filter.
- * @link https://php.net/manual/en/filter.constants.php
- * @deprecated 8.1
- */
-\define('FILTER_SANITIZE_STRING', 513);
-EOT;
-        $stubData    = new StubData($exampleStub, 'filter');
+        $stubData = $this->sourceStubber->generateConstantStub('MT_RAND_PHP');
```
staabm, line 471: I am using a real constant now. using faked data the test did not cover reality

src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
```diff
@@ -664,14 +679,14 @@ private function isCoreExtension(string $extension): bool
         return in_array($extension, self::CORE_EXTENSIONS, true);
     }
 
-    private function isDeprecatedByPhpDocInPhpVersion(Node\Stmt\Const_ $node): bool
+    private function isDeprecatedByPhpDocInPhpVersion(Node\Expr\FuncCall $node): bool
```
staabm, line 682: in phpstorm stubs we use `define('x', 1)` and not `const $x=1`

Ocramius (approved):
Indeed, the infection report looks:

```diff
-4164 mutations were generated:
+4170 mutations were generated:
-    3315 mutants were killed
+    3333 mutants were killed
       0 mutants were configured to be ignored
-      14 mutants were not covered by tests
+       2 mutants were not covered by tests
      10 covered mutants were not detected
-       7 errors were encountered
+       6 errors were encountered
       0 syntax errors were encountered
      14 time outs were encountered
-     804 mutants required more time than configured
+     805 mutants required more time than configured
```

Ocramius:
Thanks @staabm!

PR #1409, opened 2024-03-21
Remove unused dependency (roave/signature)

janedbal:
Hi, I tested this repository with [composer-dependency-analyser](https://github.com/shipmonk-rnd/composer-dependency-analyser) and it **found unused dependency** (`roave/signature`). It looks like [it really is unused](https://github.com/search?q=repo%3ARoave%2FBetterReflection+%22Roave%5CSignature%22&type=code). ~~So I added this tool to CI to prevent similar issues in future.~~

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -6,8 +6,7 @@
         "php": "~8.2.0 || ~8.3.2",
         "ext-json": "*",
         "jetbrains/phpstorm-stubs": "2023.3",
-        "nikic/php-parser": "^5.0.2",
-        "roave/signature": "^1.8.0"
+        "nikic/php-parser": "^5.0.2"
     },
     "authors": [
         {
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "ca4b0964c7625dfb5e5ce68bb59844e4",
+    "content-hash": "a26d2a21d8dfafe19518da083fee4d65",
     "packages": [
         {
             "name": "jetbrains/phpstorm-stubs",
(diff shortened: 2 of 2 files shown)


.github/workflows/continuous-integration.yml
```diff
@@ -374,3 +374,43 @@ jobs:
 
       - name: "BC Checks"
         run: tools/vendor/bin/roave-backward-compatibility-check --from=origin/${{ github.base_ref }}
+
+  composer-depedency-analyser:
```
Ocramius: Since the dependency set is tiny, let's please drop this as well, as part of the change:

https://github.com/Roave/BetterReflection/blob/ee1ce025762a8784c89e0cb4ef59e85fecbd94ac/composer.json#L5-L11

Practically, just `composer.json` and `composer.lock` should change in this PR

Ocramius (changes requested):
Indeed, the dependency was used when `roave/better-reflection` was still generating code, which is no longer the case.

Let's please reduce the patch scope though: let's just remove the dependency, and remove the added dev-dependencies.

.github/workflows/continuous-integration.yml
janedbal: Sure, do you want me to move those changes to separate MR?

Ocramius: I'd just push to the same PR - I don't really want to add additional tooling for this edge case though, since dependency removal is not really such a frequent detail (happened in the last major version only)

Ocramius (approved):
Thanks @janedbal!

.github/workflows/continuous-integration.yml
janedbal: The tool actually detects much more, just this repo does not really suffer with other issues :) Ok then.

Ocramius: Yeah, but I do kinda know this stuff, as I help maintaining https://github.com/maglnet/ComposerRequireChecker ;-)

janedbal: Yeah I know, those tools compete now :) 

We built this one just because ComposerRequireChecker was taking too much time in our CI. But we were using it successfully for a long time, thanks for it!

PR #1410, opened 2024-03-22
Improve `ReflectionEnum->getCases()` performance

staabm:
the method is showing up in profiles of https://github.com/phpstan/phpstan/issues/10772
I am aware that this PR will not fix the root cause of the perf problem, but its a low hanging fruit on my way :)

grafik

lets remove the `array_map` magic and the unnecessary closure invocation

### DIFF
--- a/src/Reflection/Adapter/ReflectionEnum.php
+++ b/src/Reflection/Adapter/ReflectionEnum.php
@@ -530,14 +530,18 @@ final class ReflectionEnum extends CoreReflectionEnum
     /** @return list<ReflectionEnumUnitCase|ReflectionEnumBackedCase> */
     public function getCases(): array
     {
-        /** @psalm-suppress ImpureFunctionCall */
-        return array_map(function (BetterReflectionEnumCase $case): ReflectionEnumUnitCase|ReflectionEnumBackedCase {
-            if ($this->betterReflectionEnum->isBacked()) {
-                return new ReflectionEnumBackedCase($case);
-            }
+        $cases = array_values($this->betterReflectionEnum->getCases());
 
-            return new ReflectionEnumUnitCase($case);
-        }, array_values($this->betterReflectionEnum->getCases()));
+        $mappedCases = [];
+        foreach ($cases as $case) {
+            if ($this->betterReflectionEnum->isBacked()) {
+                $mappedCases[] = new ReflectionEnumBackedCase($case);
+            } else {
+                $mappedCases[] = new ReflectionEnumUnitCase($case);
+            }
+        }
+
+        return $mappedCases;
     }
 
     public function isBacked(): bool
--- a/src/Reflection/Adapter/ReflectionEnum.php
+++ b/src/Reflection/Adapter/ReflectionEnum.php
@@ -530,7 +530,7 @@ final class ReflectionEnum extends CoreReflectionEnum
     /** @return list<ReflectionEnumUnitCase|ReflectionEnumBackedCase> */
     public function getCases(): array
     {
-        $cases = array_values($this->betterReflectionEnum->getCases());
+        $cases = $this->betterReflectionEnum->getCases();
 
         $mappedCases = [];
         foreach ($cases as $case) {
--- a/src/Reflection/Adapter/ReflectionEnum.php
+++ b/src/Reflection/Adapter/ReflectionEnum.php
@@ -18,6 +18,7 @@ use Roave\BetterReflection\Reflection\ReflectionEnumCase as BetterReflectionEnum
 use Roave\BetterReflection\Reflection\ReflectionMethod as BetterReflectionMethod;
 use Roave\BetterReflection\Reflection\ReflectionProperty as BetterReflectionProperty;
 use Roave\BetterReflection\Util\FileHelper;
+use Roave\BetterReflection\Util\Memoize;
 use ValueError;
 
 use function array_combine;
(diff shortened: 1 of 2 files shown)


kukulich:
@staabm yes, the best optimization would be to remove the adapter layer from PHPStan :)

Ocramius:
I'm OK with merging this, but it will eventually be re-factored again to an `array_map()` later on anyway :| 

I'm convinced method tracing is actually inflating this more than it actually is relevant. 

I'd rather memoize `cases` instead, in future.

src/Reflection/Adapter/ReflectionEnum.php
```diff
@@ -530,14 +530,18 @@ public function getCase(string $name): ReflectionEnumUnitCase|ReflectionEnumBack
     /** @return list<ReflectionEnumUnitCase|ReflectionEnumBackedCase> */
     public function getCases(): array
     {
-        /** @psalm-suppress ImpureFunctionCall */
-        return array_map(function (BetterReflectionEnumCase $case): ReflectionEnumUnitCase|ReflectionEnumBackedCase {
+        $cases = array_values($this->betterReflectionEnum->getCases());
```
Ocramius: No need to `array_values()` since iteration below constructs a packed array

Ocramius:
> lets remove the `array_map()` magic and the unnecessary closure invocation

FWIW, `array_map()` is less magic than loops in general: it's just that the engine suuuuucks at handling it :-P

src/Reflection/Adapter/ReflectionEnum.php
```diff
@@ -530,14 +530,18 @@ public function getCase(string $name): ReflectionEnumUnitCase|ReflectionEnumBack
     /** @return list<ReflectionEnumUnitCase|ReflectionEnumBackedCase> */
     public function getCases(): array
     {
-        /** @psalm-suppress ImpureFunctionCall */
-        return array_map(function (BetterReflectionEnumCase $case): ReflectionEnumUnitCase|ReflectionEnumBackedCase {
+        $cases = $this->betterReflectionEnum->getCases();
+
+        $mappedCases = [];
+        foreach ($cases as $case) {
             if ($this->betterReflectionEnum->isBacked()) {
```
Ocramius: This method call is repeated in the loop: let's pull it out of here

staabm:
> FWIW, `array_map()` is less magic than loops in general

I feel `array_map` makes me read code backwards. I really like the simplicity of `foreach`. maybe I suck in reading it as much as the engine ;-)

Ocramius:
You need to spend a couple months writing FP (Rust, Haskell, clojure, etc.), and then C-style loops start looking like the garbled mess that they are ;-)

src/Reflection/Adapter/ReflectionEnum.php
```diff
@@ -530,14 +530,18 @@ public function getCase(string $name): ReflectionEnumUnitCase|ReflectionEnumBack
     /** @return list<ReflectionEnumUnitCase|ReflectionEnumBackedCase> */
```
Ocramius: Consideration here: we perhaps return `list<ReflectionEnumUnitCase>|list<ReflectionEnumBackedCase>` rather than `list<ReflectionEnumUnitCase|ReflectionEnumBackedCase>` :thinking:

staabm: AFAIK phpstan is not able to differentiate it, but I can do it if you prefer

staabm:
> I'd rather memoize `cases` instead, in future.

done

src/Reflection/Adapter/ReflectionEnum.php
Ocramius: It's more precise for the consumer regardless, no?

src/Reflection/Adapter/ReflectionEnum.php
```diff
@@ -527,17 +530,26 @@ public function getCase(string $name): ReflectionEnumUnitCase|ReflectionEnumBack
...
+            if ($isBacked) {
+                $mappedCases[] = new ReflectionEnumBackedCase($case);
+            } else {
+                $mappedCases[] = new ReflectionEnumUnitCase($case);
             }
+        }
 
-            return new ReflectionEnumUnitCase($case);
-        }, array_values($this->betterReflectionEnum->getCases()));
+        return $this->cases = $mappedCases;
```
Ocramius: Would likely need to move this to a mutable reference of some sort :thinking: 

I'll think about an approach :-)

Ocramius: Yeah, for this particular case, I'll probably write a little `Memoize<T>` utility that is internal to BR

src/Reflection/Adapter/ReflectionEnum.php
staabm: yeah. I already changed it silently :)

src/Reflection/Adapter/ReflectionEnum.php
staabm: this util needs to be a static call, as we can't store an instance of it in the `ReflectionEnum` class, right?

staabm: ahh got it. give me a minute

staabm:
please advice on how to make psalm happy or how/where to move this thing ;)

Ocramius:
I can try helping later tonight, but currently stuck with another issue at work.

ondrejmirtes:
My opinion - this kind of micro optimization isn't worth it. It's only showing 8 % improvement in PHPStan because it's working with huge amounts of data. Once the root cause in PHPStan is fixed then the improvement isn't going to be 8 %, but barely measurable.

The root cause was already fixed: https://github.com/phpstan/phpstan-src/pull/2985

Ocramius:
Most likely, but spiking memoization as a general outcome is still worth pursuing :D

staabm:
I am fine with closing it

src/Util/Memoize.php
```diff
@@ -0,0 +1,36 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Roave\BetterReflection\Util;
+
+/** @template T */
```
Ocramius: I'm thinking of this, meanwhile:

```php
/** 
 * @template T
 * @internal do not touch: you have been warned.
 */
final class Memoize
{
    private readonly mixed $cached;
    /** @param pure-Closure(): T $cb */
    public function __construct(private Closure|null $cb) {}
    public function get(): mixed {
        if ($this->cb) {
            $this->cached = ($this->cb)();
            $this->cb = null;
       }
        
        return $this->cached;
    }
}
```

This could be used general-purpose on most adapters, if needed.

Ocramius: mostly what you wrote, except I don't care if `$cached` is `null` (valid result)

staabm: thanks - copied it over. the SA tools don't like it ;)

Ocramius: Well, that's good :D 

We are indeed bending the rules here: hence the `@internal` stuff as well.

I'll likely need to take over on this: a good measure of error suppression will be needed, but the `impure-closure` bit is most annoying (it shouldn't consider it impure :thinking: )

Ocramius:
I tried rebasing this, but making `Memoize` pass a purity type check is quite a challenge :D

PR #1411, opened 2024-03-22
Use static closures where possible

staabm:
Reduce unnecessary `$this` bind; noticed while working on https://github.com/Roave/BetterReflection/pull/1410


staabm:
scratch that... doesn't work..

PR #1412, opened 2024-03-23
Failing test - CURLOPT_FTP_SSL being deprecated is false

ondrejmirtes:
The test passes if I revert https://github.com/Roave/BetterReflection/commit/145cdb2d78b7340d061a47114880dbed0a7f5367 from https://github.com/Roave/BetterReflection/pull/1408.

/cc @staabm 

This is how the constant looks like in phpstorm-stubs:

```php
/**
 * @link https://php.net/manual/en/curl.constants.php
 * @deprecated use <b>CURLOPT_USE_SSL</b> instead.
 */
define('CURLOPT_FTP_SSL', 119);
```


staabm:
picked your test into https://github.com/Roave/BetterReflection/pull/1413

thank you

PR #1413, opened 2024-03-23
Fix non-version dependent deprecations

staabm:
closes https://github.com/Roave/BetterReflection/pull/1412


Ocramius (approved):
Thanks @staabm!

PR #1414, opened 2024-03-26
More efficient `ReflectionClass#getParentClass()` - prevent loading all parent classes upfront

staabm:
Before this PR `getParentClass` was unnecessary greedy resolving the whole parent hierarchy even though only a single level would be sufficient

the method stood out on a recent blackfire profile. notice the "explosion" of calls from ~1.500 to ~67.400 in

grafik

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -100,6 +100,7 @@
       <code><![CDATA[reflectClass]]></code>
       <code><![CDATA[reflectClass]]></code>
       <code><![CDATA[reflectClass]]></code>
+      <code><![CDATA[reflectClass]]></code>
       <code><![CDATA[toLowerString]]></code>
     </ImpureMethodCall>
   </file>
--- a/src/Reflection/ReflectionClass.php
+++ b/src/Reflection/ReflectionClass.php
@@ -1086,12 +1086,16 @@ class ReflectionClass implements Reflection
      */
     public function getParentClass(): ReflectionClass|null
     {
-        $parentClass = $this->getParentClasses()[0] ?? null;
-
-        if ($parentClass === null) {
+        $parentClassName = $this->getParentClassName();
+        if ($parentClassName === null) {
             return null;
         }
 
+        if ($this->name === $parentClassName) {
+            throw CircularReference::fromClassName($parentClassName);
+        }
+
+        $parentClass = $this->reflector->reflectClass($parentClassName);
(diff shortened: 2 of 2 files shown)


Ocramius (approved):
Excellent!

@staabm food for thought here, but how would we ensure no regressions on these cases? :thinkin

staabm:
We could assert how often methods are called internally, but this would tighten the test to the impl

Ocramius:
I was thinking of counting opcode executions via phpbench or similar, in future :thinking:

staabm:
Or we could add blackfire performance tests with assertions on memory consumption/cpu time

Ocramius:
> assertions on memory consumption/cpu time
 
 Unlikely runnable in CI due to missing stable workers :thinking: 
 
 Paging @dantleech for ideas :D

DavidBadura:
We benchmark every pull request to see if anything has changed in CPU/Memory consumption. 

https://github.com/patchlevel/event-sourcing/pull/552#issuecomment-2018344313

At the moment we have not yet defined a threshold (diff in percent) to make the test fail. 
But we wanted to explore that too.

DavidBadura:
The trick is to use the same runner to run the base (old) benchmark and the new benchmark. 
Yes, there are still some differences because of the load on the runners, but the difference is still very small. 
It is important that there are not two different runners. Maybe this helps.

Ocramius:
@DavidBadura I was really hoping there would be a deterministic instruction counter somewhere: CPUs are shared on runners, which make regression testing of benchmarks on shared infrastructure almost impossible :cold_sweat:

dantleech:
> Paging @dantleech for ideas :D

I'm honestly not sure how much of an indicator nb. opcodes executed is to performance, as I guess some are more expensive than others, although it would be a solid metric for complexity and I guess a massive jump is likely going to be an issue, I might look into adding it to PHPBench at some point.

You can never be 100% confident about benchmarks as the universe is constantly changing state, but you can get a good level of confidence with statistics, probably even on a GH runner given enough samples and enough time. But PHPBench's analysis isn't optimal for that. I've had reasonable success using @DavidBadura 's method, failing if the result is > +10% f.e. and caught some regressions, although I'm not doing anything like that right now (even +100% is valuable)

I think what would be really useful is a service to aggregate the data and see it change over time, and such a service could also analyse the results vs. the enviornments and send alerts etc.

PR #1416, opened 2024-03-27
Prevent autoloading when determinig constant values

### LINKED ISSUE #1415: `PhpStormStubsSourceStubber` calls into application autoloader
while debugging today, I realized that when BetterReflection is building up its reflection knowledge based on phpstorm stubs, it is invoking application autoloaders.

I found stacktraces like

grafik

in which you can see the top-level frame `Rocket::autoload`. thats my application autoloader which is registered via phpstan bootstrap file.

my questions:
- should BetterReflection trigger autoloading while analyzing phpstorm stubs?
- if autoloading is really needed, should better reflection isolate the autoloading involved from side-effects like a application level autoloader?

I came to this question, because our application autoloading needs todo some gymnastics and it is invoked a lot for all kind of classes which `BetterReflection` is locating while parsing stub files.

looking deeper it might even be the case this only happens from within PHPStan.. not sure yet

staabm:
closes https://github.com/Roave/BetterReflection/issues/1415

### DIFF
--- a/src/SourceLocator/SourceStubber/PhpStormStubs/CachingVisitor.php
+++ b/src/SourceLocator/SourceStubber/PhpStormStubs/CachingVisitor.php
@@ -13,8 +13,11 @@ use Roave\BetterReflection\Util\ConstantNodeChecker;
 
 use function array_key_exists;
 use function assert;
+use function class_exists;
 use function constant;
+use function count;
 use function defined;
+use function explode;
 use function in_array;
 use function is_resource;
 use function sprintf;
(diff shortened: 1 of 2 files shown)


src/SourceLocator/SourceStubber/PhpStormStubs/CachingVisitor.php
```diff
@@ -195,6 +203,16 @@ public function clearNodes(): void
      */
     private function updateConstantValue(Node\Expr\FuncCall|Node\Const_ $node, string $constantName): void
     {
+        // prevent autoloading while discovering class constants
+        $parts = explode('::', $constantName, 2);
```
kukulich, line 202: 👍

src/SourceLocator/SourceStubber/PhpStormStubs/CachingVisitor.php
```diff
@@ -195,6 +198,16 @@ public function clearNodes(): void
      */
     private function updateConstantValue(Node\Expr\FuncCall|Node\Const_ $node, string $constantName): void
     {
+        // prevent autoloading while discovering class constants
```
Ocramius, line 201: BTW, thanks for commenting on this :-)

Ocramius (approved):
Thanks @staabm!

PR #1418, opened 2024-03-29
Assert parsing PHPStorm stubs will never trigger autoload

staabm:
looking again at my patch in https://github.com/Roave/BetterReflection/commit/b9e11a2f6c51c54a7396d86a1f39beb88e2ecd7a#diff-41fe06041ece0f581f2a6b1bb811daf7e9d02d18efd83a1ca636caf6ea7e2330R206 I wondered whether we should also use `interface_exists()` and or `trait_exists()` to prevent other cases of autoading.

I couldn't reproduce a case with phpstorm stubs though, which would trigger autoloading for a interface or trait-constant.
therefore I figured we should just add a security measure, to make sure non of the tests in `PhpStormStubsSourceStubberTest` will unnoticed trigger autoloading.


Ocramius (approved):

Ocramius:
Thanks @staabm

PR #1419, opened 2024-04-02
Default value `null` does not make promoted property type nullable

ondrejmirtes:


ondrejmirtes:
This is the relevant PHP runtime behaviour: https://3v4l.org/RQVHb

Ocramius (approved):
Wow, heck of a quirky behaviour!

PR #1423, opened 2024-04-21
Update CI jobs to PHP 8.3

staabm:
I left the coding standard job on 8.2 since this is the lowest supported version and we should make sure cs-fixing does not introduce features which are not compatible with the lowest supported one

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -192,7 +192,7 @@ jobs:
         dependencies:
           - "locked"
         php-version:
-          - "8.2"
+          - "8.3"
         operating-system:
           - "ubuntu-latest"
           - "windows-latest"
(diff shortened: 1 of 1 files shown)


kukulich:
PHP version for CS is set here: https://github.com/Roave/BetterReflection/blob/3a6f27b020d7f144b2c2e86cfed60ef5df6e55d8/phpcs.xml.dist#L3

And it´s wrong 🙈

staabm:
> And it´s wrong 🙈

good catch, fix in https://github.com/Roave/BetterReflection/pull/1425

staabm:
> PHP version for CS is set here:

PHPStan is recommended to be run in the very same major/minor-version you want to use at runtime later on.
Is PHPCS more forgiving and mixing the versions is not strictly necessary?

Ocramius:
> PHPStan is recommended to be run in the very same major/minor-version you want to use at runtime later on.

We do CS on the lowest version, to prevent breakages due to newer php version replacements.

We do SA on the newest php version, to ensure types still match the engine development.

Ocramius (approved):
🚢 here meanwhile

PR #1424, opened 2024-04-21
Utilize more readable string functions

staabm:


Ocramius:
Btw, perhaps we need a rector action, since these replacements can often be automated

src/Identifier/Identifier.php
```diff
@@ -26,7 +26,7 @@ public function __construct(string $name, private IdentifierType $type)
         if (
             $name === self::WILDCARD
             || $name === ReflectionFunction::CLOSURE_NAME
-            || strpos($name, ReflectionClass::ANONYMOUS_CLASS_NAME_PREFIX) === 0
+            || str_starts_with($name, ReflectionClass::ANONYMOUS_CLASS_NAME_PREFIX)
```
Ocramius, line 29: Remember to import yer functions 😛

staabm:
> Btw, perhaps we need a rector action, since these replacements can often be automated

I did it locally with rector :) .. and stumbled over a [bug in rector](https://github.com/rectorphp/rector-src/pull/5838), which I fixed atm

src/Identifier/Identifier.php
staabm: 😇

Ocramius:
Thanks @staabm!

PR #1425, opened 2024-04-21
Raise min phpcs `php_version` to 8.2

staabm:
refs https://github.com/Roave/BetterReflection/pull/1423#issuecomment-2068114870

### DIFF
--- a/src/Reflection/Adapter/ReflectionIntersectionType.php
+++ b/src/Reflection/Adapter/ReflectionIntersectionType.php
@@ -35,7 +35,7 @@ class ReflectionIntersectionType extends CoreReflectionIntersectionType
         return $this->betterReflectionType->__toString();
     }
 
-    public function allowsNull(): bool
+    public function allowsNull(): false
     {
         return $this->betterReflectionType->allowsNull();
     }
--- a/src/Reflection/Adapter/ReflectionIntersectionType.php
+++ b/src/Reflection/Adapter/ReflectionIntersectionType.php
@@ -35,7 +35,6 @@ class ReflectionIntersectionType extends CoreReflectionIntersectionType
         return $this->betterReflectionType->__toString();
     }
 
-    /** @return false */
     public function allowsNull(): bool
     {
         return $this->betterReflectionType->allowsNull();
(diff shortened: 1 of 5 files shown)


src/Reflection/Adapter/ReflectionIntersectionType.php
```diff
@@ -35,7 +35,6 @@ public function __toString(): string
         return $this->betterReflectionType->__toString();
     }
 
-    /** @return false */
```
Ocramius, line 38: This stricter type is intentional 🤔

src/Reflection/ReflectionIntersectionType.php
```diff
@@ -53,7 +53,6 @@ public function getTypes(): array
         return $this->types;
     }
 
-    /** @return false */
```
Ocramius, line 56: This stricter type is intentional 🤔

src/Reflection/Adapter/ReflectionIntersectionType.php
staabm: it is removed by phpcs,.. I guess we can't change the native type because BC?

Ocramius: We can 👍

staabm: done

Ocramius (commented):
Groan:

```
There was 1 error:

1) Roave\BetterReflectionTest\Reflection\Adapter\ReflectionIntersectionTypeTest::testAdapterMethods with data set #1 ('allowsNull', null, true, [])
PHPUnit\Framework\MockObject\IncompatibleReturnValueException: Method allowsNull may not return value of type bool, its declared return type is "false"

/home/runner/work/BetterReflection/BetterReflection/test/unit/Reflection/Adapter/ReflectionIntersectionTypeTest.php:61

ERRORS!
Tests: 13152, Assertions: 73866, Errors: 1, Skipped: 1.
```

I think our test is too strict here.

Wdyt, @kukulich?

staabm:
hmm I think the test is buggy.. I wonder why it worked before

Ocramius:
I think it expects exactly the same types, although we intentionally do stricter types here

Ocramius (approved):
Thanks @staabm!

PR #1426, opened 2024-04-21
Finalize classes

staabm:
I wonder whether we want to disallow subclassing..?


Ocramius:
Not always, but we should evaluate it.

PR #1428, opened 2024-05-31
Improve `getNamespacedName()` return type

staabm:
triggered by https://github.com/phpstan/phpstan-src/pull/3013/files#r1564724535

refs https://github.com/phpstan/phpstan-src/pull/3013

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -49,6 +49,10 @@
     "conflict": {
        "thecodingmachine/safe": "<1.1.3"
     },
+    "scripts": {
+        "psalm": "tools/vendor/bin/psalm",
+        "psalm-baseline": "tools/vendor/bin/psalm --set-baseline=psalm-baseline.xml"
+    },
     "suggest": {
         "composer/composer": "Required to use the ComposerSourceLocator"
     },
--- a/composer.json
+++ b/composer.json
@@ -50,6 +50,8 @@
        "thecodingmachine/safe": "<1.1.3"
     },
     "scripts": {
+        "cs": "tools/vendor/bin/phpcs",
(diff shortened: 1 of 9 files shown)


composer.json
```diff
@@ -49,6 +49,10 @@
     "conflict": {
        "thecodingmachine/safe": "<1.1.3"
     },
+    "scripts": {
```
staabm, line 52: I am not used to the psalm commands.. have to google them all the time.. added them here to ease contributing for non-psalm experienced people like me

Ocramius: Not very happy to have `composer` as a wrapper to run scripts, but it's a non-blocker for this patch

src/SourceLocator/Ast/Strategy/NodeToReflection.php
```diff
@@ -28,6 +29,10 @@ public function __invoke(
     ): ReflectionClass|ReflectionConstant|ReflectionFunction {
         $namespaceName = $namespace?->name?->name;
 
+        if ($namespaceName === '') {
+            throw new LogicException('Namespace name should never be empty');
+        }
```
Ocramius, line 34: Is this something we can test at all? Or is it an upstream type refinement required?

staabm: the upstream type is already refined, but the change itself is not yet released

https://github.com/nikic/PHP-Parser/pull/993

Ocramius: Let's then roll with this as-is: a lock file update will expose the type contradiction later on 👍

Ocramius:
Thanks @staabm!

PR #1435, opened 2024-06-18
Removed need for manually checking for parsing errors on `composer.(json|lock)` and `installed.json`

Ocramius:
This marginally reduces code size, and reduces `mixed` types a bit, but would benefit
tons from removing direct usage of `json_decode()`, which is an unsafe construct, when
the results of its calls go un-checked.

Ideally, we'd use `azjezz/psl`, but that library comes with added dependencies that we can't afford
here, for now.

### DIFF
--- a/src/SourceLocator/Type/Composer/Factory/MakeLocatorForComposerJson.php
+++ b/src/SourceLocator/Type/Composer/Factory/MakeLocatorForComposerJson.php
@@ -21,12 +21,13 @@ use function array_merge;
 use function array_values;
 use function assert;
 use function file_get_contents;
-use function is_array;
 use function is_dir;
 use function is_file;
 use function is_string;
 use function json_decode;
 use function realpath;
+
+use const JSON_THROW_ON_ERROR;
 
 /**
  * @psalm-type ComposerAutoload array{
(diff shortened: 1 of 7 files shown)


src/SourceLocator/Type/Composer/Factory/MakeLocatorForComposerJson.php
```diff
@@ -65,12 +65,8 @@ public function __invoke(string $installationPath, Locator $astLocator): SourceL
...
         assert(is_string($composerJsonContent));
 
-        /** @psalm-var array{autoload: ComposerAutoload}|null $composer */
-        $composer = json_decode($composerJsonContent, true);
-
-        if (! is_array($composer)) {
-            throw FailedToParseJson::inFile($composerJsonPath);
-        }
+        /** @psalm-var array{autoload: ComposerAutoload} $composer */
+        $composer = json_decode($composerJsonContent, true, flags: JSON_THROW_ON_ERROR);
```
Ocramius, line 69: Mostly interested if `psalm --stats` says that this will improve our types or not: I believe `mixed` is still considered here.

Ocramius:
```
97.980% src/SourceLocator/Type/Composer/Factory/MakeLocatorForComposerJson.php (2 mixed)
98.942% src/SourceLocator/Type/Composer/Factory/MakeLocatorForComposerJsonAndInstalledJson.php (2 mixed)
98.742% src/SourceLocator/Type/Composer/Factory/MakeLocatorForInstalledJson.php (2 mixed)
```

No real improvement here :-\

kukulich (approved):
👍

PR #1441, opened 2024-08-01
Remove unnecessary constants

### LINKED ISSUE #1440: ReflectionClass::IS_READONLY is a typed class constant in PHP 8.4
PHP 8.4 made the ReflectionClass::IS_READONLY class constant typed in https://github.com/php/php-src/pull/12378. Since `Roave\BetterReflection\Reflection\Adapter\ReflectionClass` overrides it, the following exception is thrown when using PHP 8.4:

```
Type of Roave\BetterReflection\Reflection\Adapter\ReflectionClass::IS_READONLY must be compatible with ReflectionClass::IS_READONLY of type int
```

akmandev:


Ocramius:
Handled in #1442

PR #1442, opened 2024-08-03
Compatibility with typed constants in PHP 8.4

kukulich:
Fixes https://github.com/Roave/BetterReflection/issues/1440

PR #1443, opened 2024-08-03
Remove unnecessary constants

kukulich:
Fix of https://github.com/Roave/BetterReflection/pull/1442

Yes, @akmandev was right!

The broken build in https://github.com/Roave/BetterReflection/pull/1441 has confused me... 😞  The constants really already exist in 8.2 and we can safely remove them.

The really problem was broken PHPStan configuration. Thanks to @ondrejmirtes to figured out.

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -3,8 +3,6 @@ includes:
     - tools/vendor/phpstan/phpstan-phpunit/rules.neon
 
 parameters:
-    phpVersion: 80100
-
     level: 6
 
     paths:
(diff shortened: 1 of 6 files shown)


phpstan.neon
```diff
@@ -3,7 +3,7 @@ includes:
     - tools/vendor/phpstan/phpstan-phpunit/rules.neon
 
 parameters:
-    phpVersion: 80100
+    phpVersion: 80200
```
ondrejmirtes: You can remove this line because PHPStan will follow this too: https://github.com/Roave/BetterReflection/blob/7f16c808bb64f34b172d14966e348ccdcd97f8b8/composer.json#L64

ondrejmirtes (changes requested):

Ocramius (approved):
Excellent, thanks!

Ocramius:
@kukulich merge when you think it's ready

PR #1444, opened 2024-09-02
Failing test: `ReflectionClass::getTraitAliases` does not maintain case

janedbal:
I'm not sure if this is intentional, but it is harder to work with and does [not comply with the phpdoc of the `getTraitAliases` method](https://github.com/Roave/BetterReflection/blob/5ea77d6/src/Reflection/ReflectionClass.php#L1359).

### DIFF
--- a/test/unit/Fixture/TraitFixture.php
+++ b/test/unit/Fixture/TraitFixture.php
@@ -30,6 +30,7 @@ trait TraitFixtureTraitC
     public function a() {}
     public function b() {}
     public function c() {}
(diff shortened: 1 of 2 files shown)


Ocramius:
If I see this correctly, we need to look for `strtolower()` usages in this codebase

kukulich:
It's caused by this commit: https://github.com/Roave/BetterReflection/commit/0ba851cb

I will look at it.

janedbal:
Thank you!

PR #1445, opened 2024-09-03
Fix support of Enum stubs from PHPStorm

VincentLanglet:
Hi @Ocramius,

An (old) assert was added by @kukulich in https://github.com/Roave/BetterReflection/commit/933dcb9c61745c29702199f0a1260d36c083cda0

This one is not true anymore, cause PHPStorm stubs use `@since` on Enum like
https://github.com/JetBrains/phpstorm-stubs/commit/2630c4604859a43fe761108360f4b4352056aa74#diff-11cfa64a565c311363b652c7494df2be636c9934afac14510bdd70e50d7044ceR6

Such error was discovered by the PHPStan CI which run on PHP 8.4 https://github.com/phpstan/phpstan-src/actions/runs/10680283358/job/29601421334?pr=3393 ; but I'm unsure how much work it would require so far to have a full Ci running on PHP 8.4. I can try in another PR if you want to.

### DIFF
--- a/src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
+++ b/src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
@@ -514,7 +514,7 @@ final class PhpStormStubsSourceStubber implements SourceStubber
     {
         $newStmts = [];
         foreach ($stmts as $stmt) {
-            assert($stmt instanceof Node\Stmt\ClassConst || $stmt instanceof Node\Stmt\Property || $stmt instanceof Node\Stmt\ClassMethod);
+            assert($stmt instanceof Node\Stmt\ClassConst || $stmt instanceof Node\Stmt\Property || $stmt instanceof Node\Stmt\ClassMethod || $stmt instanceof Node\Stmt\EnumCase);
 
             if (! $this->isSupportedInPhpVersion($stmt)) {
                 continue;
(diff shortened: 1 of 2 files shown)


Ocramius:
@VincentLanglet are the stubs in here already bumped here? We bump them regularly, but it needs adjustments in `composer.json` too

VincentLanglet:
> @VincentLanglet are the stubs in here already bumped here? We bump them regularly, but it needs adjustments in `composer.json` too

No, those stubs are on master branch and but no tags are created yes.
https://github.com/JetBrains/phpstorm-stubs/compare/v2024.2...master

Should I target the latest commit ?

Ocramius:
> Should I target the latest commit ?

For trying them out: yes.

For merging this (once fixed): needs to be a tagged version.

VincentLanglet:
> > Should I target the latest commit ?
> 
> For trying them out: yes.

Done. I also tried to add a PHP 8.4 Ci (for testing purpose), because the error occured only on PHP 8.4 PHPStan ci. (cf https://github.com/phpstan/phpstan-src/actions/runs/10680283358/job/29601421334?pr=3393)

(Not sure if there is an easy way to avoid waiting for workflow approval at every commit so I try to offer an easy-merge PR https://github.com/Roave/BetterReflection/pull/1446)

kukulich:
@VincentLanglet You can remove the PHP 8.4 changes and use [random extension](https://github.com/JetBrains/phpstorm-stubs/blob/master/random/random.php#L252-L255) for unit tests.

composer.json
```diff
@@ -5,7 +5,7 @@
     "require": {
         "php": "~8.2.0 || ~8.3.2",
         "ext-json": "*",
-        "jetbrains/phpstorm-stubs": "2024.2",
+        "jetbrains/phpstorm-stubs": "dev-master#56f6b9e55f5885e651553843a1aaf9ec9c586c04",
```
Ocramius: Tracking this TODO: to be moved to a tag, when available, before merging

Ocramius (approved):

VincentLanglet:
I reworked the PR @Ocramius,

Since I dunno when a new tag of PHPStorm stubs will be released, 
- I added a test with the enum case of random extension => tests are failing
- I added the fix in a new commit => tests are green

This way it should be merge-able now, without waiting for a PHPStorm stub new tag. 
(and https://github.com/ondrejmirtes/BetterReflection will be able to rebase the fix too)

Also, I discovered you'll need a fix for SplObjectStorage in PHP 8.4 so I opened another PR https://github.com/Roave/BetterReflection/pull/1448.

Ocramius (approved):
Thanks @VincentLanglet!

PR #1446, opened 2024-09-04
Add missing entries to gitattributes

VincentLanglet:
Not sure if it's done on purpose or just an oversight.


Ocramius (approved):
Thanks @VincentLanglet

PR #1452, opened 2024-09-06
`ReflectionClass::getTraitAliases()` should maintain case

kukulich:
Fixes #1444

### DIFF
--- a/src/Reflection/ReflectionClass.php
+++ b/src/Reflection/ReflectionClass.php
@@ -119,7 +119,13 @@ class ReflectionClass implements Reflection
     /** @var array<non-empty-string, ReflectionMethod> */
     private array $immediateMethods;
 
-    /** @var array{aliases: array<non-empty-string, non-empty-string>, modifiers: array<non-empty-string, int-mask-of<ReflectionMethodAdapter::IS_*>>, precedences: array<non-empty-string, non-empty-string>} */
+    /** @var array{
+     *     aliases: array<non-empty-string, non-empty-string>,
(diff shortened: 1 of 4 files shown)


src/Reflection/ReflectionClass.php
```diff
@@ -358,16 +364,16 @@ public function getExtensionName(): string|null
...
-            if ($this->traitsData['modifiers'][$methodHash] & Modifiers::VISIBILITY_MASK) {
-                $methodModifiers = ($methodModifiers & ~ Modifiers::VISIBILITY_MASK) | $this->traitsData['modifiers'][$methodHash];
+            if ($this->traitsData['modifiers'][$lowerCasedMethodHash] & Modifiers::VISIBILITY_MASK) {
+                $methodModifiers = ($methodModifiers & ~ Modifiers::VISIBILITY_MASK) | $this->traitsData['modifiers'][$lowerCasedMethodHash];
             }
 
-            if ($this->traitsData['modifiers'][$methodHash] & Modifiers::FINAL) {
+            if ($this->traitsData['modifiers'][$lowerCasedMethodHash] & Modifiers::FINAL) {
                 $methodModifiers |= Modifiers::FINAL;
             }
```
Ocramius, line 378: I think this calls for a small `@internal` class that keeps this trait data: the fact that we need to look up by lowercase hash seems to be something akin a `Map<TTraitMethod, Modifiers>`.

Ocramius (approved):
Let's just ship this as-is meanwhile :D

Ocramius:
Thanks @kukulich!

PR #1453, opened 2024-09-06
Failing test: missing class method when trait aliases are used

janedbal:
See real reflection result: https://3v4l.org/TcNgp

### DIFF
--- a/test/unit/Fixture/TraitFixture.php
+++ b/test/unit/Fixture/TraitFixture.php
@@ -133,3 +133,22 @@ class TraitFixtureF
     }
 
 }
+
+
+
+trait Trait1FixtureG {
+    public function method1() {}
+    public function method2() {}
+}
+
+trait Trait2FixtureG {
+    public function method3() {}
+    public function method4() {}
+}
+
+class ClassFixtureG {
+    use Trait1FixtureG, Trait2FixtureG {
+        method1 as alias1;
+        method3 as alias3;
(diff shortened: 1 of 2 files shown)


janedbal:
For a reference, it is causing false-positives in our PHPStan-based dead-code-detector: https://github.com/shipmonk-rnd/dead-code-detector/pull/70/files#diff-16ea0828915ea4592d186f3dfab87f40606c659da6d84ed4204d69a8d34cf058R6

Ocramius:
I'm really really really convinced that traits should never have existed in PHP :D

janedbal:
> I'm really really really convinced that traits should never have existed in PHP :D

Totally agree! The dead-code-detector would have been 3 times less complex without them :)

test/unit/Fixture/TraitFixture.php
```diff
@@ -133,3 +133,22 @@ class TraitFixtureF
...
+
+trait Trait2FixtureG {
+    public function method3() {}
+    public function method4() {}
+}
+
+class ClassFixtureG {
+    use Trait1FixtureG, Trait2FixtureG {
+        method1 as alias1;
+        method3 as alias3;
```
Ocramius, line 152: BTW, didn't even know this multi-trait-import-with-alias syntax was legal :scream:

janedbal:
Thank you!

PR #1460, opened 2024-11-15
Fixed missing class method when multiple trait aliases are used

kukulich:
Fixes #1453

Yes, the code is ugly but the PHP feature is ugly too... I really hope I will not see any other trait bug in near future!

### DIFF
--- a/src/Reflection/ReflectionClass.php
+++ b/src/Reflection/ReflectionClass.php
@@ -53,7 +53,6 @@ use function array_reverse;
 use function array_slice;
 use function array_values;
 use function assert;
-use function end;
 use function in_array;
 use function is_int;
 use function is_string;
@@ -120,7 +119,7 @@ class ReflectionClass implements Reflection
     private array $immediateMethods;
 
     /** @var array{
-     *     aliases: array<non-empty-string, non-empty-string>,
+     *     aliases: array<trait-string, list<array{alias: non-empty-string, method: non-empty-string, hash: non-empty-string}>>,
      *     modifiers: array<non-empty-string, int-mask-of<ReflectionMethodAdapter::IS_*>>,
      *     precedences: array<non-empty-string, non-empty-string>,
      *     hashes: array<non-empty-string, non-empty-string>,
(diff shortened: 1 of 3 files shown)


src/Reflection/ReflectionClass.php
```diff
@@ -1366,18 +1377,36 @@ static function (ReflectionClass $trait): string {
...
+
+        $traits       = $this->getTraits();
+        $traitAliases = [];
+
+        foreach ($traits as $trait) {
+            if (! array_key_exists($trait->getName(), $this->traitsData['aliases'])) {
+                continue;
+            }
+
+            foreach ($this->traitsData['aliases'][$trait->getName()] as $traitAliasDefinition) {
```
Ocramius: I'm quite worried about nesting multiple `foreach` blocks here: the code complexity is exploding quite a bit

src/Reflection/ReflectionClass.php
```diff
@@ -1438,37 +1467,37 @@ private function computeTraitsData(ClassNode|InterfaceNode|TraitNode|EnumNode $n
...
-                $traitsData['hashes'][$lowerCasedMethodHash] = $methodHash;
-
-                if ($adaptation instanceof Node\Stmt\TraitUseAdaptation\Alias) {
-                    if ($adaptation->newModifier !== null) {
-                        /** @var int-mask-of<ReflectionMethodAdapter::IS_*> $modifier */
-                        $modifier                                       = $adaptation->newModifier;
-                        $traitsData['modifiers'][$lowerCasedMethodHash] = $modifier;
+                $usedTraits = $adaptation->trait !== null ? [$adaptation->trait] : $traitNames;
+
+                foreach ($usedTraits as $usedTrait) {
```
Ocramius: Same remarks about foreach complexity

src/Reflection/ReflectionClass.php
```diff
@@ -1418,7 +1447,7 @@ public function getTraitAliases(): array
      *   //   ['MyTrait1::foo' => 'MyTrait2::foo']
      *
      * @return array{
-     *     aliases: array<non-empty-string, non-empty-string>,
+     *     aliases: array<non-empty-string, list<array{alias: non-empty-string, method: non-empty-string, hash: non-empty-string}>>,
```
Ocramius: From a consumer of this method perspective, what is this hash?

test/unit/Fixture/TraitFixture.php
```diff
@@ -135,3 +135,20 @@ class TraitFixtureF
     }
 
 }
+
+trait Trait1FixtureG {
```
Ocramius: IMO worth moving to a string source locator within the test: it's only used within that one test, and shouldn't be mixed with the rest

Ocramius (approved):
The test additions and clarity thereof is fantastic : my only concerns are with code complexity, and this code is exploding with them.

I would solit this part, even at major performance costs, just ti keep it manageable.

Traits are an edge case in good design: they are a problem that keeps haunting us, and I'd like to push them in their own "naughty" corner.

src/Reflection/ReflectionClass.php
kukulich: It's more described above - the hash was already there before but now I need more data about aliases.

test/unit/Fixture/TraitFixture.php
kukulich: Done.

src/Reflection/ReflectionClass.php
kukulich: I've just tried to refactored it a little.

src/Reflection/ReflectionClass.php
kukulich: I've just tried to refactored it a little.

Ocramius (approved):
Thanks @kukulich, this looks much much better!

Ocramius:
🚢 thanks @kukulich!

PR #1461, opened 2024-11-20
Update README.md

rpmcmurphy:
Typo fixed in a line from 'moar to more'.


Ocramius:
It's intentional 😁

rpmcmurphy:
Wow. I did not guess that at all. Thanks.

PR #1462, opened 2024-11-22
PHP 8.4 support

### LINKED ISSUE #1455: Does not work with php 8.4
Can 8.4 be added?  Right now, composer.json locks this library to 8.2 or 8.3

kukulich:


Ocramius:
@kukulich I've only seen this patch from afar: do ping when you believe it should be looked at in detail (currently quite busy with work, sorry)

kukulich:
@Ocramius It should be ready to review :) I hope it's full support of PHP 8.4.

ondrejmirtes:
FYI PHPStan is living on the edge - I rebased my fork on top of this PR and already using that in 2.0.x 😂 Might report some issues in a few days when I get the chance to test it thoroughly.

Ocramius:
Yeah, I imagined :D 

I just got off work now, so I'll probably be able to review tomorrow :)

phpstan.neon
```diff
@@ -25,6 +25,10 @@ parameters:
         - phar://%currentWorkingDirectory%/test/unit/Fixture/autoload.phar/vendor/autoload.php
 
     ignoreErrors:
+        -
+            identifier: class.notFound
+            message: '#(unknown class|invalid type) PropertyHookType#'
```
Ocramius, line 30: Is this needed because we run PHPStan on lower versions too? :thinking:

src/Reflection/Adapter/ReflectionAttribute.php
```diff
@@ -54,4 +58,13 @@ public function __toString(): string
     {
         return $this->betterReflectionAttribute->__toString();
     }
+
+    public function __get(string $name): mixed
```
Ocramius, line 63: Did we add this because upstream declares the variable? Can't we set `name` in the constructor? :thinking:

src/Reflection/Adapter/ReflectionClass.php
```diff
@@ -33,6 +33,12 @@
  */
 final class ReflectionClass extends CoreReflectionClass
 {
+    /** @internal */
+    public const SKIP_INITIALIZATION_ON_SERIALIZE_COMPATIBILITY = 8;
+
+    /** @internal */
+    public const SKIP_DESTRUCTOR_COMPATIBILITY = 16;
```
Ocramius: I wonder if we should `@inheritDoc` on these, so we understand they come from above

src/Reflection/Adapter/ReflectionClass.php
```diff
@@ -450,6 +456,50 @@ public function newInstanceArgs(array|null $args = null): object
         throw new Exception\NotImplemented('Not implemented');
     }
 
+    /** @param int-mask-of<self::SKIP_*> $options */
+    public function newLazyGhost(callable $initializer, int $options = 0): object
+    {
+        throw new Exception\NotImplemented('Not implemented');
```
Ocramius: I think we should throw `Exception\NotImplementedByDesign()` or such, or perhaps `Exception\NotImplementedBecauseItTriggersAutoloading()` or such.

This would signal (to other people using the library) that this will never work in BetterReflection.

I wonder if we should also do an `@return never` on each of these

src/Reflection/Adapter/ReflectionEnum.php
```diff
@@ -376,6 +376,50 @@ public function newInstanceArgs(array|null $args = null): object
         throw new Exception\NotImplemented('Not implemented');
     }
 
+    /** @param int-mask-of<ReflectionClass::SKIP_*> $options */
+    public function newLazyGhost(callable $initializer, int $options = 0): object
+    {
+        throw new Exception\NotImplemented('Not implemented');
```
Ocramius: Similar to above: I think we probably need something slightly better than `NotImplemented`, just to signal a design decision, rather than a "TODO"

src/Reflection/Adapter/ReflectionEnum.php
```diff
@@ -376,6 +376,50 @@ public function newInstanceArgs(array|null $args = null): object
         throw new Exception\NotImplemented('Not implemented');
     }
 
+    /** @param int-mask-of<ReflectionClass::SKIP_*> $options */
```
Ocramius: Similar to above: wondering about `@return never` + `@throws` here

src/Reflection/Adapter/ReflectionNamedType.php
```diff
@@ -44,12 +63,17 @@ public function allowsNull(): bool
 
     public function isBuiltin(): bool
     {
-        $type = strtolower($this->betterReflectionType->getName());
+        return $this->isBuiltin;
+    }
+
+    private function computeIsBuiltin(string $namedType, bool $isBuiltin): bool
```
Ocramius: This method is only used in `__construct`: I'd endorse making it `static`

src/Reflection/Adapter/ReflectionObject.php
```diff
@@ -396,6 +396,50 @@ public function newInstanceArgs(array|null $args = null): ReflectionObject
         throw new Exception\NotImplemented('Not implemented');
     }
 
+    /** @param int-mask-of<ReflectionClass::SKIP_*> $options */
+    public function newLazyGhost(callable $initializer, int $options = 0): object
+    {
+        throw new Exception\NotImplemented('Not implemented');
```
Ocramius: Similar notes around `NotImplemented`

src/Reflection/ReflectionClass.php
```diff
@@ -596,6 +597,8 @@ private function addEnumMethods(EnumNode $node, array $methods): array
     {
         $internalLocatedSource = new InternalLocatedSource('', $this->getName(), 'Core');
         $createMethod          = function (string $name, array $params, Node\Identifier|Node\NullableType $returnType) use ($internalLocatedSource): ReflectionMethod {
+            assert($name !== '');
```
Ocramius, line 600: I think it's worth documenting the `@param` on the closure instead, if compatible

src/Reflection/ReflectionProperty.php
```diff
@@ -213,14 +233,23 @@ public function isDefault(): bool
...
+
     /**
      * Get the core-reflection-compatible modifier values.
      *
      * @return int-mask-of<ReflectionPropertyAdapter::IS_*>
      */
     public function getModifiers(): int
     {
-        return $this->modifiers;
+        /** @var int-mask-of<ReflectionPropertyAdapter::IS_*> $modifiers */
```
Ocramius, line 248: This is kinda weird - is our tooling not able to determine this sum's result?  :thinking:

src/Reflection/ReflectionProperty.php
```diff
@@ -265,6 +304,16 @@ public function isStatic(): bool
...
     }
 
+    public function isFinal(): bool
+    {
+        return ($this->modifiers & ReflectionPropertyAdapter::IS_FINAL_COMPATIBILITY) === ReflectionPropertyAdapter::IS_FINAL_COMPATIBILITY;
+    }
+
+    public function isAbstract(): bool
+    {
+        return ($this->modifiers & ReflectionPropertyAdapter::IS_ABSTRACT_COMPATIBILITY) === ReflectionPropertyAdapter::IS_ABSTRACT_COMPATIBILITY;
```
Ocramius: You should be able to cast to `(bool)` without using `===`. Should apply to the rest of equalities in this API :thinking:

src/Reflection/ReflectionProperty.php
```diff
@@ -592,9 +671,123 @@ private function computeModifiers(PropertyNode $node): int
...
+        $modifiers += $node->isProtectedSet() ? ReflectionPropertyAdapter::IS_PROTECTED_SET_COMPATIBILITY : 0;
         $modifiers += $node->isPublic() ? CoreReflectionProperty::IS_PUBLIC : 0;
+        $modifiers += ($node->flags & ReflectionPropertyAdapter::IS_FINAL_COMPATIBILITY) === ReflectionPropertyAdapter::IS_FINAL_COMPATIBILITY ? ReflectionPropertyAdapter::IS_FINAL_COMPATIBILITY : 0;
+        $modifiers += ($node->flags & Modifiers::ABSTRACT) === Modifiers::ABSTRACT ? ReflectionPropertyAdapter::IS_ABSTRACT_COMPATIBILITY : 0;
 
+        /** @phpstan-ignore return.type */
         return $modifiers;
     }
+
+    private function computeImmediateVirtual(PropertyNode $node): bool
```
Ocramius, line 685: this method :scream:

src/Reflection/ReflectionProperty.php
```diff
@@ -592,9 +671,123 @@ private function computeModifiers(PropertyNode $node): int
...
+    {
+        if ($node->hooks === []) {
```
Ocramius, line 687: This block seems unnecessary - let's drop it, and we gain a removed array allocation too :)

src/Reflection/ReflectionProperty.php
Ocramius: BTW, I read through it, and this is computing whether the property is "immediate virtual" or not: I'm unsure what that means. Can we give it a better name?

src/Reflection/ReflectionProperty.php
```diff
@@ -592,9 +671,123 @@ private function computeModifiers(PropertyNode $node): int
...
+        // Short syntax
+        if (count($setHookBody) === 1 && $setHookBody[0] instanceof Node\Stmt\Return_) {
+            return false;
+        }
+
+        $visitor   = new FindingVisitor(static fn (Node $node): bool => $node instanceof Node\Expr\Assign);
+        $traverser = new NodeTraverser($visitor);
+        $traverser->traverse($setHookBody);
+
+        foreach ($visitor->getFoundNodes() as $assigNode) {
```
Ocramius, line 764: Overall, the logic works, but any code including closures that use `$this` on another object may trip it.

src/Reflection/ReflectionProperty.php
```diff
@@ -592,9 +671,123 @@ private function computeModifiers(PropertyNode $node): int
...
+    }
+
+    /** @return array{get?: ReflectionMethod, set?: ReflectionMethod} */
+    private function createImmediateHooks(PropertyNode $node): array
+    {
+        $hooks = [];
+
+        foreach ($node->hooks as $hook) {
+            $hookName = $hook->name->name;
+            assert($hookName === 'get' || $hookName === 'set');
```
Ocramius, line 789: This assert will work for now, but may fail if hooks expand in scope (`isset`, for example)

src/Reflection/ReflectionProperty.php
```diff
@@ -592,9 +671,123 @@ private function computeModifiers(PropertyNode $node): int
...
+    }
+
+    /** @return array{get?: ReflectionMethod, set?: ReflectionMethod} */
+    private function createCachedHooks(): array
+    {
+        $hooks = $this->immediateHooks;
+
+        if (isset($hooks['get'], $hooks['set'])) {
+            return $hooks;
+        }
```
Ocramius, line 823: Perhaps better to check if `$hooks === []`?

src/Reflection/ReflectionPropertyHookType.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Roave\BetterReflection\Reflection;
+
+use PropertyHookType as CoreReflectionPropertyHookType;
+
+enum ReflectionPropertyHookType: string
```
Ocramius, line 17: Overall OK with this type, although some minimal docs would be required. A link to the upstream RFC would suffice, for example.

src/Reflection/StringCast/ReflectionPropertyStringCast.php
```diff
@@ -37,6 +39,18 @@ public static function toString(ReflectionProperty $propertyReflection): string
...
+    private static function docCommentToString(ReflectionProperty $propertyReflection, bool $indent): string
+    {
+        $docComment = $propertyReflection->getDocComment();
+
+        if ($docComment === null) {
+            return '';
+        }
+
+        return ($indent ? preg_replace('/(\n)(?!\n)/', '\1    ', $docComment) : $docComment) . "\n";
+    }
```
Ocramius: This doc comment indent logic is repeated multiple times already: perhaps worth externalizing?

The logic is roughly:

* give an indent number
* apply indent if the docComment is not empty

src/SourceLocator/Type/PhpInternalSourceLocator.php
```diff
@@ -41,8 +45,19 @@ private function getClassSource(Identifier $identifier): InternalLocatedSource|n
...
+
+        if (class_exists($className, false)) {
+            $reflectionClass = new CoreReflectionClass($className);
+
+            if (strtolower($reflectionClass->getName()) !== strtolower($className)) {
+                $aliasName  = $className;
+                $className  = $reflectionClass->getName();
+                $identifier = new Identifier($className, $identifier->getType());
+            }
+        }
```
Ocramius, line 58: Tricky stuff! So now we can handle aliases too :o

phpstan.neon
kukulich: `PropertyHookType` is only in PHP 8.4. We can require PHP 8.4 and it will not be necessarily. However we will not be able to run `infection` or `Psalm` now.

I've made BR classes compatible for 8.2+, the only incompatibility is this enum in adapters. However the adapters work in PHP 8.2 until you try to call methods using the enum.

src/Reflection/Adapter/ReflectionAttribute.php
kukulich: Yes, it's change in PHP 8.4: https://github.com/php/php-src/blob/PHP-8.4/UPGRADING#L409

It's not possible to change it in constructor: https://3v4l.org/lgklF#v8.4.1
I've used the some pattern as used in other adapters.

phpstan.neon
Ocramius: Ah, I see what you mean now: so we keep running everything on 8.3 until we got PHP 8.4 as a default lockfile state test setup.

src/SourceLocator/Type/PhpInternalSourceLocator.php
kukulich: Yes, there's probably first internal alias in PHP 8.4 :)

src/Reflection/ReflectionProperty.php
kukulich: The assert is there mostly for better static analysis and better types, eg. `array{get?: ReflectionMethod, set?: ReflectionMethod}`

Ocramius: Alrighty, will glance over it, for now :)

src/Reflection/ReflectionProperty.php
kukulich: The only situation when I don't need to check parent class is when I have both hooks.

I've added a comment to clarify this is just optimization.

src/Reflection/ReflectionPropertyHookType.php
kukulich: Minimal documentation and motivation added.

src/Reflection/StringCast/ReflectionPropertyStringCast.php
kukulich: Added `ReflectionStringCastHelper`. However I've preserved the boolean parameter but changed the name to be more meaningful.

src/Reflection/Adapter/ReflectionClass.php
kukulich: I've added `@see`. I think `@inheritDoc` does not make sense because the constants don't exist in parent - we use `_COMPATIBILITY` suffix otherwise it does not work.

src/Reflection/Adapter/ReflectionClass.php
kukulich: Added new exception and `@return never` in last two commits.

Ocramius:
@kukulich can you check (and eventually dismiss) the remaining open threads? Happy to ship afterwards :)

src/Reflection/ReflectionProperty.php
kukulich: We use `immeditate` in other methods to compute info coming from current class. So I've used the same pattern... Naming is hard :D

src/Reflection/ReflectionProperty.php
kukulich: The block is necessary :) I have 19 failing tests without the code :D The logic after this code would have be more complicated without this if.

src/Reflection/ReflectionProperty.php
kukulich: Yes, PHPStan needs it currently.

src/Reflection/ReflectionProperty.php
kukulich: Yes, PHPStan needs it currently.

src/Reflection/ReflectionProperty.php
kukulich: Simplified and added other commit to simply other equalities.

src/Reflection/ReflectionClass.php
kukulich: I've tried but Psalm does not understand it.

src/Reflection/ReflectionProperty.php
Ocramius: Fair enough :)

Ocramius (approved):
LGTM: stellar work as usual, @kukulich! :rocket:

src/Reflection/ReflectionProperty.php
kukulich: Yes, I can make it more complicated to ignore $this in anonymous classes, but I’n not sure if it will be enough.

PR #1465, opened 2024-12-11
Property hooks fixes for asymmetric promoted properties 

kukulich:


Ocramius (approved):
This hook stuff is like Vader telling Han Solo to pray that the deal won't get worse.

Anyway, simplified code **and** fixed an edge case: thanks @kukulich !

PR #1466, opened 2024-12-11
Fixed hooks parameters and return types

kukulich:
Just another bug found by @ondrejmirtes

### DIFF
--- a/src/Reflection/ReflectionFunctionAbstract.php
+++ b/src/Reflection/ReflectionFunctionAbstract.php
@@ -39,13 +39,13 @@ trait ReflectionFunctionAbstract
      * @var array<non-empty-string, ReflectionParameter>
      * @psalm-allow-private-mutation
      */
-    private array $parameters;
+    protected array $parameters;
 
     /** @psalm-allow-private-mutation */
     private bool $returnsReference;
 
     /** @psalm-allow-private-mutation */
-    private ReflectionNamedType|ReflectionUnionType|ReflectionIntersectionType|null $returnType;
+    protected ReflectionNamedType|ReflectionUnionType|ReflectionIntersectionType|null $returnType;
 
     /**
      * @var list<ReflectionAttribute>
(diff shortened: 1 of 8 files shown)


src/Reflection/ReflectionFunctionAbstract.php
```diff
@@ -39,13 +39,13 @@ trait ReflectionFunctionAbstract
...
      */
-    private array $parameters;
+    protected array $parameters;
 
     /** @psalm-allow-private-mutation */
     private bool $returnsReference;
 
     /** @psalm-allow-private-mutation */
-    private ReflectionNamedType|ReflectionUnionType|ReflectionIntersectionType|null $returnType;
+    protected ReflectionNamedType|ReflectionUnionType|ReflectionIntersectionType|null $returnType;
```
Ocramius, line 48: Not a fan of using `protected`  (plus a trait), so we may really need to do some changing here in the next versions.

test/unit/Reflection/ReflectionMethodTest.php
```diff
@@ -810,4 +811,71 @@ public function testWithCurrentClass(): void
...
+            ['hookWithInteger', 'int'],
+            ['hookWithUnion', 'string|int'],
+            ['hookWithoutType', null],
+        ];
+    }
+
+    #[DataProvider('getPropertyHookReturnTypeProvider')]
+    public function testGetPropertyHookReturnType(string $propertyName, string|null $returnType): void
+    {
+        $reflector = new DefaultReflector(new SingleFileSourceLocator(__DIR__ . '/../Fixture/PropertyHooks.php', $this->astLocator));
```
Ocramius, line 873: One note on the design of these tests: having everything inside a third-party file makes them a bit harder to locate when jumping around source, compared to using a string locator inlined with the test

Ocramius (approved):
:ship: thanks @kukulich!

PR #1467, opened 2024-12-11
Kill some mutants

kukulich:


Ocramius (approved):
Thanks for improving the state of mutations here, @kukulich!

PR #1470, opened 2024-12-18
Property visibility is not relevant for virtuality

kukulich:
Days without bug in property hooks: 0


Ocramius (approved):
Thanks @kukulich!

PR #1474, opened 2024-12-30
Property hooks can be marked `final` - expose that in reflection API

kukulich:


src/Reflection/ReflectionMethod.php
```diff
@@ -305,13 +305,18 @@ public function getModifiers(): int
...
+            $modifiers += $node->isStatic() ? CoreReflectionMethod::IS_STATIC : 0;
+            $modifiers += $node->isPublic() ? CoreReflectionMethod::IS_PUBLIC : 0;
+            $modifiers += $node->isProtected() ? CoreReflectionMethod::IS_PROTECTED : 0;
+            $modifiers += $node->isPrivate() ? CoreReflectionMethod::IS_PRIVATE : 0;
+            $modifiers += $node->isAbstract() ? CoreReflectionMethod::IS_ABSTRACT : 0;
+        }
+
         $modifiers += $node->isFinal() ? CoreReflectionMethod::IS_FINAL : 0;
 
         return $modifiers;
```
Ocramius, line 322: BTW, no idea how PHP optimizes this, but it's a giant cascade of ternaries D: 

I wonder if this could be a single bitwise or operation:

```php
$modifiers = $modifiers |
    ($node->isStatic() & CoreReflectionMethod::IS_STATIC) |
    ($node->isPublic() & CoreReflectionMethod::IS_PUBLIC) |
    ($node->isProtected() & CoreReflectionMethod::IS_PROTECTED) |
    ($node->isPrivate() & CoreReflectionMethod::IS_PRIVATE) |
    ($node->isAbstract() & CoreReflectionMethod::IS_ABSTRACT);
```

Not for this patch, just for future use-cases. The method calls here are still the biggest offender.

Ocramius (approved):
Looking good!

Ocramius:
Promoted from bug to improvement, since this augments hooks, rather than fixing a regression.

Ocramius:
Thanks @kukulich!

Issue #614, opened 2020-05-28
SingleFileSourceLocator is too ineffective about unknown identifiers


ondrejmirtes:
When asked about an unknown identifier, it always parses the source code again.

Instead, it should parse the file only a single time and remember all the parsed symbols.

I know there's MemoizingSourceLocator but it doesn't help really. It will only remember `null` when asked about the same identifiers for the second time.

ondrejmirtes:
My PoC implementation that works well: https://github.com/phpstan/phpstan-src/blob/master/src/Reflection/BetterReflection/SourceLocator/OptimizedSingleFileSourceLocator.php

Ocramius:
Possibly related to how the `AbstractSourceLocator` does not cache the created "located source":

https://github.com/Roave/BetterReflection/blob/5b2a0a79f3ec993b94d183e04c66d4ce99ea247c/src/SourceLocator/Type/AbstractSourceLocator.php#L39-L49
https://github.com/Roave/BetterReflection/blob/5b2a0a79f3ec993b94d183e04c66d4ce99ea247c/src/SourceLocator/Type/AbstractSourceLocator.php#L52-L59

Note how the entire source is re-created each time:

https://github.com/Roave/BetterReflection/blob/5b2a0a79f3ec993b94d183e04c66d4ce99ea247c/src/SourceLocator/Type/SingleFileSourceLocator.php#L36-L49

We can most likely avoid the expensive I/O operations ( `file_get_contents()` ), at the price of more memory usage.

Still, the `AstLocator` will then repeat the traversal of the whole tree of symbols: 

https://github.com/Roave/BetterReflection/blob/5b2a0a79f3ec993b94d183e04c66d4ce99ea247c/src/SourceLocator/Ast/Locator.php#L58-L69

PR #1475, opened 2025-01-04
Adding methods for sequential processing and filtering of class reflections

shcherbanich:
I made a small optimization of the `FileIteratorSourceLocator` so that all files are not loaded into the array before filtering, but only after it, and also added new functionality for sequential work with files and the ability to filter them.

**Motivation**

When working with really big projects, performance issues often arise. They can be solved with some techniques, such as sequential file processing. This was my main motivation for this work.

**What has been done?**

1) New methods for sequentially obtaining `*Reflection` have been added to `SourceLocators`;
2) A wrapper method has been added to `DefaultReflector` for sequentially obtaining reflect classes;
3) A filtering system has been implemented for obtained `*Reflection`. In most cases, filters are triggered before all complex operations with AST, so filtering in this way is a much more productive method than obtaining individual classes using `reflectClass`;
4) The method for processing files in `FileIteratorSourceLocator` has been changed. Now they are processed sequentially, and not immediately loaded into an array. This reduces resource consumption and speeds up the work of this locator.

### DIFF
--- a/demo/parsing-whole-directory/example3.php
+++ b/demo/parsing-whole-directory/example3.php
@@ -1,6 +1,6 @@
 <?php
 
-# parse all classes in a directory that use some dependencies from /vendor
+# example of class iterator operation
 
 require_once __DIR__ . '/../../vendor/autoload.php';
 
--- a/demo/parsing-whole-directory/example3.php
+++ b/demo/parsing-whole-directory/example3.php
@@ -1,0 +1,37 @@
+<?php
+
+# parse all classes in a directory that use some dependencies from /vendor
+
+require_once __DIR__ . '/../../vendor/autoload.php';
+
+use Roave\BetterReflection\BetterReflection;
+use Roave\BetterReflection\Reflector\DefaultReflector;
+use Roave\BetterReflection\SourceLocator\Type\AggregateSourceLocator;
+use Roave\BetterReflection\SourceLocator\Type\DirectoriesSourceLocator;
+use Roave\BetterReflection\SourceLocator\Type\SourceFilter\FileSizeFilter;
+use Roave\BetterReflection\SourceLocator\Type\SourceFilter\AggregateFilter;
+use Roave\BetterReflection\SourceLocator\Type\SourceFilter\SourceContainsFilter;
+
+$directories = [__DIR__ . '/../../src', __DIR__ . '/../../demo'];
+
+$sourceLocator = new AggregateSourceLocator([
+    new DirectoriesSourceLocator(
+        $directories,
+        (new BetterReflection())->astLocator()
+    ),
+]);
+
+$reflector = new DefaultReflector($sourceLocator);
+
+$classReflections = $reflector->iterateClasses(
+    new AggregateFilter(
+        new FileSizeFilter(10000),
+        new SourceContainsFilter(['class ReflectionMethod', 'class ReflectionClass'])
+    ),
+);
+
+print iterator_count($classReflections);
+
+$classReflections = $reflector->iterateClasses(new SourceContainsFilter(['class MyClass']));
+
+print iterator_count($classReflections);
--- a/src/Reflector/DefaultReflector.php
+++ b/src/Reflector/DefaultReflector.php
@@ -4,14 +4,15 @@ declare(strict_types=1);
 
 namespace Roave\BetterReflection\Reflector;
 
+use Iterator;
 use Roave\BetterReflection\Identifier\Identifier;
 use Roave\BetterReflection\Identifier\IdentifierType;
 use Roave\BetterReflection\Reflection\ReflectionClass;
 use Roave\BetterReflection\Reflection\ReflectionConstant;
 use Roave\BetterReflection\Reflection\ReflectionFunction;
 use Roave\BetterReflection\Reflector\Exception\IdentifierNotFound;
+use Roave\BetterReflection\SourceLocator\Type\SourceFilter\SourceFilter;
 use Roave\BetterReflection\SourceLocator\Type\SourceLocator;
-
 use function assert;
 
 final class DefaultReflector implements Reflector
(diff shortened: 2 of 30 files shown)


src/Reflector/DefaultReflector.php
```diff
@@ -56,6 +57,23 @@ public function reflectAllClasses(): iterable
...
+    {
+        /** @var Generator<ReflectionClass> $allClasses */
+        $allClasses = $this->sourceLocator->iterateIdentifiersByType(
+            $this,
+            new IdentifierType(IdentifierType::IDENTIFIER_CLASS),
+            $filter,
+        );
+
+        yield from $allClasses;
+    }
```
Ocramius, line 75: I don't think `SourceFilter` should be given at call-time: it should be an implementation detail of the source locator only, IMO.

src/Reflector/DefaultReflector.php
```diff
@@ -56,6 +57,23 @@ public function reflectAllClasses(): iterable
         return $allClasses;
     }
 
+    /**
+     * Iterate classes available in the scope specified by the SourceLocator.
+     *
+     * @return Generator<ReflectionClass>
```
Ocramius, line 63: Overall, I'd say that we can change the default `reflectAllClasses()` API to return an `iterable<ReflectionClass>`, rather than adding new API here.

Yes, it's a breaking change, but it's a good one to have

src/SourceLocator/Type/AbstractSourceLocator.php
```diff
@@ -68,4 +70,31 @@ final public function locateIdentifiersByType(Reflector $reflector, IdentifierTy
...
+
+    /**
+     * {@inheritDoc}
+     *
+     * @throws ParseToAstFailure
+     */
+    public function iterateIdentifiersByType(
+        Reflector $reflector,
+        IdentifierType $identifierType,
+        ?SourceFilter $sourceFilter,
```
Ocramius, line 82: Mentioned elsewhere: this is IMO something that shouldn't be given at call-time, but should be instead intrinsic of the way you are looking at a codebase (effectively, a `FilteredSourceLocator`)

src/SourceLocator/Type/SourceFilter/SourceFilter.php
```diff
@@ -0,0 +1,12 @@
...
+declare(strict_types=1);
+
+namespace Roave\BetterReflection\SourceLocator\Type\SourceFilter;
+
+interface SourceFilter
+{
+    public function getKey(): string;
+
+    public function isAllowed(string $source, ?string $name, ?string $filename = null): bool;
+}
```
Ocramius, line 12: I think this API is too specific to your use-case, to be considered as part of the library itself.

Also, following general FP patterns, and having BC boundaries in mind, I'd rather say that `isAllowed()` should be relying on a userland callback, and since each locator operates slightly differently, it's hard to design a generic `Context` object to be used for filtering decisions

demo/parsing-whole-directory/example3.php
```diff
@@ -0,0 +1,37 @@
...
+    ),
+]);
+
+$reflector = new DefaultReflector($sourceLocator);
+
+$classReflections = $reflector->iterateClasses(
+    new AggregateFilter(
+        new FileSizeFilter(10000),
+        new SourceContainsFilter(['class ReflectionMethod', 'class ReflectionClass'])
+    ),
```
Ocramius, line 30: From what I see, you attempted to design this code so that you could push a runtime filter down the stack, into the lookup operation, for example, to select all classes in a specific namespace. 

I understand the rationale, but am a bit conflicted on the design, and the attempt to create a structured filter here, which is extremely risky due to the internals of source locators being squishy, whilst this interface being `public`.

I don't have a clear design/suggestion for you right now, but I think a first attempt should be done at creating a single `FilteringSourceLocator` that can wrap a `$next` one and apply filtering at that level, operating on a `LocatedSource|null`: that would make the use-case more specific, and would isolate the filtering change onto a single implementation.

Ocramius (changes requested):
Patch has very good quality, and is worth looking at, but I think you spiked an approach and pushed it too hard here, leading to a bigger than necessary change upfront.

Before deciding on which approach to pick, we should attempt a second patch that:

* [ ] isolates filtering into a single, dedicated source locator implementation
* [ ] makes the default implementation of `iterable<ReflectionClass>` a `Generator` (happy to break BC there, and release a new major version!)

src/SourceLocator/Type/SingleFileSourceLocator.php
```diff
@@ -43,9 +44,28 @@ public function __construct(private string $fileName, Locator $astLocator)
...
+        }
+
         return new LocatedSource(
-            file_get_contents($this->fileName),
-            $identifier->getName(),
+            $content,
+            $name,
             $this->fileName,
         );
     }
```
Ocramius, line 71: We should probably attempt filtering in an alternative implementation of `SingleFileSourceLocator`, then look for who is instantiating `SingleFileSourceLocator`, and see if we can somehow inject the constructor for it there :thinking: 

It's just an idea, but it would allow for filtering and transforming sources before skipping or passing them to parsing/reflection

Ocramius:
Related: https://github.com/Roave/BetterReflection/issues/614

shcherbanich:
@Ocramius , thank you for the review!

I agree that the changes turned out to be too big. In this case, I suggest splitting this work into 3 separate parts with 3 separate PRs:

1) Changing the default behavior of reflectAllClasses / reflectAllFunctions / reflectAllConstants in DefaultReflector and using the generator as the default solution in source locators;

2) Implementing filtering using a new source locator (I'll think about how to do it better);

3) Optimizing SingleFileSourceLocator (I'll try to remake it in the form you noted in the comments).

I'll start doing this in the specified order if you don't mind :)


**UPD: first part completed here:** https://github.com/Roave/BetterReflection/pull/1476

staabm:
> When working with really big projects, performance issues often arise. They can be solved with some techniques, such as sequential file processing. This was my main motivation for this work.

would be great to have some kind of benchmark to compare the performance before/after this PR

PR #1476, opened 2025-01-05
Using generators to get a list of reflections

shcherbanich:
Part of PR https://github.com/Roave/BetterReflection/pull/1475 related to using generators instead of arrays to get a list of reflections

**What has been done?**

1) `locateIdentifiersByType` now returns a generator to process each individual class sequentially

2) `reflectAllClasses` / `reflectAllFunctions` / `reflectAllConstants` methods of `DefaultReflector` class now return generators instead of arrays

3) `FindReflectionOnLine` now takes reflections one by one using generators

### DIFF
--- a/src/Reflector/DefaultReflector.php
+++ b/src/Reflector/DefaultReflector.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Roave\BetterReflection\Reflector;
 
+use Generator;
 use Roave\BetterReflection\Identifier\Identifier;
 use Roave\BetterReflection\Identifier\IdentifierType;
 use Roave\BetterReflection\Reflection\ReflectionClass;
@@ -43,11 +44,11 @@ final class DefaultReflector implements Reflector
     /**
      * Get all the classes available in the scope specified by the SourceLocator.
      *
-     * @return list<ReflectionClass>
+     * @return Generator<ReflectionClass>
      */
-    public function reflectAllClasses(): iterable
+    public function reflectAllClasses(): Generator
     {
-        /** @var list<ReflectionClass> $allClasses */
+        /** @var Generator<ReflectionClass> $allClasses */
         $allClasses = $this->sourceLocator->locateIdentifiersByType(
             $this,
             new IdentifierType(IdentifierType::IDENTIFIER_CLASS),
@@ -79,11 +80,11 @@ final class DefaultReflector implements Reflector
     /**
      * Get all the functions available in the scope specified by the SourceLocator.
      *
-     * @return list<ReflectionFunction>
+     * @return Generator<ReflectionFunction>
      */
-    public function reflectAllFunctions(): iterable
+    public function reflectAllFunctions(): Generator
     {
-        /** @var list<ReflectionFunction> $allFunctions */
+        /** @var Generator<ReflectionFunction> $allFunctions */
         $allFunctions = $this->sourceLocator->locateIdentifiersByType(
             $this,
             new IdentifierType(IdentifierType::IDENTIFIER_FUNCTION),
@@ -115,11 +116,11 @@ final class DefaultReflector implements Reflector
     /**
      * Get all the constants available in the scope specified by the SourceLocator.
      *
-     * @return list<ReflectionConstant>
+     * @return Generator<ReflectionConstant>
      */
-    public function reflectAllConstants(): iterable
+    public function reflectAllConstants(): Generator
     {
-        /** @var list<ReflectionConstant> $allConstants */
+        /** @var Generator<ReflectionConstant> $allConstants */
         $allConstants = $this->sourceLocator->locateIdentifiersByType(
             $this,
             new IdentifierType(IdentifierType::IDENTIFIER_CONSTANT),
--- a/src/Reflector/Reflector.php
+++ b/src/Reflector/Reflector.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Roave\BetterReflection\Reflector;
 
+use Generator;
 use Roave\BetterReflection\Reflection\ReflectionClass;
 use Roave\BetterReflection\Reflection\ReflectionConstant;
 use Roave\BetterReflection\Reflection\ReflectionFunction;
@@ -21,9 +22,9 @@ interface Reflector
     /**
      * Get all the classes available in the scope specified by the SourceLocator.
      *
-     * @return list<ReflectionClass>
+     * @return Generator<ReflectionClass>
      */
-    public function reflectAllClasses(): iterable;
+    public function reflectAllClasses(): Generator;
 
     /**
      * Create a ReflectionFunction for the specified $functionName.
@@ -35,9 +36,9 @@ interface Reflector
     /**
      * Get all the functions available in the scope specified by the SourceLocator.
      *
-     * @return list<ReflectionFunction>
+     * @return Generator<ReflectionFunction>
      */
-    public function reflectAllFunctions(): iterable;
+    public function reflectAllFunctions(): Generator;
 
     /**
      * Create a ReflectionConstant for the specified $constantName.
@@ -49,7 +50,7 @@ interface Reflector
     /**
      * Get all the constants available in the scope specified by the SourceLocator.
      *
-     * @return list<ReflectionConstant>
+     * @return Generator<ReflectionConstant>
      */
-    public function reflectAllConstants(): iterable;
+    public function reflectAllConstants(): Generator;
 }
--- a/src/SourceLocator/Type/AbstractSourceLocator.php
+++ b/src/SourceLocator/Type/AbstractSourceLocator.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Roave\BetterReflection\SourceLocator\Type;
 
+use Generator;
 use Roave\BetterReflection\Identifier\Identifier;
 use Roave\BetterReflection\Identifier\IdentifierType;
 use Roave\BetterReflection\Reflection\Reflection;
@@ -54,15 +55,15 @@ abstract class AbstractSourceLocator implements SourceLocator
      *
      * @throws ParseToAstFailure
      */
-    final public function locateIdentifiersByType(Reflector $reflector, IdentifierType $identifierType): array
+    final public function locateIdentifiersByType(Reflector $reflector, IdentifierType $identifierType): Generator
     {
         $locatedSource = $this->createLocatedSource(new Identifier(Identifier::WILDCARD, $identifierType));
 
         if (! $locatedSource) {
-            return [];
+            return;
         }
 
-        return $this->astLocator->findReflectionsOfType(
+        yield from $this->astLocator->findReflectionsOfType(
             $reflector,
             $locatedSource,
             $identifierType,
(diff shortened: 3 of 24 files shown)


src/Reflector/DefaultReflector.php
```diff
@@ -43,11 +44,11 @@ public function reflectClass(string $identifierName): ReflectionClass
     /**
      * Get all the classes available in the scope specified by the SourceLocator.
      *
-     * @return list<ReflectionClass>
+     * @return Generator<ReflectionClass>
      */
-    public function reflectAllClasses(): iterable
+    public function reflectAllClasses(): Generator
```
Ocramius, line 49: We don't need to change the native type here: `iterable` is a supertype of `Generator`, and works fine

src/Reflector/DefaultReflector.php
```diff
@@ -43,11 +44,11 @@ public function reflectClass(string $identifierName): ReflectionClass
     /**
      * Get all the classes available in the scope specified by the SourceLocator.
      *
-     * @return list<ReflectionClass>
+     * @return Generator<ReflectionClass>
```
Ocramius, line 47: We should roll back to `iterable<int, ReflectionClass>` here

Ocramius: This change is the actual BC break, since `list<T>` is obviously an extremely small type

src/Reflector/DefaultReflector.php
```diff
@@ -79,11 +80,11 @@ public function reflectFunction(string $identifierName): ReflectionFunction
     /**
      * Get all the functions available in the scope specified by the SourceLocator.
      *
-     * @return list<ReflectionFunction>
+     * @return Generator<ReflectionFunction>
      */
-    public function reflectAllFunctions(): iterable
+    public function reflectAllFunctions(): Generator
```
Ocramius, line 85: Same as comment above: the native type can stay `iterable`, and the return type can be `iterable<int, ReflectionFunction>`

src/Reflector/DefaultReflector.php
```diff
@@ -115,11 +116,11 @@ public function reflectConstant(string $identifierName): ReflectionConstant
     /**
      * Get all the constants available in the scope specified by the SourceLocator.
      *
-     * @return list<ReflectionConstant>
+     * @return Generator<ReflectionConstant>
      */
-    public function reflectAllConstants(): iterable
+    public function reflectAllConstants(): Generator
```
Ocramius, line 121: Same as comment above: the native type can stay `iterable`, and the return type can be `iterable<int, ReflectionConstant>`

src/Reflector/Reflector.php
```diff
@@ -21,9 +22,9 @@ public function reflectClass(string $identifierName): ReflectionClass;
     /**
      * Get all the classes available in the scope specified by the SourceLocator.
      *
-     * @return list<ReflectionClass>
+     * @return Generator<ReflectionClass>
      */
-    public function reflectAllClasses(): iterable;
+    public function reflectAllClasses(): Generator;
```
Ocramius, line 27: This interface should be rolled back, and the `@return` types should switch from `list` to `iterable`

src/SourceLocator/Type/AbstractSourceLocator.php
```diff
@@ -54,15 +55,15 @@ public function locateIdentifier(Reflector $reflector, Identifier $identifier):
      *
      * @throws ParseToAstFailure
      */
-    final public function locateIdentifiersByType(Reflector $reflector, IdentifierType $identifierType): array
+    final public function locateIdentifiersByType(Reflector $reflector, IdentifierType $identifierType): Generator
```
Ocramius, line 58: I'd suggest declaring `iterable` here too: `Generator` is a very concrete type.

We can `yield from` any `iterable` anyway, and we don't need to force every implementation into a `Generator`.

See https://3v4l.org/j4S0l

src/SourceLocator/Type/AggregateSourceLocator.php
```diff
@@ -32,14 +30,10 @@ public function locateIdentifier(Reflector $reflector, Identifier $identifier):
         return null;
     }
 
-    /**
-     * {@inheritDoc}
-     */
-    public function locateIdentifiersByType(Reflector $reflector, IdentifierType $identifierType): array
+    public function locateIdentifiersByType(Reflector $reflector, IdentifierType $identifierType): Generator
```
Ocramius, line 33: The implementation looks good, but the return type should probably be `iterable`

src/SourceLocator/Type/AnonymousClassObjectSourceLocator.php
```diff
@@ -55,9 +55,14 @@ public function locateIdentifier(Reflector $reflector, Identifier $identifier):
      *
      * @throws ParseToAstFailure
      */
-    public function locateIdentifiersByType(Reflector $reflector, IdentifierType $identifierType): array
```
Ocramius, line 58: The implementation of this class can be rolled back

src/SourceLocator/Type/ClosureSourceLocator.php
```diff
@@ -56,9 +56,14 @@ public function locateIdentifier(Reflector $reflector, Identifier $identifier):
      *
      * @throws ParseToAstFailure
      */
-    public function locateIdentifiersByType(Reflector $reflector, IdentifierType $identifierType): array
+    public function locateIdentifiersByType(Reflector $reflector, IdentifierType $identifierType): Generator
```
Ocramius, line 59: The implementation of this class can be rolled back

src/SourceLocator/Type/Composer/PsrAutoloaderLocator.php
```diff
@@ -54,11 +55,11 @@ public function locateIdentifier(Reflector $reflector, Identifier $identifier):
     /**
      * Find all identifiers of a type
      *
-     * @return list<Reflection>
+     * @return Generator<Reflection>
      */
-    public function locateIdentifiersByType(Reflector $reflector, IdentifierType $identifierType): array
+    public function locateIdentifiersByType(Reflector $reflector, IdentifierType $identifierType): Generator
```
Ocramius, line 60: The implementation of this class can be rolled back, except for the return type being `iterable`

src/SourceLocator/Type/DirectoriesSourceLocator.php
```diff
@@ -55,11 +56,8 @@ public function locateIdentifier(Reflector $reflector, Identifier $identifier):
         return $this->aggregateSourceLocator->locateIdentifier($reflector, $identifier);
     }
 
-    /**
-     * {@inheritDoc}
-     */
-    public function locateIdentifiersByType(Reflector $reflector, IdentifierType $identifierType): array
+    public function locateIdentifiersByType(Reflector $reflector, IdentifierType $identifierType): Generator
```
Ocramius, line 59: The implementation of this class can be rolled back, except for the return type being `iterable`

src/SourceLocator/Type/FileIteratorSourceLocator.php
```diff
@@ -83,8 +84,8 @@ public function locateIdentifier(Reflector $reflector, Identifier $identifier):
      *
      * @throws InvalidFileLocation
      */
-    public function locateIdentifiersByType(Reflector $reflector, IdentifierType $identifierType): array
+    public function locateIdentifiersByType(Reflector $reflector, IdentifierType $identifierType): Generator
```
Ocramius, line 87: The implementation of this class can be rolled back, except for the return type being `iterable`

src/SourceLocator/Type/MemoizingSourceLocator.php
```diff
@@ -37,17 +38,23 @@ public function locateIdentifier(Reflector $reflector, Identifier $identifier):
...
         if (array_key_exists($cacheKey, $this->cacheByIdentifierTypeKeyAndOid)) {
-            return $this->cacheByIdentifierTypeKeyAndOid[$cacheKey];
+            yield from $this->cacheByIdentifierTypeKeyAndOid[$cacheKey];
+
+            return;
+        }
+
+        $items = [];
+        foreach ($this->wrappedSourceLocator->locateIdentifiersByType($reflector, $identifierType) as $item) {
+            yield $items[] = $item;
```
Ocramius, line 54: The original implementation caches items before returning them: the new implementation caches items after returning them.

I'm unsure about approach here, but it is a subtle change that can lead to uncached results if the consumer of this iterator crashes

src/SourceLocator/Type/SourceLocator.php
```diff
@@ -26,7 +27,7 @@ public function locateIdentifier(Reflector $reflector, Identifier $identifier):
     /**
      * Find all identifiers of a type
      *
-     * @return list<Reflection>
+     * @return Generator<Reflection>
      */
-    public function locateIdentifiersByType(Reflector $reflector, IdentifierType $identifierType): array;
+    public function locateIdentifiersByType(Reflector $reflector, IdentifierType $identifierType): Generator;
```
Ocramius, line 32: We should move to `iterable` rather than `Generator`

Ocramius (changes requested):

Ocramius:
@asgrim @kukulich I think the BC break is worth it here: what do you think? Is it OK to move the source locators and API from `list<T>` to `iterable<int, T>`?

asgrim:
> @asgrim @kukulich I think the BC break is worth it here: what do you think? Is it OK to move the source locators and API from `list<T>` to `iterable<int, T>`?

Reasonable change yes; any impact to our own projects we can handle but definitely keen for @kukulich to check here; depending how phpstan consumes these APIs, may or may not be impactful

kukulich:
@ondrejmirtes will this change be a problem for PHPStan?

ondrejmirtes:
@kukulich Yes, it's a BC break. For example I have custom source locator wrappers which do this:

```php
	public function locateIdentifiersByType(Reflector $reflector, IdentifierType $identifierType): array
	{
		return $this->sourceLocator->locateIdentifiersByType($reflector, $identifierType);
	}
```

If these changes really lead to better performance (so that BR does not do work that would be thrown away), it might be a better idea to introduce new methods or interfaces that return generators, and keep the old classes and interfaces there.

After that we can keep the old methods that will just do `iterator_to_array` on the new methods. Optionally we could deprecated the "old" methods.

Unfortunately if these changes go through and BetterReflection releases a new major, I could not upgrade to it in PHPStan with a clear conscience until 3.0 because these source locators are part of public PHPStan API. I know that for example Rector depends on them and also creates its own implementations.

shcherbanich:
@Ocramius @asgrim @kukulich Given all the complexities, let me go back to the first implementation, creating new methods (e.g. `iterateIdentifiersByType` / `iterateAllClasses` / ... ) and marking the old ones as deprecated, as noted in the comment above. This won't be difficult, and the change can be released safely as soon as possible.

Ocramius:
@shcherbanich let it rest a bit before jumping at it: as you can see, discussion is underway, at least :D  

> I could not upgrade to it in PHPStan with a clear conscience until 3.0 because these source locators are part of public PHPStan API.

This is a real bummer: stuff like source locators in BetterReflection are marked internal for good reason (their API is squishy), so this is unfortunate.

That said, I fully understand the problem, and will think about it today/tomorrow, before we decide whether/how to proceed.

ondrejmirtes:
Are they really internal? I'm also implementing my own and I'd say it's often useful for projects to write their own. So it might be better to un-internal them.

PR #1477, opened 2025-01-06
Simplify `FileIteratorSourceLocator`, prevent stat calls on paths that aren't PHP files

staabm:
I think this saves a system call per non PHP file as `getRealPath` is only invoked for php-files after the PR


Ocramius (approved):
LGTM :+1:

Ocramius:
Thanks @staabm!

PR #1478, opened 2025-01-24
Property from interface is abstract property

ondrejmirtes:
Days without bug in property hooks: 0

PHP proof: https://3v4l.org/B8iGK

I took the fix from ReflectionMethod because the logic is the same and it works properly there.

### DIFF
--- a/test/unit/Fixture/PropertyHooks.php
+++ b/test/unit/Fixture/PropertyHooks.php
@@ -164,3 +164,8 @@ class FinalPropertyHooks
         final set => strtolower($value);
     }
 }
+
+interface InterfaceWithProperty
+{
+    public int $abstractPropertyFromInterface { get; set; }
+}
--- a/src/Reflection/ReflectionProperty.php
+++ b/src/Reflection/ReflectionProperty.php
@@ -309,7 +309,8 @@ class ReflectionProperty
 
     public function isAbstract(): bool
     {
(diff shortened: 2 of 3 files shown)


ondrejmirtes:
Kukulich couldn't do this in his skiing gloves so I took over this time 😂

Ocramius:
> Days without bug in property hooks: 0

This new tech is keeping us employed (unpaid)!

test/unit/Fixture/PropertyHooks.php
```diff
@@ -164,3 +164,8 @@ class FinalPropertyHooks
         final set => strtolower($value);
     }
 }
+
+interface InterfaceWithProperty
+{
+    public int $abstractPropertyFromInterface { get; set; }
```
Ocramius, line 170: What happens on the concrete class / implementation? Is the declarer changed / the property becomes non-abstract?

Ocramius (changes requested):

test/unit/Fixture/PropertyHooks.php
ondrejmirtes: In class the `abstract` keyword is required so we already have that covered: https://3v4l.org/KcHto#v8.4.3

Ocramius: Fun! thanks for clarifying!

Ocramius (approved):

Ocramius:
Thanks @ondrejmirtes!

ondrejmirtes:
Thank you!

PR #1479, opened 2025-01-25
fix: some functions falsely detected as generators

sebkehr:
Relates to phpstan/phpstan-src#3794 and phpstan/phpstan#12462.

@ondrejmirtes asked me to submit an according fix here as well. 

The proposed changes fix functions returning closures, arrow functions or anonymous classes containing yield expressions to be falsely recognized as generators.

### DIFF
--- a/src/Reflection/ReflectionFunctionAbstract.php
+++ b/src/Reflection/ReflectionFunctionAbstract.php
@@ -8,6 +8,8 @@ use PhpParser\Node;
 use PhpParser\Node\Expr\Throw_ as NodeThrow;
 use PhpParser\Node\Expr\Yield_ as YieldNode;
 use PhpParser\Node\Expr\YieldFrom as YieldFromNode;
+use PhpParser\Node\FunctionLike as FunctionLikeNode;
+use PhpParser\Node\Stmt\Class_ as ClassNode;
 use PhpParser\Node\Stmt\ClassMethod as MethodNode;
 use PhpParser\NodeTraverser;
 use PhpParser\NodeVisitor\FindingVisitor;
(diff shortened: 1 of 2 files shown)


src/Reflection/ReflectionFunctionAbstract.php
```diff
@@ -360,7 +362,11 @@ private function nodeIsOrContainsYield(Node $node): bool
         foreach ($node->getSubNodeNames() as $nodeName) {
             $nodeProperty = $node->$nodeName;
 
-            if ($nodeProperty instanceof Node && $this->nodeIsOrContainsYield($nodeProperty)) {
+            if ($nodeProperty instanceof Node &&
```
Ocramius: Operator precedence making this a bit of a puzzle to read: let's add some parentheses, please 😁

test/unit/Reflection/ReflectionFunctionAbstractTest.php
```diff
@@ -236,6 +236,9 @@ public static function generatorProvider(): array
             ['<?php function foo() { $foo->func(yield $foo); }', true],
             ['<?php function foo() { new Foo(yield $foo); }', true],
             ['<?php function foo() { yield from []; }', true],
+            ['<?php function foo() { return fn () => yield $foo; }', false],
+            ['<?php function foo() { return function () { yield $foo; }; }', false],
+            ['<?php function foo() { return new class () { function bar() { yield $foo; } }; }', false]
```
Ocramius: I think we even had a PR discussion about this, last year: thanks for fixing this pothole!

Ocramius (changes requested):
Need to apply CS (and ideally squash that commit, please)

Ocramius (approved):
Thanks @sebkehr!

PR #1487, opened 2025-02-10
Fixed property from interface implemented by trait

ondrejmirtes:


ondrejmirtes:
Days without bugs in property hooks: 0
Days without bugs in traits: 0

ondrejmirtes:
The problem here is that BR looks for the property in interface first and completely skips the implementation in the used trait.

Ocramius (approved):
Noice!

I mean, it's still a monstrosity, but the fix is nice.

Ocramius:
Thanks @kukulich!

ondrejmirtes:
Thank you very much!

PR #1491, opened 2025-02-23
Upgrade to Psalm 6.6.2

kukulich:
To unblock upgrade to Psalm 6.6.

It's also possible to add `final` to all classes (instead of `api`) and use https://github.com/dg/bypass-finals

Or we can use `@psalm-suppress ClassMustBeFinal`

### DIFF
--- a/psalm.xml
+++ b/psalm.xml
@@ -17,4 +17,7 @@
     <projectFiles>
         <directory name="src" />
     </projectFiles>
+    <issueHandlers>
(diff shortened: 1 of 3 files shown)


Ocramius:
I think we should suppress the rule in `psalm.xml`

kukulich:
@Ocramius 👍  

I'll also ignore this rule: https://psalm.dev/docs/running_psalm/issues/MissingOverrideAttribute/
We still support PHP 8.2 and the attribute is part of PHP 8.3.

PR #1492, opened 2025-02-23
Upgrade to Psalm 6.8.6

kukulich:


kukulich:
@Ocramius We may add the `#[Override]` attribute. I forgot it will work in lower PHP versions too. What do you think?

Ocramius:
Yes: relatively useless attribute in first place.

PR #1502, opened 2025-04-08
Inline array_map(fn, $arr) and use foreach-loop instead

staabm:
Based on a perf review togehter with nielsdos (php-src core develoepr) we identified `array_map`, when used with a callback function to be a hotspot in PHPStan analysis, see https://phpc.social/@markusstaab/114301182551454759

I've did some research and used some hack to record the call count of `array_map` when running PHPStan and isolate the following list of [call sites from BetterReflection](https://gist.github.com/staabm/3483fecb637886fab239d1657235b496#file-sorted-log).

the list contains 3 outstanding callers which dominate the call-counts. these occurences have been adjusted with this PR.

⚠️ still work in progress

### DIFF
--- a/src/Reflection/Adapter/ReflectionClass.php
+++ b/src/Reflection/Adapter/ReflectionClass.php
@@ -384,14 +384,12 @@ final class ReflectionClass extends CoreReflectionClass
     {
         $traits = $this->betterReflectionClass->getTraits();
 
-        /** @var list<trait-string> $traitNames */
-        $traitNames = array_map(static fn (BetterReflectionClass $trait): string => $trait->getName(), $traits);
+        $result = [];
+        foreach ($traits as $trait) {
+            $result[$trait->getName()] = new self($trait);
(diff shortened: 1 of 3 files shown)


ondrejmirtes:
Please note that I'm skeptical about this: https://github.com/phpstan/phpstan-src/pull/3926#issuecomment-2786497515

Ocramius:
I'm also not really looking forward to maintaining more loops: would rather endorse pushing the concern to php-src, at the cost of performance overhead here, for now.

staabm:
Will be optimized in 
- https://github.com/php/php-src/pull/18273
- https://github.com/php/php-src/pull/18326

Ocramius:
Nice!

staabm:
php 8.5.x is running ~10% faster thanks to recent fixes.

see https://github.com/php/php-src/pull/18273#issuecomment-2815265546

PR #1508, opened 2025-05-27
PHPStan level 7

kukulich:


src/Reflection/ReflectionProperty.php
```diff
@@ -688,11 +688,12 @@ private function computeModifiers(PropertyNode $node): int
             ! ($modifiers & (ReflectionPropertyAdapter::IS_PROTECTED_SET_COMPATIBILITY | ReflectionPropertyAdapter::IS_PRIVATE_SET_COMPATIBILITY))
             && ! $node->isPublicSet()
             && $node->isPublic()
-            && ($modifiers & CoreReflectionProperty::IS_READONLY)
+            && ($modifiers & ReflectionPropertyAdapter::IS_READONLY)
         ) {
             $modifiers += ReflectionPropertyAdapter::IS_PROTECTED_SET_COMPATIBILITY;
         }
 
+        /** @phpstan-ignore return.type */
```
Ocramius, line 696: Not fixable?

Ocramius (approved):
Good improvement: let's keep moving this forward

src/Reflection/ReflectionProperty.php
kukulich: PHPStan does some optimizations and is not able to calculate all the possible combinations.

PR #1510, opened 2025-06-03
[Demo] Use Infection with native PHPStan integration

maks-rafalko:
This is not intended to be merged (at least for now), more like to have a discussion and to compare with https://github.com/Roave/BetterReflection/pull/1505

----

This PR:

- uses Infection with native integration of PHPStan. Now, we can just run `infection --static-analysis-tool=phpstan` and it will do the job. So it's opt-in, by default PHPStan is not used
- what is the difference between native integration and tools like [`roave/infection-static-analysis-plugin/`](https://github.com/Roave/infection-static-analysis-plugin/) and [`phpstan/mutant-killer-infection-runner`](https://github.com/phpstan/mutant-killer-infection-runner)? Infection runs PHPStan processes in parallel (depending on `--threads`) while those plugins run Psalm/PHPStan processes sequentially because of the nature of the the "hack" used there.

There is still too much work to do, here are the major issues for now:

- it's slow. For some (not yet known) reason PHPStan process takes at least 1s even for 1-file project
- we don't use `--stop-on-first-error` from PHPStan (because it doesn't exist yet xD) 
- many of the Mutants are killed by PHPStan when they shouldn't

For example: 

```diff
@@ @@
             if ($reflection instanceof ReflectionFunction && $this->containsLine($reflection, $lineNumber)) {
                 return $reflection;
             }
-            if ($reflection instanceof ReflectionConstant && $this->containsLine($reflection, $lineNumber)) {
+            if (true && $this->containsLine($reflection, $lineNumber)) {
                 return $reflection;
             }
         }

$ '/app/vendor/bin/phpstan' '--tmp-file=/tmp/infection/mutant.a99f5bbe9bbbb1514da2bdbe5e315d5c.infection.php' '--instead-of=/app/src/Util/FindReflectionOnLine.php' '--error-format=json' '--no-progress' '-vv'
  
PHPStan error: Left side of && is always true.
```

^ PHPStan always kills this Mutant, but in reality the tests can easily have a whole here, not testing properly the left side of this condition. I think on Infection side we need to mark some mutators (mutator operators) to **not** require static analysis, and just finish if they are escaped, not wasting time with static analysis.

cc @ondrejmirtes @staabm


maks-rafalko:
Please run GH actions. (working example is on my fork: https://github.com/maks-rafalko/BetterReflection/pull/2)

infection.json.dist
```diff
@@ -10,6 +10,9 @@
             "report": "/^\\d\\.\\d\\.x$/"
         }
     },
+    "phpStan": {
+        "customPath": "tools/vendor/bin/phpstan"
+    },
```
Ocramius, line 15: This kinda means that infection is built "for phpstan", rather than having a point-cut where to interject in mutations, no? :thinking: 

I'm fine with it, it is an improvement (and a big one), just a bit bummed, since I thought the idea was to go towards tooling that could interfere with a mutation, and now we're going towards a tightly integrated ecosystem instead?

maks-rafalko: Well, I don't see a problem here. Infection is tightly integrated with PHPUnit, for example, that's why we also have the same `phpUnit` config setting. But _it's not limited_ to it. The same is for PHPStan - we have a dedicated section for it, but Infection is not limited to PHPStan (there is already a ground for Psalm integration).

I thought initially that we could name it `staticAnalysis.customPath` but there is a couple of moments:

- Infection's architecture now allows N "mutant killers" to be executed, so PHPUnit -> PHPStan -> Psalm can live together at the same time, so `staticAnalysis.customPath` can't be used, we need both
- it's just more clear what this means and IDE autocompletes it, because `phpStan` key is in our json schema. It could be generic, but less convenient.

Anyway, this can be changed in the future, we are experimenting now.

> rather than having a point-cut where to interject in mutations, no
> ...
> I thought the idea was to go towards tooling that could interfere with a mutation

No, we are *not* building a "plugin system" (https://github.com/infection/infection/issues/1323) that has been discussed some time ago. Probably it will be like this somewhere in the future, but for now - as easy as tight integration (again, because we are experimenting)

Ocramius: > No, we are not building a "plugin system" (https://github.com/infection/infection/issues/1323) that has been discussed some time ago.

Thanks for clarifying!

maks-rafalko:
GH actions are not executed automatically :(

Now (among tens of other improvements) the output shows killed mutants by PHPStan by marking them with `A` so it looks like this:

```diff
[...]

.................A.............AAM.A....A.........   (3650 / 4409)
.......A......M.MM.AA..A.......A...A.........A....   (3700 / 4409)
......M.AAA...A.................A.A.M..MA....A....   (3750 / 4409)
.................A.......A.....................M..   (3800 / 4409)
.AA...A..A...............A.AAA......A............E   (3850 / 4409)
E..A..........A...................................   (3900 / 4409)
..........................A.......................   (3950 / 4409)
..............................T...................   (4000 / 4409)
..................................................   (4050 / 4409)
.........T........................................   (4100 / 4409)
..................................................   (4150 / 4409)
T..............E..................AA..............   (4200 / 4409)
.........M................................A.......   (4250 / 4409)
.....................................A..........E.   (4300 / 4409)
E.E.............................................AA   (4350 / 4409)
....................A.............A...............   (4400 / 4409)
......A.A                                            (4409 / 4409)

4409 mutations were generated:
    3660 mutants were killed by Test Framework
+     212 mutants were killed by Static Analysis
       0 mutants were configured to be ignored
      11 mutants were not covered by tests
      47 covered mutants were not detected
       7 errors were encountered
       0 syntax errors were encountered
       9 time outs were encountered
     463 mutants required more time than configured

Metrics:
         Mutation Score Indicator (MSI): 98%
         Mutation Code Coverage: 99%
         Covered Code MSI: 98%
```

staabm:
> GH actions are not executed automatically :(

as long as the PR has a merge conflict with the target branch no github actions will start

maks-rafalko:
looks like it requires it since I'm not a contributor

Ocramius:
Implemented in #1553, thanks @maks-rafalko!

PR #1524, opened 2025-09-09
Support indirect `Stringable` interface via `__toString()` in trait

staabm:
required for https://github.com/phpstan/phpstan/issues/13469

### DIFF
--- a/src/Reflection/ReflectionClass.php
+++ b/src/Reflection/ReflectionClass.php
@@ -1315,7 +1315,12 @@ class ReflectionClass implements Reflection
             return $interfaces;
         }
 
-        foreach (array_keys($this->immediateMethods) as $immediateMethodName) {
+        $methods = $this->immediateMethods;
+        foreach ($this->getTraits() as $trait) {
+            $methods += $trait->immediateMethods;
+        }
+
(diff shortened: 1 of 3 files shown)


src/Reflection/ReflectionClass.php
```diff
@@ -1332,6 +1332,25 @@ private function addStringableInterface(array $interfaces): array
...
+                        }
+                    } catch (IdentifierNotFound) {
+                        // Stringable interface does not exist on target PHP version
+                    }
+
+                    // @infection-ignore-all Break_: There's no difference between break and continue - break is just optimization
+                    break 2;
+                }
+            }
+        }
```
Ocramius: Instead of repeating this whole block, merge methods from the traits into immediate methods above?

Ocramius (changes requested):
Patch meaningful, but the implementation (just inside the `private` method) needs to be refactored

src/Reflection/ReflectionClass.php
staabm: good point. fixed.

Ocramius (approved):
Excellent, thanks @staabm!

PR #1526, opened 2025-09-18
Add `ReflectionEnumCase#hasValueExpression()`

VincentLanglet:
Hi @Ocramius 

We encountered an issue in PHPStan when someone analyse some (invalid) PHP code with an invalid backed enum with a case without value, PHPStan is crashing (https://github.com/phpstan/phpstan/issues/7927).

Our strategy would be to introduce a `ReflectionEnumCase::hasValueExpression` method (basically a is valid enum case). And before adding it to the Phpstan/BetterReflection fork, I wanted to know if you would be interested/ok having this method in BetterReflection because it might have some other interest.

I tried added a test case to show the situation.

Feel free to close this PR if the method doesn't interest you.
Thanks !

### DIFF
--- a/test/unit/Fixture/Enums.php
+++ b/test/unit/Fixture/Enums.php
@@ -71,8 +71,3 @@ enum IsDeprecated
 interface InterfaceForEnum
 {
 }
-
-enum InvalidEnum: int
-{
-    case INVALID;
-}
--- a/test/unit/Fixture/Enums.php
+++ b/test/unit/Fixture/Enums.php
@@ -71,3 +71,8 @@ enum IsDeprecated
 interface InterfaceForEnum
 {
 }
+
+enum InvalidEnum: int
+{
+    case INVALID;
+}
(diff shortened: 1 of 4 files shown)


Ocramius (approved):
LGTM, let's just make sure we expand a bit on what the intended method use-case is, in the docblock

test/unit/Fixture/Enums.php
```diff
@@ -71,3 +71,8 @@ enum IsDeprecated
 interface InterfaceForEnum
 {
 }
+
+enum InvalidEnum: int
```
Ocramius: Needs to be moved somewhere where it won't be analysed 🤔

VincentLanglet: Fixed

VincentLanglet:
> LGTM, let's just make sure we expand a bit on what the intended method use-case is, in the docblock

I tried to add some explanations in the phpDoc

Ocramius (approved):
🚢  thanks @VincentLanglet

PR #1527, opened 2025-09-19
Stop returning `ReflectionEnumBackedCase` without value

VincentLanglet:
Other solution for https://github.com/Roave/BetterReflection/pull/1528

Only one of both PR should be merged


Ocramius:
Thanks @VincentLanglet!

PR #1528, opened 2025-09-19
Add ReflectionEnumBackedCase::hasBackingValue

VincentLanglet:
Sorry @Ocramius, I feel like I missed something in https://github.com/Roave/BetterReflection/pull/1526

PHPStan (and I assume some other code) is working with `ReflectionEnumBackedCase`, because it uses
```
ReflectionEnum::getCases()
```
and since `betterReflectionEnumCase` is not public, I cannot call `hasValueExpression` on it.

So I have two solution in mind:

1) Exposing `ReflectionEnumBackedCase::hasBackingValue` => This PR

2) The other solution would be to modify `ReflectionEnum::getCase()` and `ReflectionEnum::getCases()` to check for `$this->betterReflectionEnum->isBacked()` and/or `$case->hasValueExpression()`, 
https://github.com/Roave/BetterReflection/blob/d9f0f93b6c11d6738df13d8eb8777eb713fda82b/src/Reflection/Adapter/ReflectionEnum.php#L578-L604

This would be PR https://github.com/Roave/BetterReflection/pull/1527

Only one of both PR should be merged

### DIFF
--- a/src/Reflection/Adapter/ReflectionEnumBackedCase.php
+++ b/src/Reflection/Adapter/ReflectionEnumBackedCase.php
@@ -125,6 +125,11 @@ final class ReflectionEnumBackedCase extends CoreReflectionEnumBackedCase
         return new ReflectionEnum($this->betterReflectionEnumCase->getDeclaringEnum());
     }
 
+    public function hasBackingValue(): bool
+    {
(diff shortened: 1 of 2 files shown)


Ocramius:
I think #1527 provides an implementation change without changing/expanding the API, which is preferable, thanks! :+1:

PR #1538, opened 2025-11-04
`ReflectionProperty#setValue()` always requires 2 arguments

### LINKED ISSUE #1537: Deprecation in 8.3 `ReflectionProperty::setValue`
`ReflectionProperty::setValue` with 1 argument is deprecated in PHP 8.3 in favor of `setValue(null, $value)`.
Source: https://www.php.net/manual/en/reflectionproperty.setvalue.php#refsect1-reflectionproperty.setvalue-changelog

Example in codebase: https://github.com/Roave/BetterReflection/blob/6.66.x/src/Reflection/Adapter/ReflectionClass.php#L603
Stumbled upon this as I was doing research for PHPCompatibility.

Ocramius:
Ref: https://www.php.net/manual/en/reflectionproperty.setvalue.php#refsect1-reflectionproperty.setvalue-changelog

In `ext-reflection`, usage with a single argument is deprecated.

Fixes #1537

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -6,6 +6,7 @@
     colors="true"
     executionOrder="random"
     cacheDirectory=".phpunit.cache"
+    failOnDeprecation="true"
 >
     <source>
         <include>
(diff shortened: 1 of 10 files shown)


Ocramius:
@afilina can you skim over this, and see if it solves your issue?

afilina:
@Ocramius I was merely doing static analysis on top 2000 packages to understand how the deprecated feature was used. That addresses the deprecation. Thanks! It also confirms that my research is sound :)

PR #1544, opened 2026-01-13
Fixed `ReflectionMethod::isGenerator()` to work on class methods too

kukulich:
I've made only a small unit test because it's already tested for `ReflectionFunctionAbstract`: https://github.com/Roave/BetterReflection/blob/6.67.x/test/unit/Reflection/ReflectionFunctionAbstractTest.php#L219-L253


Ocramius (approved):
Very nice and clean, thanks @kukulich!

PR #1546, opened 2026-01-21
PHPUnit 12.5

kukulich:


Ocramius:
```
5 tests triggered 1 PHP deprecation:

1) /home/runner/work/BetterReflection/BetterReflection/src/SourceLocator/SourceStubber/PhpStormStubs/CachingVisitor.php:217
Constant E_STRICT is deprecated since 8.4, the error level was removed
```

That seems like our problem :thinking:

kukulich:
`PhpStormStubsSourceStubber` reads the constant value if the constant is available. Some constans change their value in different environment or different PHP versions, so stubs are not able to have the right constant value.

https://github.com/Roave/BetterReflection/blob/6.69.x/src/SourceLocator/SourceStubber/PhpStormStubs/CachingVisitor.php#L217

The error is silenced by `@` but PHPUnit modifies error reporting. That's why the deprecated warnings were disabled in PHPUnit.

I don't know what's the best and nice solution how to solve it... :(

kukulich:
@Ocramius Solved in https://github.com/Roave/BetterReflection/pull/1546/changes/627fd5ed0f55ab3c09575629e3ef62c2a91d087d

Ocramius (approved):
thanks @kukulich !

PR #1548, opened 2026-02-09
Test deprecated constant via `define()`

staabm:
while hunting a bug I suspected BetterReflection would not properly see constants as deprecated if PHPStorm stubs contain it.

seems I was wrong,...  but I think the test is still useful, so I figured I would submit it.

background: https://php.watch/versions/8.4/E_STRICT-deprecated


Ocramius (approved):
Thanks @staabm!

PR #1555, opened 2026-03-31
fix: Pass promoted property default value

aragon999:
I ran into the issue, when changing normal properties to promoted properties: https://github.com/shopware/shopware/pull/15889 in combination with the `roave-backward-compatibility-check`.

I think actually the test `Roave\BetterReflectionTest\Reflection\ReflectionPropertyTest::testIsPromoted` had a bug, but I still added a new one, according to the contribution guidelines.


Ocramius:
Build failure is not related to your work, @aragon999: patch is valid.

Man, the infection+psalm issue is plaguing this repository now.

aragon999:
I see, thank you for the quick response, and also thank you for the correction of the import order :see_no_evil: 

I'm afraid I don't know enough about the combination of "infection+psalm" to be of help in that regard, but thank you for maintaining the project :-)

aragon999:
I noticed that with this change, reflection helper fails when the default value is an object. I will look into it, how it can be solved.

Ocramius:
I may need to roll back the dependencies in here: will first try and see if I can reproduce the problem locally, or if it's a CI issue.

Ocramius:
> reflection helper fails when the default value is an object.

That's normal: BetterReflection **cannot** instantiate any objects, since that would imply autoloading, which is forbidden by the core invariant of this library.

The entire point of BetterReflection is to reflect without causing side-effects (which occur in `ext-reflection` instead)

aragon999:
> > reflection helper fails when the default value is an object.
> 
> That's normal: BetterReflection **cannot** instantiate any objects, since that would imply autoloading, which is forbidden by the core invariant of this library.
> 
> The entire point of BetterReflection is to reflect without causing side-effects (which occur in `ext-reflection` instead)

Yes, I was not thinking about checking the instance, but rather passing the information that the default value is of a specific class. As with this change the BC checker fails
```

[BC] SKIPPED: Unable to compile initializer in class Shopware\Core\Profiling\Doctrine\ProfilingMiddleware in file /src/Core/Profiling/Doctrine/ProfilingMiddleware.php (line 17)
[BC] SKIPPED: Unable to compile initializer in class Shopware\Core\Framework\Context in file /src/Core/Framework/Context.php (line 52)
[BC] SKIPPED: Unable to compile initializer in class Shopware\Core\Content\Media\Thumbnail\ExternalThumbnailsParameters in file /src/Core/Content/Media/Thumbnail/ExternalThumbnailsParameters.php (line 14)
```
See for example the following run: https://github.com/shopware/shopware/actions/runs/23821976081/job/69436210922?pr=15893#step:7:1020
And the following code: https://github.com/shopware/shopware/blob/trunk/src/Core/Content/Media/Thumbnail/ExternalThumbnailsParameters.php#L14

But I will further look into it. I hopefully find some time tomorrow.

Ocramius (commented):
Giving a LGTM here meanwhile: I still need to fix the CI pipeline before merging, though

Ocramius:
Thanks @aragon999!

PR #1557, opened 2026-04-07
Add IS_PROTECTED_SET_COMPATIBILITY modifier with readonly class

### LINKED ISSUE #1556: ReflectionProperty might miss `IS_PROTECTED_SET_COMPATIBILITY` modifier with readonly class
BetterReflection::computeModifiers has a special behavior about public readonly property to add the protected set modifier
https://github.com/Roave/BetterReflection/blob/7f0e5fdb10373fbf17f9dbadcc7c63473e6ec712/src/Reflection/ReflectionProperty.php#L687-L694
in order to reflect the existing PHP behavior when the property is public but readonly
https://3v4l.org/gvoPg#v8.5.3

But the same PHP behavior exists when the class is readonly
https://3v4l.org/NaiQT#v8.5.3

So I wonder @Ocramius if
- The ReflectionProperty should automatically add the IS_PROTECTED_SET_COMPATIBILITY modifier if the class is readonly and the property is public without publicSet. (Similar to the existing logic in computeModifiers)
- Or if it's expected to not check the class at all in the ReflectionProperty, in the same way that the ReflectionProperty is not set as "readonly" or "final" when the class is and the property is not directly defined readonly/final.

VincentLanglet:
Closes https://github.com/Roave/BetterReflection/issues/1556

### DIFF
--- a/src/Reflection/ReflectionProperty.php
+++ b/src/Reflection/ReflectionProperty.php
@@ -688,7 +688,10 @@ class ReflectionProperty
             ! ($modifiers & (ReflectionPropertyAdapter::IS_PROTECTED_SET_COMPATIBILITY | ReflectionPropertyAdapter::IS_PRIVATE_SET_COMPATIBILITY))
             && ! $node->isPublicSet()
             && $node->isPublic()
-            && ($modifiers & ReflectionPropertyAdapter::IS_READONLY)
+            && (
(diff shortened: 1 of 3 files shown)


Ocramius (approved):
Looking good! Like #1555, the problem here is the CI setup, for which I'm still wondering about a fix...

Ocramius:
Thanks @VincentLanglet!

PR #1574, opened 2026-08-06
Compile closures and first-class callables in constant expressions (PHP 8.5)

janedbal:
## Problem

PHP 8.5 allows static closures and first-class callables in constant expressions ([RFC](https://wiki.php.net/rfc/closures_in_const_expr)), for example in attribute arguments:

```php
#[Assert\Callback(static function (self $object, ExecutionContextInterface $context): void {
    // ...
})]
readonly class CreatedAt { /* ... */ }
```

`CompileNodeToValue` does not recognize `Expr\Closure` and first-class-callable nodes. Any call to `ReflectionAttribute::getArguments()` (or `ReflectionParameter::getDefaultValue()`, etc.) on such code throws:

```
Unable to compile expression in class ...: unrecognized node type PhpParser\Node\Expr\Closure
```

Real-world report: shipmonk-rnd/dead-code-detector#404 (PHPStan extensions crash via the bundled fork of this library). Downstream mitigation: shipmonk-rnd/dead-code-detector#405. Fork alternative (given the design-invariant objection below): ondrejmirtes/BetterReflection#44.

## Fix

Compile these nodes into real `Closure` instances, like native reflection does:

- Pretty-print the node and `eval()` it inside the declaring namespace. The language guarantees const-expr closures are **static and capture no variables**, so evaluating the declaration itself executes no user code (unlike `new` initializers, which stay rejected).
- Bind the declaring class scope (`Closure::bind`) so `self::` works inside the body, like at runtime — but only when the class is already loaded; autoloading is never triggered.
- Resolve `self`/`parent`/`static` in first-class callables (`self::foo(...)`) before printing.

`ReflectionParameter::getDefaultValue()` and `ReflectionProperty::getDefaultValue()` are widened to `mixed`, because a default value can now be a `Closure`. Native reflection declares `mixed` there as well (the adapters already do).

## BC check

The `Check Backward Compatibility` job fails on the two `getDefaultValue()` return types (narrow union to `mixed`). This is intentional: a default value can legally be a `Closure` since PHP 8.5, so the narrow type would produce a `TypeError` instead. I see no way to support this without the widening; happy to adjust if you prefer another approach (e.g. a dedicated exception instead of the type change).

Note: arrow functions need no handling — PHP 8.5 rejects them in constant expressions.

## Tests

The new tests in `ReflectionAttributeTest` and `CompileNodeToValueTest` fail on current 6.73.x with the exact error above and pass with this change. The scope-binding test requires PHP 8.5 (the fixture must be natively loadable); the rest also run on PHP 8.4, since only the analyzed code needs 8.5 syntax.

Co-Authored-By: Claude Code

### DIFF
--- a/src/NodeCompiler/CompileNodeToValue.php
+++ b/src/NodeCompiler/CompileNodeToValue.php
@@ -4,8 +4,10 @@ declare(strict_types=1);
 
 namespace Roave\BetterReflection\NodeCompiler;
 
+use Closure;
 use PhpParser\ConstExprEvaluator;
 use PhpParser\Node;
+use PhpParser\PrettyPrinter\Standard as PrettyPrinter;
 use Roave\BetterReflection\Reflection\ReflectionClass;
 use Roave\BetterReflection\Reflection\ReflectionClassConstant;
 use Roave\BetterReflection\Reflection\ReflectionEnum;
@@ -14,6 +16,7 @@ use Roave\BetterReflection\Reflector\Exception\IdentifierNotFound;
 use Roave\BetterReflection\Util\FileHelper;
 
 use function assert;
+use function class_exists;
 use function constant;
 use function defined;
 use function dirname;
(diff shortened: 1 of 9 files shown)


Ocramius:
The main problem here is that this causes autoloading, which goes against the foundational design invariant of this library

src/NodeCompiler/CompileNodeToValue.php
```diff
@@ -176,6 +190,43 @@ public function __invoke(Node $node, CompilerContext $context): CompiledValue
...
+        $code = (new PrettyPrinter())->prettyPrintExpr($node);
+
+        $namespace = $context->getNamespace();
+
+        // The namespace wrapper keeps the fallback to global scope for unqualified function calls
+        $code = $namespace !== null && $namespace !== ''
+            ? sprintf('namespace %s; return %s;', $namespace, $code)
+            : sprintf('return %s;', $code);
+
+        $closure = eval($code);
```
Ocramius, line 217: Not feasible/acceptable, sorry. See other feedback.

Ocramius (changes requested):

PR #1579, opened 2026-09-09
`LanguageLevelTypeAware` pseudo-types should be turned into phpdoc-type

### LINKED ISSUE #1578: `LanguageLevelTypeAware` turned into native `resource` type
in https://github.com/phpstan/phpstan-src/pull/6378 PHPStan surfaced a problem caused by `#[LanguageLevelTypeAware(...)]`

in e.g.
```php
function finfo_buffer(#[LanguageLevelTypeAware(['8.1' => 'finfo'], default: 'resource')] $finfo, string $string, int $flags = FILEINFO_NONE, #[\JetBrains\PhpStorm\Deprecated('Deprecated: it has no effect', since: '8.5')] $context = null): string|false {}
```

better reflection reports a native type `Name`'d `resource` which is than looked up as a class-name.
[`resource` is only valid as a phpdoc type](https://3v4l.org/h3QkW#v).

I am still in the process of understanding how this all fits together, therefore opening a issue for now

staabm:
closes https://github.com/Roave/BetterReflection/issues/1578

### DIFF
--- a/src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
+++ b/src/SourceLocator/SourceStubber/PhpStormStubsSourceStubber.php
@@ -18,6 +18,17 @@ use PhpParser\BuilderFactory;
 use PhpParser\BuilderHelpers;
 use PhpParser\Comment\Doc;
 use PhpParser\Node;
+use PhpParser\Node\ComplexType;
+use PhpParser\Node\Identifier;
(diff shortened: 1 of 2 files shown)


SanderMuller (approved):

Ocramius (approved):
Tricky one, good catch @staabm!

PR #1580, opened 2026-09-10
Use StructArmed on QA

samsonasik:
This PR adds `StructArmed` to QA:

https://github.com/boundwize/structarmed

It is a configurable PHP architecture guards with built in presets (psr4, mvc, and so on). This PR apply `StructArmed` with register:

- PSR4 Preset
  This PR start apply StructArmed with initialize using PSR4 preset, and already help fix 8 violations.

- I also enable ruleset so it show more ability of layering setup to define which layers each layer is allowed to depend on.

- Register to Github workflow.


samsonasik:
@Ocramius ready for review/merge 👍

samsonasik:
Rebased 👍

Ocramius:
Hey @samsonasik, just FYI, I didn't ignore your patches about this (and on other repositories as well), but I simply didn't get to play around with the tool yet.

@kukulich @asgrim do you have opinions on a layer structure we could follow for this project, BTW?

test/unit/SourceLocator/Type/Composer/Psr/Excetion/InvalidPrefixMappingTest.php
```diff
@@ -2,7 +2,7 @@
 
 declare(strict_types=1);
 
-namespace Roave\BetterReflectionTest\SourceLocator\Type\Composer\Psr\Exception;
+namespace Roave\BetterReflectionTest\SourceLocator\Type\Composer\Psr\Excetion;
```
samsonasik, line 5: For note: this probably just need folder rename instead, as current folder is named `Excetion` instead of `Exception` so my first attempt is just follow folder name.

samsonasik:
@Ocramius that's fine, if ruleset layering need in separate PR and just need apply PSR4 for starter, just let me know ;)

kukulich:
I'm not against new check tool however the configuration looks too complicated to me: https://github.com/Roave/BetterReflection/pull/1580/changes#diff-6c750795fe8e9ac3fc3936402fef25d2ca69ca1d01ef0f16c2ad8b27cde4b041

samsonasik:
@kukulich sure, I've updated the config to use `layer()` over `layerPattern()` when possible to simplify the config 👍 https://github.com/Roave/BetterReflection/pull/1580/changes/09704bd0005d215c627bb4356d9a64561bcd5be1

Less regex and should be ok now 👍 Just let me know if you need another simplification even more.

samsonasik:
@kukulich I've cleaned even more so no more regex in the config https://github.com/Roave/BetterReflection/pull/1580/changes/6b07a11d67e6119664f72fb551e768d2438b7eea 👍 

It clean now, no more complicated config 👍

Just let me know if you don't want the `ruleset()` definition and just the PSR4 preset for this starter PR :)

test/unit/SourceLocator/Type/Composer/Psr/Excetion/InvalidPrefixMappingTest.php
samsonasik: I've renamed `Excetion` to `Exception` folder for it 👍  https://github.com/Roave/BetterReflection/pull/1580/changes/1a5979357e83e8b4b314e3b8d04d3b619c3a80e2

samsonasik:
Rebased 👍  ready for review/merge 👍

PR #1586, opened 2026-09-19
PHP 8.6 checks

kukulich:


src/Reflection/Adapter/ReflectionParameter.php
```diff
@@ -45,6 +46,11 @@ public function getName(): string
         return $this->betterReflectionParameter->getName();
     }
 
+    public function getDocComment(): string|false
+    {
+        throw new NotImplemented('Not implemented');
```
Ocramius: I suppose we'll have to implement these?

kukulich: Yes, I just want to have green build on PHP 8.6. Other PR will follow :)

